PR #459, opened 2016-09-23
Added the ability to handle multiple WWW-Authenticate headers

winniequinn:
@aslagle  

This will allow the application to ignore `WWW-Authenticate` headers that are not for basic auth instead of simply giving up.

### DIFF
--- a/Simplified/NYPLBasicAuth.m
+++ b/Simplified/NYPLBasicAuth.m
@@ -7,24 +7,12 @@ void NYPLBasicAuthHandler(NSURLAuthenticationChallenge *const challenge,
                           (NSURLSessionAuthChallengeDisposition disposition,
                            NSURLCredential *credential))
 {
-  if(![challenge.protectionSpace.authenticationMethod isEqualToString:NSURLAuthenticationMethodHTTPBasic]) {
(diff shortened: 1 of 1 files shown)


Simplified/NYPLBasicAuth.m
```diff
@@ -34,6 +39,11 @@ void NYPLBasicAuthCustomHandler(NSURLAuthenticationChallenge *challenge,
     @throw NSInvalidArgumentException;
   }
   
+  if(![challenge.protectionSpace.authenticationMethod isEqualToString:NSURLAuthenticationMethodHTTPBasic]) {
+    completionHandler(NSURLSessionAuthChallengeRejectProtectionSpace, nil);
```
aslagle: testing

Simplified/NYPLBasicAuth.m
```diff
@@ -34,6 +39,11 @@ void NYPLBasicAuthCustomHandler(NSURLAuthenticationChallenge *challenge,
...
+    return;
+  }
+
```
aslagle: sdfsdfsdf

winniequinn:
Agreed! Duplication removed!

PR #463, opened 2016-10-03
Doubled timeout interval for the first page in catalog feeds

### LINKED ISSUE #462: App often shows Connection Failed for feeds
After some research, this seems to simply be due to the connection timing out. The timeout should be increased appropriately.

winniequinn:
Closes #462 

@aferditamuriqi


aferditamuriqi:
👍

PR #477, opened 2016-10-26
added circulation analytics with first event open_book

### LINKED ISSUE #468: Implement Circulation Analytics
- online/offline queue to send events to the analytics endpoint
   - first event we will be sending is open_book

gioneill:
fixes #468

### DIFF
--- a/Simplified/NYPLCirculationAnalytics.swift
+++ b/Simplified/NYPLCirculationAnalytics.swift
@@ -6,7 +6,8 @@ final class NYPLCirculationAnalytics : NSObject {
   
   class func postEvent(event: String, withBook book: NYPLBook) -> Void {
     
-    guard let requestURL = NSURL(string: book.analyticsURL.absoluteString.stringByAppendingString("/\(event)")) else { return }
+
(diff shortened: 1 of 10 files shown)


Simplified/SimplyE-Bridging-Header.h
```diff
@@ -7,3 +7,4 @@
 #import "NYPLOPDSLink.h"
 #import "NYPLBook.h"
 #import "NYPLAccount.h"
+#import "NYPLSession.h"
```
aferditamuriqi: not sure if we need session here

Simplified/NYPLCirculationAnalytics.swift
```diff
@@ -0,0 +1,35 @@
...
+       (request, response, data, error)  in
+      
+      if (error != nil) || (response?.statusCode != 200) {
+        // Error posting event
+      }
+      
+    }
+  }
+  
+  // Server currently not validating authentication in header, but including
```
aferditamuriqi: will move this out of this class, for now it's ok to keep it here.

PR #500, opened 2016-11-07
Settings license update

gioneill:
1. Updated rel link to "license"
2. Fixed a bug that was showing the "connection failed" UIAlert when it should not have.
3. Updated HTML files to fix some text formatting.


winniequinn:
@gioneill What do you think about having local files for things that can be local and going over the network with no fallback for everything else? I'm not really sure on the justification for showing possibly out-of-date policies and licenses. I think it's better to just say "try again when you have a network connection" than to show something that's incorrect.

gioneill:
The work in the PR has been absorbed into the multi_library_develop branch.
In response to the comment, the only locally-pulled html file is now the "Software Licenses." The rest have been updated to go over the network with no fallback.

PR #507, opened 2016-11-22
Feature/offline analytics lastread

jrryhrtn:
I believe the changes are ready for review and integration.  I'm unable to run unit tests on the master branch, but I have created some on my fork under swift3.0 branch (which you are welcome to review).

### DIFF
--- a/Podfile
+++ b/Podfile
@@ -13,6 +13,9 @@ target 'SimplyE' do
 	pod 'CHRTextFieldFormatter'
   pod "NYPLCardCreator", :git => "https://github.com/NYPL-Simplified/CardCreator-iOS.git"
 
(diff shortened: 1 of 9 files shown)


gioneill:
Hi Jerry thanks for your help on this. My only question is if it would be preferable to use the Reachability built into Alamofire to save having to add a (admittedly small) dependency?

jrryhrtn:
Yes, I agree.  I noticed that AlamoFire had integrated Reachability after the fact.   I was planning on making the change, got distracted on Android.  I can push the change to this branch by Monday morning, if you can hold off merging until then.

jrryhrtn:
I updated feature/offline-analytics-lastread, removing extra 3rd party Reachability module, and replaced with AlamoFire's imply

PR #509, opened 2016-11-29
Swift3.0

jrryhrtn:
Converted entire project to swift 3.0.  Also, the pod directory is now committed to the repo per the latest cocoa pod recomendations:  https://guides.cocoapods.org/using/using-cocoapods.html

The Podfile version has not been updated to pull the swift 3.0 version of card creator, but the swift 3.0 updates have been performed on the committed source of the card creator in the local Pods folder.  You will need to update appropriately, and do a pod update.  

Unit tests have also been updated to 3.0, and should build/run as expected.  This seemed to be broken prior to the swift 3.0 update.


winniequinn:
@jrryhrtn We do not commit the `Pods` directory for our projects. Would you be able to update this pull request so that the `Pods` directory is not included?

jrryhrtn:
removed Pods folder from branch, and update .gitignore to ignore the Pods folder (like it used to be).  Please note, the current card creator pod referenced in the Podfile is not Swift 3.0 compliant, and results in 124 compilation errors.  These were fixed in the locally committed Pod folder, as a temp fix until the pod source was updated.

winniequinn:
Thanks, @jrryhrtn! I am away for the next several weeks but @aferdita may be able to review this before then.

aferditamuriqi:
@winniequinn yes i will be reviewing, and @gioneill as well, since he has the card creator updated to Swift 3.0 already.

PR #550, opened 2017-02-06
Offline queue #494

### LINKED ISSUE #494: Offline queue for circulation analytics and annotations

gioneill:
resolves #494 

Includes:
- Offline URLSession Queue
- Updates to Annotations JSON (both sync and post)
- Last-Read Sync (switch enabled in accounts)

Test Cases for Queue

- When a network request fails due to a relevant connectivity error, add it to the queue
- The queue should be able to serialize any type of URLSession we would potentially use
- When queueing a request, if it's based on "last-read sync", it should replace the last entry for that same book
- When reachability detects an internet connection, the queue should be retried (if any entries are queued).
- If any particular request has been retried more than X times, remove that request from the queue.
- If a request is retried and fails again (any reason the request may fail while having a valid connection) leave it in the queue, but increment retry count.
- If a request is retried and succeeds with 200 from the server, remove it from the queue.
- For circulation analytics, all instances should be queued

Design Goals for Queue

- Stay away from needing an instance of the class.
- SQLite.swift has a Swifty API as well as a straight wrapper to execute SQL statements
- Reachability listener added to NYPLReachability (shoots off queue-retry when internet connection returns). Its singleton is now instantiated and stored as a property in NYPLAppDelegate. (Before it was lazily instantiated during a user sign out).


Simplified/NYPLNetworkQueue.swift
```diff
@@ -0,0 +1,173 @@
...
+  {
+    if (Int(requestRow[sqlRetries]) > QueueMaxRetryCount) {
+      deleteRow(db: db, id: Int(requestRow[sqlID]))
+      Log.info(#file, "Removing after \(Int(requestRow[sqlRetries])) retries")
+      return
+    }
+    
+    do {
+      let ID = Int(requestRow[sqlID])
+      let newValue = Int(requestRow[sqlRetries]) + 1
```
winniequinn: It seems like it'll be really easy to max out the allowed retries if the user is offline or has an intermittent connection or if the server goes down for awhile. While I do agree that events that are never going to send should be dropped from the queue eventually, I think another approach is needed here. Perhaps we should only be removing events when the server has returned a particular status code indicating malformation or similar? Maybe that plus a limit in terms of time rather than attempts?

gioneill: To your first point, the queue should only be retrying when NYPLReachability sends a notification that the host is reachable.
I do like adding some kind of time limit though. Something like 30 days? I'm afraid of having a situation where our filter is too narrow which allows requests to queue so large that you have a growing N number of requests trying to send and failing for some unknown reason on each retry.

Simplified/NYPLAnnotations.swift
```diff
@@ -12,177 +12,158 @@ final class NYPLAnnotations: NSObject {
...
   {
-
-    //FIXME: this network code needs to be tested before being re-implemented. Alamofire was removed. Remove this comment after turning annotations/sync back on.
+    let jsonData: Data?
+    do {
+      jsonData = try JSONSerialization.data(withJSONObject: parameters, options: [.prettyPrinted])
+    } catch {
+      Log.error(#file, "Network request abandoned. Could not create JSON from given parameters.")
+      return
+    }
```
gioneill: I'm actually curious: are you saying that jsonData could be nil because the function may not throw an error but still assign nil? I was thinking I was equally safe with a do-catch block.

winniequinn: Not quite. I'm saying that due to the way you've written in, the type system will not understand `jsonData` is non-nil because you've declared it as type `Data?`. If you use the `guard let` form instead with `try?` to convert exceptions to `nil`, you can assign `jsonData` the non-nil type `Data`. It happens not to matter in this case because all uses of `jsonData` accept `nil`, but it might make the code a bit clearer.

Simplified/NYPLAnnotations.swift
```diff
@@ -12,177 +12,158 @@ final class NYPLAnnotations: NSObject {
...
+            do {
+              jsonData = try JSONSerialization.jsonObject(with: data!, options: []) as? [String:Any]
+            } catch {
+              Log.error(#file, "JSON could not be created from data.")
+              completionHandler(nil, nil)
+              return
+            }
+            
+            if let json = jsonData {
+              let total:Int = json["total"] as! Int
```
gioneill: I realize the Git Comparison may be out of whack here since this was pasted from another branch, but this class was outside the scope of this PR. Since it's linked to Last-Read Sync which is pretty fragile to adjustments at the moment, I'll defer to Aferdita for changes to this area.

Simplified/NYPLNetworkQueue.swift
winniequinn: > To your first point, the queue should only be retrying when NYPLReachability sends a notification that the host is reachable.

Ah, okay! Perhaps were good as-is then. 

> I do like adding some kind of time limit though. Something like 30 days?

I think that might be okay! Perhaps something like "discard after **at least** M days **and at least** N failures" would cover both cases? I wouldn't want something to be dropped because it was tried three times during an hour of a temporary server outage and I also wouldn't want something to be dropped just because the user's device has been off for a month and we never got to try again.

Simplified/NYPLNetworkQueue.swift
```diff
@@ -0,0 +1,173 @@
...
+  private static let sqlUrl = Expression<String>("requestUrl")
+  private static let sqlMethod = Expression<String>("requestMethod")
+  private static let sqlParameters = Expression<Data?>("requestParameters")
+  private static let sqlHeader = Expression<Data?>("requestHeader")
+  private static let sqlRetries = Expression<Int>("retryCount")
+  
+  
+  // MARK: - Public Functions
+  
+  class func addRequest(_ libraryID: Int,
```
gioneill: On whether it makes more sense to use credentials at "time it was queued" versus the "time of the attempt", is it an equal risk that the user is no longer on the same library account relevant to that annotation?

Simplified/NYPLAnnotations.swift
winniequinn: Ah, strange! I'm not sure how that happened.

Simplified/NYPLNetworkQueue.swift
winniequinn: I think they're both risks, yes! I suspect is more likely that a PIN will change that an account will be closed though for whatever that's worth. We should be able to reasonably handle both cases regardless.

Simplified/NYPLAnnotations.swift
```diff
@@ -12,177 +12,158 @@ final class NYPLAnnotations: NSObject {
...
+        
+        for (headerKey, headerValue) in NYPLAnnotations.headers {
+          request.setValue(headerValue, forHTTPHeaderField: headerKey)
+        }
+        
+        let dataTask = URLSession.shared.dataTask(with: request) { (data, response, error) in
+          
+          if error != nil {
+            completionHandler(nil, error as? NSError)
+          } else {
```
aferditamuriqi: @winniequinn yes a return should be called actually after any completion handler. looking at the code i am also removing the error, as it is not really needed so the completion handler will just have the result or nil

Simplified/NYPLAnnotations.swift
```diff
@@ -12,177 +12,158 @@ final class NYPLAnnotations: NSObject {
...
+            
+            let jsonData: [String:Any]?
+            do {
+              jsonData = try JSONSerialization.jsonObject(with: data!, options: []) as? [String:Any]
```
aferditamuriqi: @winniequinn you're right makes sense.

Simplified/NYPLAnnotations.swift
```diff
@@ -12,177 +12,158 @@ final class NYPLAnnotations: NSObject {
...
+            } catch {
+              Log.error(#file, "JSON could not be created from data.")
+              completionHandler(nil, nil)
```
aferditamuriqi: getting rid of the actual error. as it wasn't being used, but if at some later point it makes sense to make use of such an error, we can rethink this.

Simplified/NYPLAnnotations.swift
```diff
@@ -12,177 +12,158 @@ final class NYPLAnnotations: NSObject {
...
+    return ["Authorization" : "\(authenticationValue)",
+            "Content-Type" : "application/json"]
+  }
+  
+  class func iso8601CurrentTime() -> String
+  {
+    let dateFormatter = DateFormatter()
+    let enUSPosixLocale = Locale(identifier: "en_US_POSIX")
+    dateFormatter.locale = enUSPosixLocale
+    dateFormatter.dateFormat = "yyyy-MM-dd'T'HH:mm:ssZZZZZ"
```
aferditamuriqi: if the result is the same, i am ok with using  RFC3339String, we just need to make sure it is the same on both iOS and Android

Simplified/NYPLNetworkQueue.swift
```diff
@@ -0,0 +1,173 @@
...
+                               NSURLErrorInternationalRoamingOff,
+                               NSURLErrorCallIsActive,
+                               NSURLErrorDataNotAllowed,
+                               NSURLErrorSecureConnectionFailed]
+let QueueMaxRetryCount = 3
+
+enum HTTPMethodType: String {
+  case GET, POST, HEAD, PUT, DELETE, OPTIONS, CONNECT
+}
+
```
aferditamuriqi: i see this has already been moved :)

Simplified/NYPLNetworkQueue.swift
aferditamuriqi: I think i am ok with the reties. if a users device has been off for a month and that device has un-posted reading positions, i would think that user won't be able to read that book anyways, that book might have expired already, and for numerous other reasons we should be good with the retries, and in case of there is a server outage for an hour or so, as far as i understood, that is not a reason to increase the retry, (correct me if i am wrong)

PR #568, opened 2017-03-19
Fix xcodeproj -> project warning in Podfile

winniequinn:
Cocoapods issued the following warning on `pod install`:

    [!] `xcodeproj` was renamed to `project`. Please update your
    Podfile accordingly.

This fix simply implements the above suggestion.

---

If everything looks good, please feel free to merge!

### DIFF
--- a/Podfile
+++ b/Podfile
@@ -1,4 +1,4 @@
-xcodeproj 'Simplified.xcodeproj'
+project 'Simplified.xcodeproj'
 
 # Uncomment this line to define a global platform for your project
 platform :ios, '8.0'


gioneill:
I also found that I needed to update CocoaPods on my machine on Friday. Was that just a coincidence?

winniequinn:
@gioneill That will happen if "Podfile.lock" requires a newer version than what you have installed:

```
$ tail -n 1 Podfile.lock
COCOAPODS: 1.2.0
```

It seems like "Podfile.lock" is actually not checked in at all for `master` at the moment. I'm also having a build issue were Bugsnag cannot be found. I'll see if I can fix these things and get another pull request up shortly!

winniequinn:
(The Bugsnag issue was just silliness on my end, it seems!)

PR #571, opened 2017-03-19
Make .xcconfig files understandable to CocoaPods

winniequinn:
## Build-breaking change!

 Merging this into `master` will break builds for people who are currently building without DRM. To fix their builds, they will simply need to comment out the last line of "Simplified.xcconfig". This change has been reflected in "README.md" and is justified below.

**Please notify everyone on Slack before merging this change.**

If this breakage is unacceptable, let me know and I will attempt another, more complicated solution to this problem. I would prefer to go this route for now though and then take another pass at it later so that the app can build cleanly without DRM by default.

(On the plus side, merging this would remove the last of our CocoaPods warnings! Hurray!)

---

Prior to this commit, we had four `*.xcconfig` files:

* `Simplified.debug.xcconfig`, which just included `Simplified.xcconfig` and `Simplified+RMSDK.xcconfig`;

* `Simplified.release.xcconfig`, which was identical to `Simplified.debug.xcconfig`;

* `Simplified.xcconfig`, which contained the bulk of the project configuration;

* `Simplified+RMSDK.xcconfig`, which included headers for DRM-related libraries and set `FEATURE_DRM_CONNECTOR=1`.

The project was set to use the `Simplified.debug.xcconfig` and `Simplified.release.xcconfig` configurations for debug and release respectfully. The SimplyE target _within_ debug and release was set to use `Simplified+RMSDK.xcconfig`, however.

It was the later portion that caused CocoaPods to issue the following warnings:

    [!] CocoaPods did not set the base configuration of your project
    because your project already has a custom config set. In order for
    CocoaPods integration to work at all, please either set the base
    configurations of the target `SimplyE` to `Pods/Target Support
    Files/Pods-SimplyE/Pods-SimplyE.debug.xcconfig` or include the
    `Pods/Target Support Files/Pods-SimplyE/Pods-SimplyE.debug.xcconfig`
    in your build configuration (`Simplified+RMSDK.xcconfig`).

    [!] CocoaPods did not set the base configuration of your project
    because your project already has a custom config set. In order for
    CocoaPods integration to work at all, please either set the base
    configurations of the target `SimplyE` to `Pods/Target Support
    Files/Pods-SimplyE/Pods-SimplyE.release.xcconfig` or include the
    `Pods/Target Support
    Files/Pods-SimplyE/Pods-SimplyE.release.xcconfig` in your build
    configuration (`Simplified+RMSDK.xcconfig`).

The following fixed all of the above:

* I moved all configuration into `Simplified.xcconfig`.

* I set the project to use `Simplified.xcconfig` for both debug and release.

* I removed the superfluous `Simplified.debug.xcconfig` and `Simplified.release.xcconfig`.

* In the project settings, I removed the configuration setting for the target _within_ debug and release. This allowed CocoaPods to set it itself which resolved the warnings above.

This commit does not resolve long-standing issues in which DRM-free builds are not particularly easy, but it should not make them any worse either. I updated "README.md" to reflect the change: You must now comment out the last line of "Simplified.xcconfig" instead of deleting "Simplified+RMSDK.xcconfig".

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,7 +26,7 @@
 06. `rm -rf adept-ios adobe-content-filter`
 07. `cp APIKeys.swift.example Simplified/APIKeys.swift` and edit accordingly.
 08. `open Simplified.xcworkspace`
-09. Remove "Simplified+RMSDK.xcconfig" from the project.
+09. Comment out the last line of "Simplified.xcconfig" that sets `FEATURE_DRM_CONNECTOR=1`.
 10. Delete "libAdept.a" and "libAdobe Content Filter.a" from "Link Binary with Libraries" for the "SimplyE" target.
 11. Build.
 
--- a/Simplified+RMSDK.xcconfig
+++ b/Simplified+RMSDK.xcconfig
@@ -1,4 +1,0 @@
-
-GCC_PREPROCESSOR_DEFINITIONS = $(inherited) FEATURE_DRM_CONNECTOR=1
-
-HEADER_SEARCH_PATHS = $(inherited) "$(SRCROOT)/adept-ios" "$(SRCROOT)/adobe-content-filter/public"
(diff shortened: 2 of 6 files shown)


winniequinn:
I came up with a better way to do this!

PR #590, opened 2017-03-30
Increased font sizes. Adjusted auto layout for 4" screens.

gioneill:


winniequinn:
@gioneill Instead of using a fixed magic number for the row height, is it possible to make use of `UITableViewAutomaticDimension` here instead? You can check the card creator for many examples of its use.

gioneill:
I've been slowly moving a lot of the table views in the app to dynamic row heights in anticipation of adding DynamicText for accessibility. This VC is a holdout until that task begins because dynamic height won't work on default cell styles (in this case subtitled) or cells not created with a prototyped .xib in interface builder.

winniequinn:
@gioneill Got it! Thanks for explaining your rationale. ^_^

PR #596, opened 2017-04-03
Eliminated double-UIAlert for login errors. Allow user to dismiss at …

### LINKED ISSUE #595: Card Creator Dismissal UX Changes
After creating a card, the user should have a choice to dismiss the screen when reviewing their credential/barcode details. When signing in after creating a card after attempting to check out a book, both screens should dismiss. Only one error should display when there is a sign-in issue.

Only changes in iOS-Simplified are required.

gioneill:
fixes #595


gioneill:
Changed base branch to 'develop'

PR #598, opened 2017-04-05
Issues with UX under slow networks

### LINKED ISSUE #182: Verify correct timeout when retrieving search results

gioneill:
Addresses a number of issues have been addressed that should overall significantly improve the user-experience under various "slow or inconsistent (but still usable)" network connections.

fixes #182 plus other slow-network issues:

- Fixed "failed connection" on first load of catalog
- Fixed book covers failing to load
- Extended timeout duration
- Added user-messaging on extra-long loads

Detail:
- Often times (even under adequate connection quality in the 1-5 mbps range) the initial load of the catalog would present a "connection failed: refresh" error as the result of a timeout. The catalog would load successfully almost 100% of the time on the second try. This issue has been fixed.
- The downloading of the lanes and book covers happens in several steps, and often times the book cover downloads would timeout prematurely and show the auto-generated covers unnecessarily. This issue has been fixed.
- For very large books that are failing mid-download, the errors I'm getting when constrained to less than 500kbps is coming from Adept and may not be directly related just to the timeout (otherwise the downloads take several minutes but are successful at slightly faster speeds). Another issue will be created to directly investigate this.
- For slow (but not impossible) connections less than 1 mbps, the app is usable but can take upwards of a minute to load the catalog and download books, I've introduced some extra user-messaging to assure the user the download is continuing successfully. It will only phase in as a download takes in excess of 10 seconds.


gioneill:
Fades in under activity indicator after 10 seconds.
simulator screen shot apr 5 2017 1 00 01 pm copy

aferditamuriqi:
is "downloading"  the right message here? we are not really downloading the collection, just "Loading" would probably be better.

gioneill:
Changed base branch to 'develop'

PR #602, opened 2017-04-14
added old keys to text size in reader settings to prevent crash

gioneill:
fixes crash by re-adding old keys from text-size in reader settings


aferditamuriqi:
submitting a hot fix to apple.

PR #616, opened 2017-04-23
Fix leaking of NYPLReaderReadiumView

winniequinn:
This fixes a MAJOR memory leak and a major hog of CPU. The app was never releasing `NYPLReaderReadiumView` instances thus causing WebKit processes to pile up. We should get this into a release ASAP.


gioneill:
I discussed with @winniequinn and I approve the changes.

The property in Reader Settings is an unnecessary strong reference.

The polling is definitely continually holding each instance as you open more and more books. It was discovered by the new Memory Graph Debugger in Xcode. It was difficult to see during normal Instrumenting because the memory and CPU spikes were being split over non-app processes in activity monitoring!

This could also explain why the system would have trouble just terminating our app versus hitting the memory limit and hard-resetting the device. It prevented the app from seeing any memory warnings or being able to log the crashes.

PR #617, opened 2017-04-24
Allow external links to open in Safari

winniequinn:


gioneill:
Does this hurt the ability to navigate links inside the book? Using endnotes or the table of contents to reach further chapters, for example.

winniequinn:
@gioneill No, it does not. That delegate method is only called when the web view wants to open something in a new web view for whatever reason (e.g. a link with `target="_new"`). It'll never be called for a link to something else in the book.

winniequinn:
@aferditamuriqi + @gioneill: I fixed a bug in how this was implemented and added some comments so it'll be clearer to the next person. It's been tested heavily. I'm merging!

Issue #605, opened 2017-04-20
Implement local bookmarking


aferditamuriqi:
this should work on device only, server sync component will be addressed in a separate issue

referencing https://github.com/NYPL-Simplified/Simplified/issues/23

aferditamuriqi:
Page numbering needs to be addressed still. currently showing Content CFI.

Chapter title also needs some refinement, as of now if there is no chapter name, we need to figure out what information to display.

Excerpt has been left out for now, but maybe that is something which can be looked at when chapter name is being addressed.

aferditamuriqi:
There seems to be one issue, which needs testing and steps to reproduce:

when a book is downloaded the first time, bookmarks seem to not  persist, only after relaunching the book then bookmarks seems to works as expected.

aferditamuriqi:
closing this with PR https://github.com/NYPL-Simplified/Simplified-iOS/pull/624

PR #624, opened 2017-05-02
Local bookmarks #605

aferditamuriqi:
adjusted number of lines in storyboard and fixed initial value of bookmarks


aferditamuriqi:
commit message was supposed to say "fixed initial bookmarks value. bookmarks can’t be nilable." somehow it was auto corrected to "notable"

PR #632, opened 2017-05-23
Translations updated (english placeholder)

### LINKED ISSUE #621: SettingsAccountAgeCheckbox error not localized
SettingsAccountAgeCheckbox  When not checked the app throws an error for folsk previously logged in but who did not check this box.

gioneill:
fixes #621

### DIFF
--- a/Simplified/NYPLBookDetailNormalView.m
+++ b/Simplified/NYPLBookDetailNormalView.m
@@ -70,7 +70,7 @@ typedef NS_ENUM (NSInteger, NYPLProblemReportButtonState) {
 
 - (NSString *) reportButtonTitleForCurrentSize
 {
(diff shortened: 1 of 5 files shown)


gioneill:
I realize now the binary files can't be diff'ed, but I essentially found all the key/values present in English that were not present in the other languages, and copied them in. This prevents Keys from displaying for all those strings in the other languages.

aferditamuriqi:
@gioneill this should be going through transifex, and we shouldn't have english text in other languages. in general if the key is not present in other languages, the OS would default back to the default language, which I believe is set to english,

gioneill:
The OS is not defaulting to English though.. we either need to keep English versions in there temporarily, OR go back to the source of all those strings (30+) and change the keys to the english values.

simulator screen shot may 23 2017 11 08 21 am

aferditamuriqi:
there is definitely something not right there, but i would not prefer having english placeholders. let me take a look and i'll get back to you.

aferditamuriqi:
as suspected we are not using base internationalization. i'll send you some notes in slack.

PR #636, opened 2017-06-03
Localizable fallback language #621

aferditamuriqi:
the solution in this PR is a custom LocalizedString  called NYPLLocalizedString, which is used instead of NSLocalizedString. it solves the problem by using en as the fallback language, of course it's only one several other possible solution. adding my slack comment here, as reference: 

The localizable strings turn out to need a bit more than expected, i am creating a branch with the macro i have written, and there is a bit more you'd need to do, i'll explain once i check in the branch. here is the commit https://github.com/NYPL-Simplified/Simplified-iOS/commit/79f89a8c606c3b63ec83fc8452a77c3e05ad65dd

ideally this should go into the .pch, and it would use the fallback language everywhere where NSLocalizedString is being used. 
#import "NYPLLocalizedStringFallbackLanguage.h" 
it's working well for swift added in the bridging header

it works very well, when added in every objective-c header where NSLocalizedString is being used, but not when added into the .pch.
maybe i missed something, but i don't have a lot of time to spend on this...

case 1: ideally: make it work in  .pch 
case 2: add header to all objective classes where needed (about 35)
case 3: replace NSLocalizedString with NSLocalizedStringFromTableInBundle (would require changing all the 35 classes as well) 
case 4: placeholder english in other localized strings (not recommending this, as this will be forgotten about to fix later)

in either case we need to generate the localized.strings in transifex

NSLocalizedString basically shows the key itself when there is no key:value available in the language selected, there is no fallback by design. (i was pretty sure there was though)


gioneill:
Looks good now. What is strange is that the Card Creator **does** work in the way we thought the project should work: 
If a key has not yet been translated, then it shows the English text. So I'm not sure what we're doing differently in the main app with NSLocalizedString..

But at least in the future there will only be once place to make changes if we find an improvement: in StringUtilities.h

winniequinn:
I agree that something does seem strange, @gioneill. I'd like to look this over some time tomorrow before we go ahead and merge. If nothing else, I'd like to understand the new solution going into place.

aferditamuriqi:
I believe the card creator works because the card creator is not localized yet, so all is in English,  or am I wrong?

aferditamuriqi:
in iOS as I have explained in slack and also above , there is simply no way to use NSLocalizedString with a fallback language, it is required to have all keys translated or not use keys at all but use the English values as keys.  there is no perfect or right or wrong solution for this. we just have to agree on one solution, for which I spent quite some time to figure out, and my solution would be this one. 

ideally I would have used the macro and continue using NSLocalizedString but that didn't work and doesn't work.  I am of course open to any better solution if there is any.

winniequinn:
@aferditamuriqi 

Card creator has not yet been localized to my knowledge, but it is ready for localization.

Unlike the main app, card creator uses English strings that can be presented directly as keys. Comments are provided for additional context. This is, as far as I know, Apple's recommend way of using `NSLocalizedString`. For example:

```swift
self.firstNameCell = LabelledTextViewCell(
  title: NSLocalizedString(
    "First Name",
    comment: "The text field title for the first name of a user"),
  placeholder: NSLocalizedString(
    "Required",
    comment: "A placeholder for a required text field"))
```

One nice thing about this approach is that it effectively gives you English as a fallback language! Since the app will show the key if there is no localized string for a given key/language combination, everything just works.

aferditamuriqi:
@winniequinn I have as well mentioned apple's recommendation and several solutions in several comments and in slack. i provided one solution given the fact that iOS main project didn't follow through with one or the other recommendation. feel free to discard my solution and apply any changes necessary,  side note: card creator does not have a localized strings file at all, at least i am not seeing it anywhere.

Simplified/StringUtilities.m
```diff
@@ -0,0 +1,26 @@
...
+//  StringUtilities.m
+//  Simplified
+//
+//  Created by Aferdita Muriqi on 6/20/17.
+//  Copyright © 2017 NYPL Labs. All rights reserved.
+//
+
+#import "StringUtilities.h"
+
+@implementation StringUtils
```
winniequinn: Since this is an Objective-C class that's part of the app itself, we should be sticking with the established `NYPL___` naming convention, e.g. `NYPLStringUtilities`.

Simplified/StringUtilities.m
```diff
@@ -0,0 +1,26 @@
...
+
+NSString * NYPLLocalizedString(NSString * key, NSString * comment) {
```
winniequinn: Despite my other comment, it seems we do not actually need a class here at all! The only thing contained in `StringUtils` is a top-level function, not a method, and thus we should probably just define the function at the top level and dispatch with the class altogether.

Simplified/StringUtilities.h
```diff
@@ -0,0 +1,15 @@
+//
+//  StringUtilities.h
+//  Simplified
+//
+//  Created by Aferdita Muriqi on 6/20/17.
+//  Copyright © 2017 NYPL Labs. All rights reserved.
+//
+
+#import <Foundation/Foundation.h>
```
winniequinn: For imports, we should be using the newer style nowadays as we are elsewhere in the project:

```objective-c
@import Foundation;
````

File: Simplified/StringUtilities.m (hunk 2 above)
Comment (winniequinn) [inline Simplified/StringUtilities.m]: This is a minor style point, but I would attach the `*` to the identifier it modifies:

```objective-c
NSString *NYPLLocalizedString(NSString *key, NSString *comment)
```

This convention is used throughout the project as far as I know. The reason for this convention is that the `*` modifies the name, not the type. For example, the following declaration declares `x` as a pointer to an integer yet it declares `y` merely as an integer:

```c
int *x, y;
```

Simplified/Simplified-Info.plist
```diff
@@ -23,7 +23,7 @@
 	<key>CFBundleSpokenName</key>
 	<string>Simply E</string>
 	<key>CFBundleVersion</key>
-	<string>1087</string>
+	<string>1385</string>
```
winniequinn: We should try not to commit build number changes. If everyone does so, we'll always have a merge conflict when multiple developers are merging into the same branch.

winniequinn:
I'll go ahead and pull this branch in after making the above-mentioned changes.

aferditamuriqi:
Please discard this PR . I closed it on purpose !

aferditamuriqi:
this pull request was closed, as i have decided i am not 100% happy about the solution, it is a solution i use in other of my projects, it is not ideal, this is my personal contribution, and this would have solved an immediate problem. but on second thought, i decided the changes made are maybe in too many files and this should be given another thought. 

please don't reopen until a perfect solution can be agreed on.

PR #638, opened 2017-06-09
Add alameda to library list

datalogics-tarar:
Added Alameda county library in California to the list of libraries in the SimplyE app along with their logos.

PR #639, opened 2017-06-09
Allow barcode fields to take letters

datalogics-tarar:
Not all pins are strictly numeric and we need to allow the pin fields to accommodate that. This is already supported in Android.


gioneill:
The existing keyboards are a positive convenience for users, especially the numeric keypad. I think it makes sense to add attributes to each library in the JSON file, so that if a library requires a barcode or PIN to take letters, that only those libraries need to use more generic alphanumeric keyboards without removing it from all the other libraries.

As an example, look at how we check for the Card Creator Registration in those same classes and implement something similar for the type of keyboard to display. Something like "alphanumericPasswordRequired"

gioneill:
These changes are now part of the main PR to add Alameda County:
https://github.com/NYPL-Simplified/Simplified-iOS/pull/638

PR #641, opened 2017-06-15
Creator bug #595

### LINKED ISSUE #595: Card Creator Dismissal UX Changes

gioneill:
Related to bug 14 in Card Creator.
https://github.com/NYPL-Simplified/CardCreator-iOS/issues/14

Fixes #595 

Allows the final confirmation screen to remain until the user chooses to dismiss it. This allows them to view and copy their newly-created credentials.

### DIFF
--- a/Simplified/NYPLAccountSignInViewController.m
+++ b/Simplified/NYPLAccountSignInViewController.m
@@ -715,6 +715,8 @@ replacementString:(NSString *)string
                          NSURLResponse *const response,
                          NSError *const error) {
        
+       [[UIApplication sharedApplication] setNetworkActivityIndicatorVisible:NO];
+       
        self.isCurrentlySigningIn = NO;
        // This cast is always valid according to Apple's documentation for NSHTTPURLResponse.
        NSInteger const statusCode = ((NSHTTPURLResponse *) response).statusCode;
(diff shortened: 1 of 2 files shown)


Simplified/NYPLAccountSignInViewController.m
```diff
@@ -715,14 +715,15 @@ - (void)validateCredentials
                          NSURLResponse *const response,
                          NSError *const error) {
        
+       [[UIApplication sharedApplication] setNetworkActivityIndicatorVisible:NO];
```
winniequinn: I am not sure where else we are using `setNetworkActivityIndicatorVisible:` in the app, but calling that method directly is generally not a good idea. The problem is that multiple parts of the app can toggle the activity indicator at the same time and step on each other in the process.

If you want to make use of the activity indicator, I'd suggest wrapping up calls to `setNetworkActivityIndicatorVisible:` with a new method that manages a private counter. When passing `YES`, turn the indicator on and increment the counter. When passing `NO`, decrement the count and turn the indicator off if the counter is equal to 0. If `NO` is passed when the counter is already 0, leave the counter at 0 and log an error.

gioneill: I'm removing this for now and creating a related ticket to implement this.

gioneill:
I'm opening a new PR based on a branch of Master.

PR #642, opened 2017-06-16
Allow unlimited length pin

datalogics-tarar:


gioneill:
These changes are now a part of the main PR:
https://github.com/NYPL-Simplified/Simplified-iOS/pull/638

PR #643, opened 2017-06-19
Book Detail View

gioneill:
Updated views, auto layout support, dynamic type, recommended and related works lanes.


Simplified/NYPLBook.h
```diff
@@ -19,6 +14,7 @@ typedef NS_ENUM(NSInteger, NYPLBookAvailabilityStatus) {
 
 @property (nonatomic, readonly) NYPLBookAcquisition *acquisition;
 @property (nonatomic, readonly) NSString *authors;
+@property (nonatomic, readonly) NSArray *authorLinks;
```
winniequinn: Would it be possible to use generics here? I know the class doesn't elsewhere yet but there should be no harm in starting.

winniequinn: One of the things `NYPLBook` is meant to do is provide information in a very concrete form in terms of how it will be later used by the application. I think it might make sense to have a separate `NYPLBookAuthor` data structure here that very precisely contains what we need instead of including links that require knowledge of `NYPLOPDS*` classes to use.

Simplified/NYPLBook.m
```diff
@@ -334,22 +368,16 @@ - (instancetype)initWithDictionary:(NSDictionary *)dictionary
...
   return self;
 }
 
 - (NSDictionary *)dictionaryRepresentation
 {
   return @{AcquisitionKey: [self.acquisition dictionaryRepresentation],
+           AlternateURLKey: NYPLNullFromNil([self.alternateURL absoluteString]),
+           AnnotationsURLKey: NYPLNullFromNil([self.annotationsURL absoluteString]),
+           AnalyticsURLKey: NYPLNullFromNil([self.analyticsURL absoluteString]),
+           AuthorLinksKey: self.authorLinks,
```
winniequinn: `self.authorLinks` is an `NSArray` of `NYPLOPDSLink` objects. The purpose of `dictionaryRepresentation` is to return an `NYPLBook` that is JSON-serializable. I'm not sure how this can work since `NYPLOPDSLink` is not one of the Cocoa classes serializable by `NSJSONSerialization`. What am I missing here?

Simplified/NYPLBookCoverRegistry.m
```diff
@@ -85,6 +85,29 @@ - (NSURL *)URLForPinnedThumbnailImageOfBookIdentifier:(NSString *const)bookIdent
           URLByAppendingPathComponent:[bookIdentifier SHA256]];
 }
 
+- (void)coverImageForBook:(NYPLBook *)book handler:(void (^)(UIImage *image))handler
+{
+
+  //Thumbnail first as placeholder
+  [self thumbnailImageForBook:book handler:handler];
```
winniequinn: This doesn't seem right to me. Given that `handler` will often have side effects, I think it would be very surprising for any user of this method to have their handler called multiple times, first with a thumbnail and later with a cover image.

Simplified/NYPLBookCoverRegistry.m
```diff
@@ -85,6 +85,29 @@ - (NSURL *)URLForPinnedThumbnailImageOfBookIdentifier:(NSString *const)bookIdent
...
+
+  [[self.session
+    dataTaskWithRequest:[NSURLRequest requestWithURL:book.imageURL]
+    completionHandler:^(NSData *const data,
+                        __attribute__((unused)) NSURLResponse *response,
+                        __attribute__((unused)) NSError *error) {
+      [[NSOperationQueue mainQueue] addOperationWithBlock:^{
```
winniequinn: It doesn't seem like this method of fetching a cover image is going to work offline. The app has a notion of "pinned" images that have been saved for offline use, and that notion of pinning is already in use for thumbnails. I would look at `thumbnailImageForBook:handler` to see how you could handle full-size cover images offline as well.

Simplified/NYPLBookDetailButtonsView.h
```diff
@@ -0,0 +1,26 @@
...
+(NYPLBookDetailButtonsView *)bookDetailButtonsView;
+
+- (void)didSelectCancelForBookDetailDownloadFailedView:
+(NYPLBookDetailButtonsView *)NYPLBookDetailDownloadFailedView;
+
+@end
+
+@interface NYPLBookDetailButtonsView : UIView
+
+@property (nonatomic, weak) NYPLBook *book;
```
winniequinn: Given that an `NYPLBook` will never hold a reference to an `NYPLBookDetailButtonsView`, I am not sure why this reference should be `weak`.

Simplified/NYPLBookDetailViewController.m
```diff
@@ -126,18 +139,92 @@ - (void)didSelectReadForBookDetailView:(__attribute__((unused)) NYPLBookDetailVi
    animated:YES];
 }
 
-- (void)didSelectTryAgainForBookDetailView:(__attribute__((unused)) NYPLBookDetailView *)detailView
+- (void)didSelectCitationsForBook:(NYPLBook *)book sender:(id)sender
 {
-  [[NYPLMyBooksDownloadCenter sharedDownloadCenter] startDownloadForBook:self.book];
+  //FIXME: add logic for launching citations here
```
winniequinn: I am not sure citations should be a part of this pull request. As far as I know, the work on citations is not yet complete.

winniequinn:
@gioneill In addition to the above review, I noticed that the "More…" button is shown even if the entire description is already visible. Is it possible to only show it when necessary? I feel like users may just stop hitting it if it often fails to given them any additional information.

I also would not mind if the detail view always bounced, but that's just me!

Simplified/NYPLBookCoverRegistry.m
gioneill: I'm not sure what you mean by "side effects". Right when the detail view appears, it begins to download the larger cover photo, but the first handler returns the cached thumbnail, so that there's a nice transition between the low-res and higher-res images. You can visually see this with a 1-2 second delay even on fast connections, so both handlers are definitely being used.. 

I'm definitely open to a better way of approaching it however.. I just figured to piggyback on the local cache / generated cover logic of the original method rather than to flesh it out fully in two different places.

Simplified/NYPLBookCoverRegistry.m
gioneill: If I'm offline then I want to fall back on either the pinned thumbnail OR the generated cover, which comes from the thumbnail handler.

I think for a time I tried to use a single method for thumbnail or cover image, adding a parameter to switch on the logic.. but made a decision to go this way with it and keep that original method untouched (probably for the fact that I wasn't interested in caching/pinning the higher res cover images).

Simplified/NYPLBook.h
gioneill: Okay I have a basic NYPLBookAuthor class that holds two properties for "name" and "link". I think I'm on the same page as you here.

Simplified/NYPLBook.m
gioneill: I take this new NYPLBookAuthor class and serialize to and from string or array of strings.

Simplified/NYPLBookDetailViewController.m
gioneill: Removed Citation Stubs.

gioneill:
@winniequinn I believe I have everything addressed. Let me know if I understood your thoughts on that data structure for authors. I'm also adjusting the PR to go to master branch.

winniequinn:
@gioneill The author changes look good! In the future, I might suggest making the fields immutable so there's no chance of them changing unexpectedly, however. You can set them in `init` and declare the properties with `let`. If you have time to get that in before merging, that would be ideal. For example:

```swift
final class NYPLBookAuthor: NSObject {

  let name: String
  let relatedBooksLink: URL?

  init(name: String, relatedBooksLink: URL?) {
    self.name = name
    self.relatedBooksLink = relatedBooksLink
  }
}
```

PR #650, opened 2017-07-07
Drm bug #645 (activation workflow changes)

### LINKED ISSUE #645: App crashes upon hitting activation limit
It will then continue to crash upon subsequent restarts making it useless. This began at some point during the multi-library work.

gioneill:
Main Features:
- Consolidates Adobe DRM Activation.. Auth and De-Auth to Signing In and Out only. Reducing the number activation steps should eliminate issues with the Adept library; sometimes crashing.
- Licensor Token saving/refreshing (in the loans feed) had to be audited to make sure it was being saved and used at appropriate times, as well as make sure expired tokens are not used.
- In a rare case where a sign out is attempted with an expired token, the app makes a second attempt with a fresh token.. in the background to the user

To the user these changes should:
- Fix a crash around launching the app / credentials / signing in and out
- Reduce number of lost activations for users
- Reduce "download failed" on loans
- No more stuck loans that cannot be returned until they expire
- Seeing updated loans when switching accounts (no pull-refresh required) (edited)

fixes #645

### DIFF
--- a/Simplified/NYPLAccountSignInViewController.m
+++ b/Simplified/NYPLAccountSignInViewController.m
@@ -702,9 +702,7 @@ replacementString:(NSString *)string
   NSMutableURLRequest *const request =
     [NSMutableURLRequest requestWithURL:[NYPLConfiguration loanURL]];
   
-  // Necessary to support longer login times when using usernames.
   request.timeoutInterval = 20.0;
-  
   request.HTTPMethod = @"HEAD";
   
   self.isCurrentlySigningIn = YES;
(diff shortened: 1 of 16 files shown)


Simplified/NYPLAccountSignInViewController.m
```diff
@@ -702,26 +704,80 @@ - (void)validateCredentials
...
+           [[NYPLAccount sharedAccount] setLicensor:loansFeed.licensor];
+         } else {
+           NYPLLOG(@"Login Failed: No Licensor Token received or parsed from OPDS Loans feed");
+           [self authorizationAttemptDidFinish:NO error:nil];
+           return;
+         }
+         
+         NSMutableArray *licensorItems = [[loansFeed.licensor[@"clientToken"] stringByReplacingOccurrencesOfString:@"\n" withString:@""] componentsSeparatedByString:@"|"].mutableCopy;
+         NSString *tokenPassword = [licensorItems lastObject];
+         [licensorItems removeLastObject];
```
winniequinn: It's not obvious to me that this correct not knowing what a licensor is. We're clearly depending on the server giving us something in some particular format, so it would be great if we could parse it and validate it earlier on in this process rather than depending on it being a string with a particular format at this point. I realize this is a holdover from the earlier implementation however, so if this is the best compromise for now, I am okay with that.

Simplified/NYPLAccountSignInViewController.m
```diff
@@ -702,26 +704,80 @@ - (void)validateCredentials
...
+         NSString *tokenUsername = [licensorItems componentsJoinedByString:@"|"];
+         
+         NYPLLOG(@"***DRM Auth/Activation Attempt***");
+         NYPLLOG_F(@"\nLicensor: %@\n",loansFeed.licensor);
+         NYPLLOG_F(@"Token Username: %@\n",tokenUsername);
+         NYPLLOG_F(@"Token Password: %@\n",tokenPassword);
+         
+         [[NYPLADEPT sharedInstance]
+          authorizeWithVendorID:[[NYPLAccount sharedAccount] licensor][@"vendor"]
```
winniequinn: Just continuing on the last point, this may cause a crash if the `vendor` field is not present when authorizing. It would be nice to go back to this at some point in the future and clean up the representation of licensors.

Simplified/NYPLConfiguration.m
```diff
@@ -22,12 +22,12 @@ + (void)initialize
   }
   
 #if defined(FEATURE_DRM_CONNECTOR)
-  [[NYPLADEPT sharedInstance] setLogCallback:^(__unused NSString *logLevel,
-                                               NSString *const exceptionName,
-                                               __unused NSDictionary *data,
-                                               NSString *const message) {
-    NSLog(@"ADEPT: %@: %@", exceptionName, message);
-  }];
+//  [[NYPLADEPT sharedInstance] setLogCallback:^(__unused NSString *logLevel,
```
winniequinn: If we don't need this code any more, let's just get rid of it.

Simplified/NYPLMyBooksViewController.h
```diff
@@ -5,6 +5,8 @@
 - (id)initWithCoder:(NSCoder *)aDecoder NS_UNAVAILABLE;
 - (id)initWithNibName:(NSString *)nibNameOrNil bundle:(NSBundle *)nibBundleOrNil NS_UNAVAILABLE;
 
+- (void)didSelectSync;
```
winniequinn: A couple of things on this one:

1. Methods with names like `didSelectSync` are generally called in response to an action attached to an interface control. I was a little surprised to see a method with this name exposed in the header. If we want to be able to tell My Books to sync, perhaps we can expose a method named `sync` instead which just calls `didSelectSync` internally?

2. In the original design, My Books would sync automatically whenever `NYPLBookRegistryDidChangeNotification` was fired. Is there a way we can go back to having My Books completely manage its own syncing again so we don't need to worry so much about actively keeping everything up-to-date as appropriate?

Simplified/NYPLSettingsAccountDetailViewController.m
```diff
@@ -377,17 +363,9 @@ - (void)logIn
...
-    [[NYPLMyBooksDownloadCenter sharedDownloadCenter] reset:self.accountType];
-    [[NYPLBookRegistry sharedRegistry] reset:self.accountType];
-    
-    [[NYPLAccount sharedAccount:self.accountType] removeAll];
-    [self setupTableData];
-    [self.tableView reloadData];
-  };
   
 #if defined(FEATURE_DRM_CONNECTOR)
+
```
winniequinn: This is a really minor thing, but it's possible to configure Xcode so it won't insert spaces on blank lines like this and clutter up diffs. Try setting the following in Preferences > Text Editing > Editing:

screen shot 2017-07-11 at 7 35 29 pm

Simplified/NYPLSettingsAccountDetailViewController.m
```diff
@@ -1383,7 +1462,9 @@ - (void)setActivityTitleWithText:(NSString *)text
   [linearView sizeToFit];
   
   self.logInSignOutCell.textLabel.text = nil;
-  [self.logInSignOutCell.contentView addSubview:linearView];
+  if (![self.logInSignOutCell.contentView viewWithTag:1]) {
```
winniequinn: I don't love using `viewWithTag:` like this because it discards type/nullability information and involves magic constants. While my preference would just be to declare a new instance variable, I'd settle for defining an integer constant for this tag so we're not passing around `1` everywhere.

Simplified/NYPLAccountSignInViewController.m
gioneill: It's a good point. I was already mid-way through implementing what you're describing in Swift, and it was making me ask questions around "what should the app allow? or what should be possible? without a valid expected licensor string?". Completing that is still a near-term goal so I'm supporting the compromise for now.

Simplified/NYPLAccountSignInViewController.m
gioneill: This is a little more vulnerable than I was thinking. For now I'm going to protect against nil parameters by instantly returning the completion block with NO and let the calling class handle it as failing to auth/de-auth. I'll put a PR on the other repo.

Simplified/NYPLConfiguration.m
gioneill: For some reason this callback was creating duplicate adept messages in the debugger..

Simplified/NYPLMyBooksViewController.h
gioneill: 1. I've updated to 'syncMyBooks'
2. For the most part I believe there's a bigger issue with how other classes are requesting a BookRegistry load or sync. I noticed that book lists were not loading after a library switch or fresh user sign in, and since the pull-to-refresh is not always discoverable for every user I was looking for a quick way to invoke that.

gioneill: Actually I take it back. If you are switching libraries while in My Books, then there's no reason to sync the registry. That behavior can stay as is + the end of account auth. So I'm undoing my change here.

gioneill: Last comment: I've reached (what I feel) is a good solution which is to also perform a book registry sync when switching libraries.

PR #654, opened 2017-07-11
Add Howard County to library list

datalogics-cgreen:
This adds Howard County's library to the library list for the iOS app, plus the relevant styles and assets.

As with our other library addition(s), we would like to coordinate the release, so _please do not merge just yet_.

Links such as the privacy policy and license are incoming.


gioneill:
Develop has been deprecated. Commits from this PR are incorporated into new branch `datalogics_new_libraries` which has been branched from Master.

PR #664, opened 2017-07-17
Add Frederick County Public Library to iOS client

datalogics-cgreen:
This adds Frederick County Public Library to the library list for the iOS app, plus its styles and assets.

As with our other additions, we would like to coordinate the releases of these libraries, so **please do not merge yet**.

Links such as the privacy policy and license are incoming.


gioneill:
Develop has been deprecated. Commits from this PR are incorporated into new branch `datalogics_new_libraries` which has been branched from Master.

PR #674, opened 2017-07-21
Bug device auth #673

### LINKED ISSUE #673: Xcode Analytics: Crash on downloading book
Turns out Keychain item migration is causing the app to be considered logged in, but not actually device-authorized. This throws an exception when trying to download the book. Fixing crash is part 1, and part 2 is getting ahead of the possible causes and navigating to proper UX (of either prompting for a user sign in or attempting to authorize the user's credentials).

Possible issue in bugsnag (could not find exact stack trace match):
[Bugsnag link](https://app.bugsnag.com/nypl-1/android-open-ebooks/errors/58b473fb3572695876d9adb1?filters%5Bevent.since%5D%5B%5D=30d&filters%5Berror.status%5D%5B%5D=open)

screen shot 2017-07-20 at 6 25 19 pm

gioneill:
Fixes crash caused by a nil parameter. The source of the flow turns out to be from old keychain credentials persisting across app installs. In those cases the device is considered "logged in" but is not actually DRM-authorized with Adobe.

This PR both prevents keychain items from persisting across app deletions and re-installs, and also has a one-time attempt to authorize any devices that are in this current state. It will purge the account if that retry fails, forcing the user to explicitly log back in.

fixes #673 
The related change to the adept submodule:
[PR 12](https://github.com/NYPL-Simplified/DRM-iOS-Adobe/pull/12)

### DIFF
--- a/Simplified/NYPLAccountSignInViewController.m
+++ b/Simplified/NYPLAccountSignInViewController.m
@@ -421,7 +421,7 @@ didSelectRowAtIndexPath:(NSIndexPath *const)indexPath
                        initWithStyle:UITableViewCellStyleDefault
                        reuseIdentifier:nil];
       Account *currentAccount = [[NYPLSettings sharedSettings] currentAccount];
-      if (currentAccount.eulaIsAccepted) {
+      if (currentAccount.eulaIsAccepted || [[NYPLAccount sharedAccount] hasBarcodeAndPIN]) {
         self.eulaCell.accessoryView = [[UIImageView alloc] initWithImage:
                                        [UIImage imageNamed:@"CheckboxOn"]];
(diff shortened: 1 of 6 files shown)


Simplified/NYPLAccountSignInViewController.m
```diff
@@ -621,6 +621,7 @@ - (void)showEULA
 - (void)updateLoginLogoutCellAppearance
 {
   if (self.isCurrentlySigningIn) {
+    self.eulaCell.userInteractionEnabled = NO;
```
winniequinn: I can't say I took the time to track this down fully, but why do we need this when we are also apparently calling `[UIApplication beginIgnoringInteractionEvents]`?

gioneill: You're right, actually.

PR #683, opened 2017-07-25
Bug annotations #672

### LINKED ISSUE #672: Bugsnag: crash at Annotations Header 'get'
[Bugsnag Link](https://app.bugsnag.com/nypl-1/android-open-ebooks/errors/58f3ce137e08960018feabab?filters%5Berror.status%5D%5B%5D=open&filters%5Bevent.since%5D%5B%5D=30d)

Implicitly unwrapped optionals are causing a crash here. Will add optional binding for safety when accessing. Networking methods should handle header errors further down the stream.

gioneill:
Fixes #672  crash when accessing headers var getter.
Should resolve bugsnag bug referenced ticket.

Whatever cause of missing credential header, passing an empty field as the fail-case should allow server and networking method to handle error more gracefully.

### DIFF
--- a/Simplified/NYPLAnnotations.swift
+++ b/Simplified/NYPLAnnotations.swift
@@ -298,11 +298,19 @@ final class NYPLAnnotations: NSObject {
   
   class var headers: [String:String]
   {
-    let authenticationString = "\(NYPLAccount.shared().barcode!):\(NYPLAccount.shared().pin!)"
-    let authenticationData:Data = authenticationString.data(using: String.Encoding.ascii)!
-    let authenticationValue = "Basic \(authenticationData.base64EncodedString(options: Data.Base64EncodingOptions.lineLength64Characters))"
-    
-    return ["Authorization" : "\(authenticationValue)",
+    if let barcode = NYPLAccount.shared().barcode, let pin = NYPLAccount.shared().pin {
+      let authenticationString = "\(barcode):\(pin)"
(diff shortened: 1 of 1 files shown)


Simplified/NYPLAnnotations.swift
```diff
@@ -142,7 +142,7 @@ final class NYPLAnnotations: NSObject {
           debugPrint(#file, "Posted Last-Read \(((parameters["target"] as! [String:Any])["selector"] as! [String:Any])["value"] as! String)")
         }
       } else {
-        guard let error = error as? NSError else { return }
+        guard let error = error as NSError? else { return }
```
winniequinn: This is not directly related to your change per se, but the sum of this code does not seem right to me:

1. The type of response is always `HTTPURLResponse` and can be safely cast using `as!`.

2. If the type of the response is `HTTPURLResponse` and the status code is not 200, we silently ignore it. Not at least logging an error here seems wrong to me, but I think we're actually supposed to be retrying later.

3. The logging code when the response is 200 will crash the app if undocumented keys are not present.

4. The call to `addToOfflineQueue` on line 147 will never be hit because response is always of type `HTTPURLResponse`.

gioneill: I'm delaying this to next sprint when Annotations will be fully updated to support bookmarks. The crash fix is left in as a commit, removing implicitly unwrapped optionals.

Issue #652, opened 2017-07-08
Book Detail cover size


gioneill:
When loading the new book detail, the auto-generated covers will be smaller and shrink down the layout on the rest of the page.

PR #690, opened 2017-08-01
Enlarge auto generated book cover for detail view

vuinguyen:
This PR addresses issue #652 , where the generated book cover was too small compared to the "regular" book cover in the book detail view.

### DIFF
--- a/Simplified/NYPLBookCoverRegistry.m
+++ b/Simplified/NYPLBookCoverRegistry.m
@@ -101,8 +101,7 @@ static NSUInteger const memoryCacheInMegabytes = 2;
         if(image) {
           handler(image);
         } else {
-          //[self thumbnailImageForBook:book handler:handler];  // why are we calling this again?
-          handler([NYPLTenPrintCoverView detailImageForBook:book]);
+          [self thumbnailImageForBook:book handler:handler];
         }
       }];
     }]
--- a/Simplified/NYPLBookCoverRegistry.m
+++ b/Simplified/NYPLBookCoverRegistry.m
@@ -101,7 +101,8 @@ static NSUInteger const memoryCacheInMegabytes = 2;
         if(image) {
           handler(image);
         } else {
-          [self thumbnailImageForBook:book handler:handler];
+          //[self thumbnailImageForBook:book handler:handler];  // why are we calling this again?
+          handler([NYPLTenPrintCoverView detailImageForBook:book]);
         }
       }];
     }]
@@ -144,7 +145,7 @@ static NSUInteger const memoryCacheInMegabytes = 2;
           if(image) {
             handler(image);
           } else {
-            handler([NYPLTenPrintCoverView imageForBook:book]);
+            handler([NYPLTenPrintCoverView thumbnailImageForBook:book]);
           }
         }];
       }]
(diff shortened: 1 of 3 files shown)


Simplified/NYPLBookCoverRegistry.m
```diff
@@ -101,7 +101,8 @@ - (void)coverImageForBook:(NYPLBook *)book handler:(void (^)(UIImage *image))han
         if(image) {
           handler(image);
         } else {
-          [self thumbnailImageForBook:book handler:handler];
+          //[self thumbnailImageForBook:book handler:handler];  // why are we calling this again?
```
gioneill: We can't comment this line out otherwise we lose an important feature. I should have done a better job of documenting this code when I wrote it:

- First load the thumbnail image
  - If it has already been cached it returns the cached image
  - If no image is there, then it generates the TenPrintCover
- Now try and load the detail image ( a bigger sized image )
  - **If that image is not there, fall back to the smaller thumbnail image (or a generated image)

** this is the line that you have changed, which now says "if no large image exists, then create a TenPrintCover generated image, which is not preferable to showing the smaller thumbnail slightly blurry.

For now you can revert this back to its original line.

Simplified/NYPLTenPrintCoverView+NYPLImageAdditions.h
```diff
@@ -2,7 +2,10 @@
 
 @interface NYPLTenPrintCoverView (NYPLImageAdditions)
 
++ (UIImage *)detailImageForBook:(NYPLBook *const)book;
++ (UIImage *)thumbnailImageForBook:(NYPLBook *const)book;
+
 // Must be called on the main thread.
-+ (UIImage *)imageForBook:(NYPLBook *const)book;
++ (UIImage *)imageForBook:(NYPLBook *const)book withSize:(CGSize)size;
```
gioneill: The comment "must be called..." also applies to the two methods you added, so go ahead and move the comment atop all three.

Simplified/NYPLTenPrintCoverView+NYPLImageAdditions.m
```diff
@@ -16,10 +16,22 @@ + (void)initialize {
...
+{
+  return [NYPLTenPrintCoverView imageForBook:book withSize:CGSizeMake(80,120)];
+}
+
++ (UIImage *)imageForBook:(NYPLBook *const)book withSize:(CGSize)size
+{
+  CGFloat const width = size.width;
+  CGFloat const height = size.height;
+
+  Float32 scale = (width <= 80) ? 0.4 : 1.5;
```
gioneill: This overall change is in the right direction. And now with scale needing two values, it's good practice to take interdependencies away. What I mean is that when someone looks here they would not know why 0.4 and 1.5 would be chosen, so we're going to do the same thing we did with "size" and add one more parameter to the signature:

`+ (UIImage *)imageForBook:(NYPLBook *const)book withSize:(CGSize)size textScaleFactor:(CGFloat)scale`

And move the hard-coded values to the calls for "thumbnail" and "detailView"

vuinguyen:
At this point, we have a new function to create a larger generated book cover for the book detail view, but some of the program flow needs to be refactored in order to make room to call that new function. Working with Greg to figure out that code refactoring and calling of the new function before we can wrap up this PR.

gioneill:
Creating issue #782 to continue work on this when we can prioritize it.

Issue #640, opened 2017-06-14
'Report an Issue' frame cut-off on iPad


gioneill:
On iPad when you attempt to Report An Issue (Settings-Accounts-NYPL-Report An Issue) on iPad, the Subject and Body areas move off screen when you try typing text. Same thing happens in the Book Detail on iPad.

vuinguyen:
How to duplicate the problem: Type something in the subject line.
![img_0076](https://user-images.githubusercontent.com/1022275/28897657-ebe1a9d4-779e-11e7-9744-d5e7c7b0fa0a.PNG)

Then, touch the space below the subject line and the subject line disappears!
![img_0077](https://user-images.githubusercontent.com/1022275/28897655-ebdf0d96-779e-11e7-8ab8-0685726de599.PNG)

PR #696, opened 2017-08-07
'Report an Issue' frame cut-off on iPad #640

vuinguyen:
This PR addresses issue #640 , where the subject line would move up and out of view in the "Report an Issue" screen on the iPad, when the user tapped on the message body area beneath the subject line. The problem was that the message body was too tall on the iPad. The solution is to reduce the size of the message body on the "Report an Issue" screen. Doing this, however, creates some empty gray rows beneath the message body, which may not be "aesthetically pleasing" and might confuse the user. But at least we don't lose the subject line anymore. I will continue to investigate whether anything can be done about the gray rows.

iPad screen with code change, landscape
8cbb6458-0817-4913-85b3-1a4fe4dd6b97

iPad screen with code change, portrait
fb3717a2-c016-4896-a6ee-7f119f15979e

### DIFF
--- a/Simplified/NYPLReportIssueViewController.swift
+++ b/Simplified/NYPLReportIssueViewController.swift
@@ -147,43 +147,33 @@ class NYPLReportIssueViewController: UITableViewController, UITextFieldDelegate,
   }
   
   override func tableView(_ tableView: UITableView, heightForRowAt indexPath: IndexPath) -> CGFloat {
-    
-    if HSAppearance.isIPad()
-    {
-      if(indexPath.row == 0){
-        return 44.0;
-      }else if(indexPath.row == 1){
-        return self.view.frame.size.height - 44.0;
-      }else{
-        return 44.0;
-      }
+
+    // height for the subject line of the message
+    let subjectHeight:CGFloat = 44.0
+
+    // height for the actual message body, it gets increased as we increase device screen size
+    var messageHeight:CGFloat = subjectHeight
+
+    let iPhonePadding:CGFloat  = 155.0
+    let iPadPadding:CGFloat    = 200.0
+
+    if indexPath.row == 0 {
+      return subjectHeight
     }
-    else{
-      if(indexPath.row == 0) {
-        return 44.0;
+    // we're at the message body
+    else if indexPath.row == 1 {
+      // is iPad
+      if HSAppearance.isIPad() {
+        messageHeight = messageHeight + iPadPadding
       }
-      else if(indexPath.row == 1) {
-        var messageHeight:CGFloat;
-        //Instead, get the keyboard height and calculate the message field height
-        let orientation:UIDeviceOrientation = UIDevice.current.orientation;
-        if (UIDeviceOrientationIsLandscape(orientation))
-        {
-          messageHeight = 68.0;
-        }
-        else {
-          
-          if (HSAppearance.isTall()) {
-            messageHeight = 249.0;
-          }else{
-            messageHeight = 155.0 + 44.0;
-          }
-        }
-        // return self.view.bounds.size.height - 88.0;
-        return messageHeight;
-        
+      // is iPhone
+      else {
+        messageHeight = messageHeight + iPhonePadding
       }
+
+      return messageHeight
     }
-    
+
     return 0.0
   }
   


Simplified/NYPLReportIssueViewController.swift
```diff
@@ -147,43 +147,33 @@ class NYPLReportIssueViewController: UITableViewController, UITextFieldDelegate,
...
+    if indexPath.row == 0 {
+      return subjectHeight
     }
-    else{
-      if(indexPath.row == 0) {
-        return 44.0;
+    // we're at the message body
+    else if indexPath.row == 1 {
+      // is iPad
+      if HSAppearance.isIPad() {
```
gioneill: Can you think of how you would create different padding values based just on the current size of the table view, and not just whether it's an iPhone or iPad? With split screen, and larger iPhones like the Plus, a user is not guaranteed to see the correct layout.

[Apple has a good diagram as an example.](https://developer.apple.com/library/content/documentation/WindowsViews/Conceptual/AdoptingMultitaskingOniPad/QuickStartForSlideOverAndSplitView.html)

It would be something like this:

```
func tableView(_ tableView: UITableView, heightForRowAt indexPath: IndexPath) -> CGFloat {
    let subjectHeight:CGFloat = 44.0
    let heightOfTableView = tableView.frame.size.height

    if indexPath.row == 0 {
      return subjectHeight
    } else if indexPath.row == 1 {
      return heightOfTableView - subjectHeight * 2
    } else {
      return 0
    }
}
```

gioneill:
This change does reduce the opportunity for the row to "move up" from the keyboard, because the row for text is now shorter. The drawbacks is that a user does not have the same amount of space to type their message, visually it's not optimal to see unused rows and separator lines, and finally there could be future orientations and sizes that makes the bug fully return.

This is a tough table view to fully fix, since it's relying on a storyboard provided by the HelpStack library.

After a lot of investigation I realized the fact that `UITableViewControllers` fix their cell borders to the keyboard (if they're the first responder). This cannot be changed. What can be changed, however, is using a `UIViewController` with a table view, so that is going to be the preferred route on this bug.

I'll point out the few lines of code that need changing. Then following the compiler warnings by removing the `Override` keywords.

gioneill:
1. Change the class to UIViewController, and add the protocols for UITableViewDataSource and UITableViewDelegate.
2. Add a var for the TableView:   `var tableView: UITableView!
`
3. Update the properties in viewDidLoad():
```
    self.tableView = UITableView.init(frame: self.view.frame)
    self.tableView.isScrollEnabled = false
    self.tableView.backgroundColor = UIColor.white
    self.tableView.alwaysBounceVertical = false
```
4. Revert the heightForRowAt... method
5. Fix compiler warnings

Build and run that on iPad to double check but that should be it.

vuinguyen:
Screenshots of the iPad after the latest round of code changes:

On landscape:
c34c9a9e-fa4a-4368-bf6b-672e8a5384d1


On portrait:
1099aafd-afb7-474d-a440-39a66e7337b2

Simplified/NYPLReportIssueViewController.swift
```diff
@@ -10,8 +10,10 @@ import UIKit
 import HelpStack
 import MessageUI
 
-class NYPLReportIssueViewController: UITableViewController, UITextFieldDelegate, UITextViewDelegate, MFMailComposeViewControllerDelegate {
-  
+class NYPLReportIssueViewController: UIViewController, UITableViewDataSource, UITableViewDelegate, UITextFieldDelegate, UITextViewDelegate, MFMailComposeViewControllerDelegate {
+
+  @IBOutlet var tableView: UITableView!
```
gioneill: Everything looks great! Thought you actually do not need an IBOutlet to the tableview here.

Simplified/ReportIssue.storyboard
```diff
@@ -77,6 +77,9 @@
                     </navigationItem>
                     <nil key="simulatedTopBarMetrics"/>
                     <nil key="simulatedBottomBarMetrics"/>
+                    <connections>
+                        <outlet property="tableView" destination="GRj-EQ-BXo" id="8Y9-oc-jsc"/>
+                    </connections>
```
gioneill: Outlet isn't needed for the UIViewController

vuinguyen: There was already a TableView in the ReportIssue.storyboard (plus the cells for Subject and Message), which was why I created an outlet for it in NYPLReportIssueViewController.swift. How would the controller file access the TableView otherwise without an outlet?

gioneill: Yeah you're right. You can then get rid of the line that re-initalizes the tableview in viewDidLoad()

vuinguyen:
Just did a quick test after removing the table .init function in viewDidLoad(). The "Report an Issue" screen still looks and behaves correctly on the iPad.

PR #702, opened 2017-08-11
External link handling in Reader

gioneill:
@winniequinn After careful pruning of Develop branch, in preparation for deprecation, I came across this merge which never made it to master and was never released to production.

I don't remember if this was fully reviewed/tested/cleared or not.


winniequinn:
Good catch, @gioneill!

Issue #697, opened 2017-08-09
Pull-to-refresh activity indicator not hiding correctly


gioneill:
In My Books (and possibly other areas), the activity indicator is not disappearing after some pull-to-refresh actions.

Possible steps to reproduce:
Go to NYPL
Stay logged out
Go to My Books
Pull to refresh
Log in prompt comes up modally
When it dismisses, the indicator stays when it should be hidden.

Also try this with an account that has some books to see if the table view is not moving back up to the top of the screen like it should be. One could temporarily give the view a colored background to make it easier to see what's going on.

PR #703, opened 2017-08-11
Get pull-to-fresh activity indicator to hide correctly #697

vuinguyen:
This PR addresses issue #697 , by hiding the activity indicator after the user dismisses the login prompt on either the My Books or Reservations tab. It also fixes a related problem where, if the user had an account with books, the Collection View would not move up to the top of the screen. In both cases, either an empty View or a Collection View with books now moves up to the top of screen, properly hiding the activity indicator behind it.

### DIFF
--- a/Simplified/NYPLHoldsViewController.m
+++ b/Simplified/NYPLHoldsViewController.m
@@ -113,6 +113,7 @@
   [super viewWillAppear:animated];
   if([NYPLBookRegistry sharedRegistry].syncing == NO) {
     [self.refreshControl endRefreshing];
+    self.collectionView.contentOffset = CGPointZero;
     [[NSNotificationCenter defaultCenter] postNotificationName:NYPLSyncEndedNotification object:nil];
   }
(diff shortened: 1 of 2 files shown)


Simplified/NYPLHoldsViewController.m
```diff
@@ -106,7 +106,11 @@ - (void)viewDidLoad
...
   self.navigationItem.rightBarButtonItem = self.searchButton;
-  
+}
+
+- (void)viewWillAppear:(BOOL)animated
+{
+  [super viewWillAppear:animated];
   if([NYPLBookRegistry sharedRegistry].syncing == NO) {
     [self.refreshControl endRefreshing];
     [[NSNotificationCenter defaultCenter] postNotificationName:NYPLSyncEndedNotification object:nil];
```
gioneill: Just missed adding the line to update the collection view content offset:
`self.collectionView.contentOffset = CGPointZero;`

PR #712, opened 2017-08-18
Update account sign out to allow edge case of user-changed credentials.

### LINKED ISSUE #709: Allow sign out if credentials has changed
This had been present before, but was inadvertently removed during the authorization updates.

There's no way to prevent losing the activation, but the user should at least have an opportunity to sign back in without having to delete and re-download the app.

gioneill:
fixes #709 

Fixes: user could not sign out of an account that had its credentials changed.

### DIFF
--- a/Simplified/NYPLSettingsAccountDetailViewController.m
+++ b/Simplified/NYPLSettingsAccountDetailViewController.m
@@ -402,11 +402,10 @@ NSInteger const linearViewTag = 1;
        [self deauthorizeDevice];
 
      } else {
-       [self showLogoutAlertWithErrorCode:statusCode];
+       [self showLogoutAlertWithError:error];
+       [self removeActivityTitle];
+       [[UIApplication sharedApplication] endIgnoringInteractionEvents];
      }
-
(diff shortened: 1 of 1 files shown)


Simplified/NYPLSettingsAccountDetailViewController.m
```diff
@@ -609,6 +605,22 @@ - (void)showLoginAlertWithError:(NSError *)error
   [self removeActivityTitle];
 }
 
+- (void)showLogoutAlertWithError:(NSError *)error
+{
+  NSString *title; NSString *message;
+  if (error.code == 401) {
+    title = @"Unexpected Credentials";
+    message = @"Your username or password may have changed since the last time you logged in.\n\nIf you believe this is an error, please contact your library.";
```
winniequinn: Should we be using localized strings here?

gioneill: `NYPLAlertController alertWithTitle:` wraps the parameters with NSLocalizedString

PR #719, opened 2017-08-25
Add anne arundel to library list

datalogics-tarar:


gioneill:
Thanks @datalogics-tarar !

The one change I'd recommend is using the "logo-only" image for all three assets. Even the @2x and @3x assets are occupying the same small physical thumbnail size as the @1x.

datalogics-bhaugen:
@gioneill, we are working with Anne Arundel to get updated logos in the correct sizes that only contain the logo (no additional text). We figured it would look odd, we just let libraries see it so they understand why we ask for that.

PR #767, opened 2017-09-26
Changed Alameda's entry ID to 7.

datalogics-dans:
Just noticed this issue as I was taking screenshots for Carroll County. I assume we originally had 3 as a placeholder and never went back and updated it.

### DIFF
--- a/Simplified/Accounts.json
+++ b/Simplified/Accounts.json
@@ -51,8 +51,8 @@
  "mainColor" : "497049"
  },
  {
- "id" : 3,
- "pathComponent" : "3",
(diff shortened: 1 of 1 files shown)


gioneill:
If I have a chance tomorrow, I'll need to make a change to prevent a crash here.
Since no ID=3 would exist after this change, any user that currently is using Alameda will have the app crash when it tries to launch with a new ID, and there's no way out of this cycle until you uninstall the app and reinstall.

I'll add some logic to fall back to ID=0 if its current ID ends up not existing.

PR #792, opened 2017-10-05
Fix undesired reflow on iOS 11 status bar toggle

### LINKED ISSUE #785: Not saving page position (all users/books)
On 3 out of 4 titles I tried on NYPL, there is strange behavior:

Try and skip to any chapter, page next a few times.. wait a couple seconds then see the screen "blip" as if reloading the view. When you leave the book then come back by tapping "read" again, it will start on your last page for a few moments, then skip back to the first page of the chapter.

Reproduce on any random book in NYPL catalog.

Sometimes when just going page by page through a book, waiting a moment, it will revert back to the page you were on.

On every page turn there is always a subsequent "blip" about 1-2 seconds. This is still present in the 2.1.1 1633 beta.

winniequinn:
Closes #785

### DIFF
--- a/Simplified/NYPLReaderReadiumView.m
+++ b/Simplified/NYPLReaderReadiumView.m
@@ -141,6 +141,10 @@ static void generateTOCElements(NSArray *const navigationElements,
   self.webView.navigationDelegate = self;
   self.webView.UIDelegate = self;
   self.webView.scrollView.bounces = NO;
+  // Prevent content from shifting when toggling the status bar.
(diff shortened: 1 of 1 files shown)


gioneill:
Verified fix in the following:
- 3 physical devices. iPhone 5 running iOS 10. iPad Air and iPhone 7 running iOS 11.
- Simulators for iPad Pro (split screen variations, multi-column layout, landscape and portrait), iPhone OS versions down to iOS 8 to ensure no regressions.

PR #807, opened 2017-10-24
Library Card Barcode - Scanning and Creating

### LINKED ISSUE #594: When scanning a barcode, alert displays start and stop characters
Please remove the display of the start and stop characters for in the imager alert box when scanning a barcode.

### LINKED ISSUE #607: Implement Scandit Barcode scanner

### LINKED ISSUE #608: Implement codabar barcode display

### LINKED ISSUE #803: Barcode Scanning with ZXingObjC
Phase 1 implementation, [detailed in this google doc](https://docs.google.com/document/d/1mIuRdrDUWet4rzBnyyFUZg8UqlXN2fbfUjgqldmT4l4/edit?usp=sharing), as a proof of concept to the cross platform open source library for reading and creating barcodes of any standard.

gioneill:
In the sign in controllers:

- Refactored existing logic for creating a barcode image above the username field after a patron logs in. The barcode number for encoding is being read from the authenticationIdentifier returned by the loans feed after a sign in.
- Allows hiding and showing of barcode. Adjusts brightness accordingly. Manually creating image size to reasonable levels on any type of device. Tested and worked well on self-scan and counter hand-held scanning units at SASB.
- NYPLBarcode.swift abstracts any use of the dependency away from the main app view controllers.
- For Scanning: copy/pasted repo's example view controller and removed unnecessary IBOutlets and created initializer with completion block to just return the OCR'ed string then dismiss itself.

resolves #803 
resolves #608 
resolves #607 
resolves #594


Simplified/NYPLBarcode.swift
```diff
@@ -0,0 +1,56 @@
...
+}
+
+/// Manage creation of barcode images that are scannable by physical scanners.
+/// Keep any third party dependency abstracted out of the main app.
+final class NYPLBarcode: NSObject {
+
+  class func image(string: String, size: CGSize, type: NYPLBarcodeType) -> UIImage?
+  {
+    let writer = ZXMultiFormatWriter.writer() as? ZXWriter
+    do {
```
winniequinn: Do we really want to continue here if `writer` is nil? It seems like this can result in a nil `result` being passed to `ZXImage.init` a few lines down.

Simplified/NYPLBarcode.swift
```diff
@@ -0,0 +1,56 @@
...
+        Log.error(#file, "Error creating image from barcode string.")
+        return nil
+      }
+    } catch {
+      Log.error(#file, "Failed to create barcode image with error: \(error.localizedDescription)")
+      return nil
+    }
+  }
+
+  class func size(forSuperviewBounds bounds: CGRect) -> CGSize
```
winniequinn: This seems like a strange method to expose, perhaps. Maybe the `image` method should take a `maxSize` argument instead of `size` and this logic should be kept internal to the class?

Simplified/NYPLBarcodeScanningViewController.h
```diff
@@ -0,0 +1,23 @@
+/*
+ * Copyright 2012 ZXing authors
```
winniequinn: Let's make sure we update the list of licenses in the app to include this license.

Simplified/NYPLBarcode.swift
gioneill: Good call. I've internalized that method, and changed the signature to ask for the superview width so I can make a determination (internally) for a good width depending on the screen, and fixing the height as a constant also defined internally.

Simplified/NYPLBarcodeScanningViewController.h
gioneill: Let me know if I've attributed that correctly. It's the locally-bundled `software-licenses.html` file

Simplified/Accounts.json
```diff
@@ -7,8 +7,8 @@
  "logo" : "LibraryLogoNYPL",
  "needsAuth" : true,
  "supportsSimplyESync" : false,
- "supportsBarcodeScanner" : false,
- "supportsBarcodeDisplay" : false,
+ "supportsBarcodeScanner" : true,
+ "supportsBarcodeDisplay" : true,
```
daryanypl: :)

Simplified/NYPLAccountSignInViewController.m
```diff
@@ -590,6 +591,17 @@ + (void)authorizeUsingExistingBarcodeAndPinWithCompletionHandler:(void (^)())han
 
 #pragma mark -
 
+- (void)scanLibraryCard
+{
+  [NYPLBarcode presentScannerWithCompletion:^(NSString * _Nullable resultString) {
+    if (resultString) {
+      self.usernameTextField.text = resultString;
+      [self.PINTextField becomeFirstResponder];
+      self.loggingInAfterBarcodeScan = YES;  //Prevent text from clearing
```
daryanypl: from clearing under what conditions?

Simplified/NYPLAccountSignInViewController.m
```diff
@@ -590,6 +591,17 @@ + (void)authorizeUsingExistingBarcodeAndPinWithCompletionHandler:(void (^)())han
 
 #pragma mark -
 
+- (void)scanLibraryCard
+{
+  [NYPLBarcode presentScannerWithCompletion:^(NSString * _Nullable resultString) {
```
daryanypl: is running presentScannerWithCompletion guaranteed to not throw an error?

Simplified/NYPLBarcode.swift
```diff
@@ -0,0 +1,72 @@
...
+      } else {
+        return nilWithGenericError()
+      }
+    } catch {
+      Log.error(#file, "Failed to create barcode image: \(error.localizedDescription)")
+      return nil
+    }
+  }
+
+  class func presentScanner(withCompletion completion: @escaping (String?) -> ())
```
daryanypl: present on the screen for the user to see?

Simplified/NYPLBarcode.swift
```diff
@@ -0,0 +1,72 @@
...
+  {
+    guard let scannerVC = NYPLBarcodeScanningViewController.init(completion: completion) else { return }
+    let navController = UINavigationController.init(rootViewController: scannerVC)
+    NYPLRootTabBarController.shared().safelyPresentViewController(navController, animated: true, completion: nil)
+  }
+
+  private class func imageWidthFor(_ superviewWidth: CGFloat) -> CGFloat
+  {
+    if superviewWidth > maxBarcodeWidth {
+      return maxBarcodeWidth
```
daryanypl: can there be a circumstance where something goes wrong, the barcode image is generated as empty, and so there needs to also be a minWidth, so the user can see the broken image icon?

Simplified/NYPLBarcodeScanningViewController.m
```diff
@@ -0,0 +1,264 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+@import PureLayout;
+#import <AudioToolbox/AudioToolbox.h>
```
daryanypl: what's the AudioToolbox for?

Simplified/NYPLBarcodeScanningViewController.m
```diff
@@ -0,0 +1,264 @@
...
+  } else {
+    [previewRect autoSetDimensionsToSize:CGSizeMake(250, 120)];
+  }
+}
+
+- (void)viewWillAppear:(BOOL)animated {
+  [super viewWillAppear:animated];
+
+  self.capture.delegate = self;
+  [self applyOrientation];
```
daryanypl: SimplyE orientation is locked to portrait, am I correct?

Simplified/NYPLBarcodeScanningViewController.m
```diff
@@ -0,0 +1,264 @@
...
+      return @"UPC/EAN extension";
+
+    default:
+      return @"Unknown";
+  }
+}
+
+#pragma mark - ZXCaptureDelegate Methods
+
+- (void)captureResult:(ZXCapture *)__unused capture result:(ZXResult *)result {
```
daryanypl: I'm not seeing where captureResult is being called from?

daryanypl: Also, I'd be hoping for a method doc.

Simplified/NYPLBarcodeScanningViewController.m
```diff
@@ -0,0 +1,264 @@
...
+	CGAffineTransform inverse = CGAffineTransformInvert(_captureSizeTransform);
+	NSMutableArray *points = [[NSMutableArray alloc] init];
+	NSString *location = @"";
+	for (ZXResultPoint *resultPoint in result.resultPoints) {
+		CGPoint cgPoint = CGPointMake(resultPoint.x, resultPoint.y);
+		CGPoint transformedPoint = CGPointApplyAffineTransform(cgPoint, inverse);
+		transformedPoint = [self.view convertPoint:transformedPoint toView:self.view.window];
+		NSValue* windowPointValue = [NSValue valueWithCGPoint:transformedPoint];
+		location = [NSString stringWithFormat:@"%@ (%f, %f)", location, transformedPoint.x, transformedPoint.y];
+		[points addObject:windowPointValue];
```
daryanypl: does points get used?  is there ever a time when a barcode might not be recognized, even though the points are, like maybe when a barcode is smudged?

Simplified/NYPLSettingsAccountDetailViewController.m
```diff
@@ -907,27 +909,35 @@ - (UITableViewCell *)tableView:(__attribute__((unused)) UITableView *)tableView
...
-      labelD.textAlignment = NSTextAlignmentCenter;
-      labelD.adjustsFontSizeToFitWidth = YES;
-      [cell.contentView addSubview:labelD];
-
-      
 
+      if (![self librarySupportsBarcodeDisplay]) {
+        NYPLLOG(@"A nonvalid library was attempting to create a barcode image.");
+      } else {
+        UIImage *barcodeImage = [NYPLBarcode imageFromString:@"23333103390991"
```
daryanypl: what's 23333103390991?

gioneill: This was a mistake! Thank you! A lot of people would have been using my library card...

Simplified/NYPLBarcodeScanningViewController.m
gioneill: Even though this class is inside the main project, it's written by the library authors and I'm treating it as a well-tested part of their framework.

Simplified/NYPLBarcodeScanningViewController.m
gioneill: Refer to above comment.

Simplified/NYPLBarcodeScanningViewController.m
gioneill: Good question! This is a good opportunity to highlight the common "Delegate" design pattern on iOS:
- On line 19 of the header file, we, the writers of this class, specify that we want to adhere to the Protocol `<ZXCaptureDelegate>`. Protocols are equivalent to Interfaces in Java.
- Now the compiler will enforce that we implement the required methods from the protocol.

Where is captureResult being called from? Well we "don't care", because we only care about what happens when the framework wants to call it (i.e. the implementation, which we've just written).

The framework has a reference to our class on line 78: `self.capture.delegate = self;` .. `self` is equivalent to `this` in Java. So the framework has _delegated_ some work to us and our own class, the _delegate_ has been given responsibility to handle it.

Implicit in this pattern is that the one delegating work to others, in this case the framework, is responsible to know what to do, for instance if, somehow the delegate variable is null...

Simplified/NYPLBarcodeScanningViewController.m
gioneill: Correct. This code allows the modal view handling the video scanning to interpret a barcode at any orientation.

Simplified/NYPLBarcodeScanningViewController.m
gioneill: I don't know.

Simplified/NYPLBarcode.swift
gioneill: I really have no idea if the framework would produce an "image" that does not have a barcode, but it seems unlikely. From the code I've seen the common fail-case is to return null for any number of possible issues, which I account for at the call-site. And based on the context of the feature, I'm okay completely hiding the functionality if this is the case.

Simplified/NYPLBarcode.swift
```diff
@@ -0,0 +1,72 @@
...
+  case .code128:
+    return kBarcodeFormatCode128
+  }
+}
+
+/// Manage creation and scanning of barcodes on library cards.
+/// Keep any third party dependency abstracted out of the main app.
+final class NYPLBarcode: NSObject {
+
+  class func image(fromString string: String, superviewWidth: CGFloat, type: NYPLBarcodeType) -> UIImage?
```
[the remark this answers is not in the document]
gioneill: On the move from Swift 2 to Swift 3, Apple completely renamed every method in their Foundation and UI API's from the extremely verbose, almost sentence-like structure to something that takes out all unneccessary vocabulary that either does not describe the method, or sounds redundent between the method and parameter names.

I agree that "image(fromString)..." can seem vague until you look at the callsite and read it fully with the class name prepended:
`let variableName = NYPLBarcode.image(fromString: "string" ....)`
get and set are also implicit on usage, as has been for some time on iOS, so those verbs are also ommitted.

On the second point, since `-> UIImage?` denotes that you either get back a Drawable, or null, receiving "null" is the failure case and is an extremely common pattern for quick and simple methods like this. More complex API's can implement the `throws` keyword and provide an error type enum to enforce the caller to handle all error cases.

Swift uses a documenting structure based on Markdown syntax. There are no header files, so you'd have to place above every public method you're describing. I personally felt there was enough context in the class description to fill in the two methods, but if @winniequinn thinks there needs to be improvement I'll accept the majority vote.

Simplified/NYPLAccountSignInViewController.m
gioneill: resultString is `_Nullable` which I use as check for if the method failed or not. Methods that "throw" errors are enforced at compile time. Longer discussion is below.

Simplified/NYPLAccountSignInViewController.m
gioneill: I actually went ahead and removed this comment. I feel that it's common to do a file-wide search for a variable when trying to understand its purpose, and this variable is only used twice. There is a nice description in the spot where it's used as a conditional.

Simplified/NYPLBarcode.swift
gioneill: Yeah.

gioneill:
We will certainly revisit this for future Passbook integration, so if any notes need to be addressed I'm more than happy to return to it. For now I'm merging.

PR #811, opened 2017-10-27
Add Western Maryland libraries to iOS app

datalogics-cgreen:
This PR adds two libraries -- Ruth Enlow Library of Garrett County, and Washington County Free Library -- which share a circulation manager.

In particular, I'm looking for feedback on the logos; I have only one size of logo (135x135) for either library, so I made two copies of the logo and wrote Contents.json in order to mimic existing libraries. I have a feeling that I do not need all three copies, so I would be pleased to remove two of them and edit Contents.json if this is the case.


datalogics-cgreen:
Commit 8956fce does NOT address my original comment -- I had some badly named logos that had to be removed. There are still three copies of the same logo for either library at the moment.

PR #815, opened 2017-10-30
iPhone X Layout Adjustments

gioneill:
Add safe area padding and some inescapable hard-coded padding for iPhone X.


gioneill:
@winniequinn I've added a commit to reduce dependency on hardcoded values and create a CGRect based on the UIWindow's. Because of the transient nature of the renderer actually being in the view hierarchy, this seems like the most pragmatic approach to me.

PR #819, opened 2017-10-31
Correct Western MD URLs and add email for Carroll

datalogics-dans:
Having "groups/" at the end caused issues when logging in.


datalogics-dans:
Oh, and apparently the support email address for Carroll County is in here. If that never made it in prior, I guess now is as good a time as any.

PR #832, opened 2017-11-14
Set readium-shared-js to 0.27.0 tag

### LINKED ISSUE #758: Ensure readium-shared-js is updated
This change may have gotten stuck on a rotting branch, and never moved into Master. It was part of the investigation into how and when users were losing spots on their books.

If it had not been updated, it will be a good time to test with beta users while reading books.

winniequinn:
It seems it was set to the commit _ahead_ of the 0.27.0 tag. While this didn't cause any issues, it did cause me a little confusion, hence the adjustment.

Closes #758, amends 1429c62b4ad135086af5d102d97b4d1c80e1d04c.


gioneill:
I saw the next commit as being a hot fix to their release, and thought maybe they would have intended to move up the tag but didn’t. Am I overthinking it?

winniequinn:
Haha, maybe not! That is good sleuthing! This doesn't affect us either way because we're just using the build output checked into the repository though. If you're not opposed, I'd prefer to merge this just so we match up with their formal releases.

PR #841, opened 2017-11-24
Update reader icons

gioneill:
Remove borders to match a more modern look comparable to other applications. Also will match upcoming PDF reader icons (that don't have borders). Content insets had to be updated to provide even spacing and good visual balance between them.

Icons maintain minimum area recommended by Apple for enough room to tap (44x44 pt).

Feature branch has changes to icons, so PR is not back to master, but should at least provide a way to discuss this or make it visible as a separate feature.


gioneill:
simulator screen shot - iphone 6s - 2017-11-24 at 13 22 09

PR #863, opened 2017-12-20
Overhaul reader animation and gesture code

### LINKED ISSUE #576: Reader - Single Tap For Links - Inconsistent Behavior
When a user taps on some kind of link in a book, they should be taken through to the destination with the navigation UI remaining _consistent_ before and after.

In some instances this works as expected: Navigating a table of contents inside a book, and tapping a chapter takes a user to that page and keeps the nav hidden or shown consistently.

In other instances, tapping links (like endnotes) toggles the navigation to show/hide, requiring the user to tap again somewhere on the page to toggle it back.

Example of TOC: Make Me - Lee Child (Overdrive)
Example of endnotes: Twenty Thousand Leagues Under the Seas (Standard Ebooks)

### LINKED ISSUE #858: Replace gesture handling and animations in reader with JS-based approach
The current native code for rasterizing webviews and juggling them about for page turning does not work well and occasionally crashes. The native code for tap detection is fundamentally unable to work well as has often been discussed internally. The goal is to replace everything with a simply JS-based solution until R2 can be used instead.

Obsoletes #670.

winniequinn:
This PR makes numerous improvements:

* Interactive EPUB 3.X content is now more likely to work as intended.
* It is now possible to swipe over a link to turn a page.
* Skeuomorphic page turn animations have been eliminated.
* Crashes related to the juggling of rasterized webviews will no longer happen.
* Crashes related to the use of synchronous JavaScript to detect link locations will no longer happen.
* `NYPLReaderSettings` no longer breaks abstraction boundaries by referencing an instance of `NYPLReaderReadiumView`. The proper plumbing is now handled via delegates.
* Much unnecessary code has been purged.

Closes #858
Closes #576

### DIFF
--- a/Simplified/NYPLReaderReadiumView.h
+++ b/Simplified/NYPLReaderReadiumView.h
@@ -23,7 +23,6 @@
 - (void) applyMediaOverlayPlaybackToggle;
 - (void) openPageLeft;
 - (void) openPageRight;
-- (BOOL) touchIntersectsLink:(UITouch *)touch;
 
 - (NSString*) currentChapter;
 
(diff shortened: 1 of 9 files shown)


gioneill:
Merging after 2.2.1 release per @winniequinn 's instructions.

PR #869, opened 2017-12-28
Suppress warnings

winniequinn:
This PR suppresses _some_ of the warnings that are not helpful when building the application. Most of these were coming from libraries used by the application itself. No helpful warnings were suppressed: Only code that is effectively outside of our control is affected.

A much bigger second pass is necessary to remove the rest of the warnings. This will include at a minimum:

- Upgrading the project and all dependencies to Swift 4
- Packaging Readium SDK in a binary framework so we never rebuild it
- Fixing numerous _legitimate_ issues in the application codebase

Certain submodules have already had changes pushed to master to accommodate this PR. In the future, we'll eliminate the use of Git submodules and rely on proper packages so we won't need to do this sort of thing.


Simplified.xcodeproj/project.pbxproj
```diff
@@ -1241,7 +1241,7 @@
 			attributes = {
 				CLASSPREFIX = NYPL;
 				LastSwiftUpdateCheck = 0730;
-				LastUpgradeCheck = 0820;
+				LastUpgradeCheck = 0920;
```
gioneill: What is this attribute referring to?

winniequinn: I went ahead and let Xcode verify that the project is up to its current standards. This involved Xcode enabling some new warnings and setting `LastUpgradeCheck` to indicate that Xcode 9.2 had done its job. See 541c07489ef35ed16122e8b7658f48fde3d033b3 for the exact change.

PR #879, opened 2018-01-02
Bookmark Sync Improvements + NYPLAccount stopgaps

gioneill:
2 new commits to resolve #678 

In a typical user session inside SimplyE Settings, if a user picks a non-active library account (to sign in or change settings), there will likely be several scattered calls to NYPLAccount, and a single call to the singleton "sharedAccount" will rewrite all the global variables and start creating havok with keychain credentials and other values stored in the keychain.

It's too fragile to rely on developer knowledge of this fact, so the class should ultimately be reworked for safety.

The most immediate issue was that credentials would overwrite the current active account creds. So this has been fixed in addition to showing the credentials when re-visiting the particular detail page.

### DIFF
--- a/Simplified/NYPLAccount.m
+++ b/Simplified/NYPLAccount.m
@@ -222,7 +222,7 @@ NSString * deviceIDKey = @"NYPLAccountDeviceIDKey";
 - (void)setAuthorizationIdentifier:(NSString *)identifier
 {
   if(!(identifier)) {
-    @throw NSInvalidArgumentException;
+    NYPLLOG(@"Authorization ID (Barcode String) was nil.");
   }
   
   [[NYPLKeychain sharedKeychain] setObject:identifier forKey:authorizationIdentifierKey];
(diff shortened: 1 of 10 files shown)


Simplified/NYPLAccount.m
```diff
@@ -222,7 +222,8 @@ - (void)setLicensor:(NSDictionary *)licensor
 - (void)setAuthorizationIdentifier:(NSString *)identifier
 {
   if(!(identifier)) {
-    @throw NSInvalidArgumentException;
+    NYPLLOG(@"Authorization ID (Barcode String) was nil.");
```
winniequinn: What was the motivation for not throwing an exception here? I think the app threw `NSInvalidArgumentException` because calling `setAuthorizationIdentifier:` with `nil` indicated an error elsewhere in the program. Simply doing nothing here would just mask the problem and cause it to show up later on and in a harder to detect way, would it not? My preference would be for the app to crash immediately so we'd be able to know the problem is happening and we'd have a clear stack trace.

gioneill: That's fair. I was moving the nil check perhaps a little too deep. With the current and only usecase for the property, the UX is designed with a nil-case, but someone else may have another need for it  later on. I personally don't trust everyone's OPDS feed to always have this be non nil, so I'll just check before setting it from the feed that it originates from.

PR #885, opened 2018-01-08
Fix link error and finish Swift 4 update for unit tests

winniequinn:
This allows unit tests to once again build and run. It required fixing a CocoaPods linking issue and completing the Swift 4 upgrade for unit tests. 

Note that most unit tests are currently disabled. A separate pass will be required to bring the tests up-to-date and enable them as they once again become functional. Until then, at least adding new tests will be possible.


SimplifiedTests/UpdateCheckTests.swift
```diff
@@ -17,7 +19,7 @@ class UpdateCheckTests: XCTestCase {
     
     UpdateCheck.performUpdateCheck(URL) { (result: UpdateCheck.Result) in
       switch result {
-      case .UpToDate:
+      case .upToDate:
```
gioneill: Yes this is correct casing.

PR #891, opened 2018-01-17
Sync Feature - Maintenance Refactoring and Cleanup

gioneill:
Mostly just some cleanup done while trying to find the cause of a larger issue present in the Sync feature.

- Renamed some methods and variables
- Got rid of unused protocol methods
- Added some missing completion handler calls to ensure correct sync logic
- Updated Bugsnag config (Resolves #887 )

### DIFF
--- a/Simplified/NYPLAnnotations.swift
+++ b/Simplified/NYPLAnnotations.swift
@@ -368,7 +368,9 @@ final class NYPLAnnotations: NSObject {
   }
 
   // MARK: - Bookmarks
-  
+
+  // Completion handler will return a nil parameter if there are any failures with
+  // the network request, deserialization, or sync permission is not allowed.
   class func getServerBookmarks(forBook bookID:String?, atURL annotationURL:URL?, completionHandler: @escaping (_ bookmarks: [NYPLReaderBookmark]?) -> ()) {
 
     if !syncIsPossibleAndPermitted() {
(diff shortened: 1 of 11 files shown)


Simplified/NYPLAnnotations.swift
```diff
@@ -439,7 +441,7 @@ final class NYPLAnnotations: NSObject {
     }
 
     getServerBookmarks(forBook: id, atURL: annotationUrl) { bookmarks in
-      completionHandler(bookmarks
+      completionHandler(bookmarks?
```
winniequinn: This is a small thing, but `X.filter({F}).first` is equivalent to the faster `X.first { F }`. I'd prefer the latter form to avoid constructing a new array needlessly: `first` with the block argument will return the first result that matches immediately.

Simplified/Simplified-Info.plist
```diff
@@ -17,7 +17,7 @@
 	<key>CFBundlePackageType</key>
 	<string>APPL</string>
 	<key>CFBundleShortVersionString</key>
-	<string>2.2.2</string>
+	<string>2.2.3</string>
```
winniequinn: Also a small thing, but it would be ideal to do version bumps outside of feature PRs like this: They don't have anything to do with the PR directly. It's fine to let it go this time though.

gioneill: Okay sounds good. Should it be a push directly on master after a release, or still PR'ed?

Simplified/NYPLAnnotations.swift
gioneill: That's great to know, thanks. For this particular instance, instead I decided to delete the entire method. It it no longer being used. It was designed as part of the following logic:

In order to update a bookmark with its server-given annotation ID:
- Upload a bookmark
- Download all bookmarks for the book
- Filter for this bookmark and re-create the object with the ID

This is now accomplished in other methods by:
- Upload a bookmark
- Create the object with the ID present in the response of the upload request

gioneill:
@winniequinn Two commits added after your notes. Thanks.

gioneill:
@winniequinn @winniequinn winnie

Simplified/Simplified-Info.plist
winniequinn: I think small things like that can go right to master.

PR #902, opened 2018-01-22
Prevent attempt to access index out of bounds on an array.

### LINKED ISSUE #899: NSRangeException in NYPLReaderTOCViewController
## Error in iOS - Simplified

**NSRangeException** in **NYPLReaderTOCViewController**
*** -[__NSArrayM objectAtIndex:]: index 7 beyond bounds [0 .. 0]

[View on Bugsnag](https://app.bugsnag.com/nypl-1/ios-simplified/errors/5a582109277e26001a679655?event_id=5a5c472d745892001bdd161a&i=gh&m=ci)

## Stacktrace

    Simplified/NYPLReaderTOCViewController.m:243 - -[NYPLReaderTOCViewController tableView:commitEditingStyle:forRowAtIndexPath:]

[View full stacktrace](https://app.bugsnag.com/nypl-1/ios-simplified/errors/5a582109277e26001a679655?event_id=5a5c472d745892001bdd161a&i=gh&m=ci)

*Created by Gregory O'Neill via Bugsnag*

gioneill:
fixes #899

### DIFF
--- a/Simplified/NYPLReaderTOCViewController.m
+++ b/Simplified/NYPLReaderTOCViewController.m
@@ -238,12 +238,13 @@ didSelectRowAtIndexPath:(NSIndexPath *const)indexPath
 }
 
 -(void)tableView:(UITableView *)tableView commitEditingStyle:(UITableViewCellEditingStyle)editingStyle forRowAtIndexPath:(NSIndexPath *)indexPath {
-  
   if (editingStyle == UITableViewCellEditingStyleDelete) {
-    NYPLReaderBookmark *bookmark = self.bookmarks[indexPath.row];
-    [self.bookmarks removeObjectAtIndex:indexPath.row];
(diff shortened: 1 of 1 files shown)


gioneill:
@winniequinn Like everything you've mentioned, I've spent a reasonable amount of time trying to understand how this would happen, and have not been able to find it. So I'm simply looking to prevent the crash for now.

The stack trace shows a crash due to out of bounds array access on self.bookmarks in the TOC view controller in the didCommitEdit method.

The property is set by the class that creates the object, copied from the readium view, and also updated at the end of a syncing network request. When the user toggles between the two TOC types with the segment bar, the same tableview reloads with new data (the delegate methods switch depending on which segment is selected).

There are three total events logged for it, so I'll leave it to your judgment whether or not we leave this as an open issue for now.

gioneill:
I'm going to merge.

I'm creating issue #904 to highlight that there is still a bug to solve.

PR #903, opened 2018-01-23
Offline queue sql fix #897

### LINKED ISSUE #897: SIGTRAP in SQLite.Statement.next() -> [SQLite.Binding?]?
## Error in iOS - Simplified

**SIGTRAP** in **SQLite.Statement.next() -> [SQLite.Binding?]?**

[View on Bugsnag](https://app.bugsnag.com/nypl-1/ios-simplified/errors/5a4f7a8156edd600189e25c4?event_id=5a65f14b70eab9001a4bc44f&i=gh&m=ci)

## Stacktrace

    Frameworks/SQLite.framework/SQLite:0 - SQLite.Statement.next() -> [SQLite.Binding?]?
    Frameworks/SQLite.framework/SQLite:0 - partial apply forwarder for closure #1 () -> SQLite.Row? in closure #2 () -> Swift.AnyIterator<SQLite.Row> in SQLite.Connection.prepare(SQLite.QueryType) throws -> Swift.AnySequence<SQLite.Row>
    Frameworks/SQLite.framework/SQLite:0 - partial apply forwarder for reabstraction thunk helper from @callee_owned () -> (@owned SQLite.Row?) to @callee_owned () -> (@out SQLite.Row?)
    Simplified/NYPLNetworkQueue.swift:0 - next
    Simplified/NYPLNetworkQueue.swift:0 - retryQueue
    Simplified/NYPLReachability.m:61 - -[NYPLReachability reachabilityChanged:]
    Simplified/NYPLReachabilityManager.m:96 - ReachabilityCallback

[View full stacktrace](https://app.bugsnag.com/nypl-1/ios-simplified/errors/5a4f7a8156edd600189e25c4?event_id=5a65f14b70eab9001a4bc44f&i=gh&m=ci)

*Created by Gregory O'Neill via Bugsnag*

gioneill:
Improves reliability and organization of NetworkQueue in a couple ways:

- Synchronizes an iterator on the sqlite table so that network requests and their completion blocks do not happen on different threads, which is likely to resolve #897 
- Turn NetworkQueue into a singleton instead of fully static class. NYPLReachability doesn't need to "know" about the queue. The instance is created on app launch in the app delegate.

### DIFF
--- a/Simplified/NYPLNetworkQueue.swift
+++ b/Simplified/NYPLNetworkQueue.swift
@@ -7,42 +7,51 @@ import SQLite
  will retry any queued requests and purge them if necessary.
  */
 final class NetworkQueue: NSObject {
-  
+
+  static let shared = NetworkQueue()
+
+  override init() {
+    super.init()
+    NotificationCenter.default.addObserver(forName: NSNotification.Name.NYPLReachabilityHostIsReachable,
+                                           object: nil,
+                                           queue: nil) { notification in self.retryQueue() }
+  }
+
   static let StatusCodes = [NSURLErrorTimedOut,
-                     NSURLErrorCannotFindHost,
-                     NSURLErrorCannotConnectToHost,
-                     NSURLErrorNetworkConnectionLost,
-                     NSURLErrorNotConnectedToInternet,
-                     NSURLErrorInternationalRoamingOff,
-                     NSURLErrorCallIsActive,
-                     NSURLErrorDataNotAllowed,
-                     NSURLErrorSecureConnectionFailed]
-  static let MaxRetriesInQueue = 5
-  
+                            NSURLErrorCannotFindHost,
+                            NSURLErrorCannotConnectToHost,
+                            NSURLErrorNetworkConnectionLost,
+                            NSURLErrorNotConnectedToInternet,
+                            NSURLErrorInternationalRoamingOff,
+                            NSURLErrorCallIsActive,
+                            NSURLErrorDataNotAllowed,
+                            NSURLErrorSecureConnectionFailed]
+  let MaxRetriesInQueue = 5
+
   let serialQueue = DispatchQueue(label: "org.nypl.labs.SimplyE.networkQueue")
 
   enum HTTPMethodType: String {
     case GET, POST, HEAD, PUT, DELETE, OPTIONS, CONNECT
   }
   
-  private static let path = NSSearchPathForDirectoriesInDomains(.applicationSupportDirectory, .userDomainMask, true).first!
+  private let path = NSSearchPathForDirectoriesInDomains(.applicationSupportDirectory, .userDomainMask, true).first!
   
-  private static let sqlTable = Table("offline_queue")
+  private let sqlTable = Table("offline_queue")
   
-  private static let sqlID = Expression<Int>("id")
-  private static let sqlLibraryID = Expression<Int>("library_identifier")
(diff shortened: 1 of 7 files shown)


Simplified/NYPLNetworkQueue.swift
```diff
@@ -1,41 +1,57 @@
...
+ will retry any queued requests and purge them if necessary.
+ */
 final class NetworkQueue: NSObject {
-  
+
+  static let shared = NetworkQueue()
+
+  override init() {
+    super.init()
+    NotificationCenter.default.addObserver(forName: NSNotification.Name.NYPLReachabilityHostIsReachable,
```
winniequinn: Even though this only gets used via a shared instance for now, you should probably still pair this with a `removeObserver` call in `deinit`.

Simplified/NYPLNetworkQueue.swift
```diff
@@ -1,41 +1,57 @@
...
+                            NSURLErrorCannotConnectToHost,
+                            NSURLErrorNetworkConnectionLost,
+                            NSURLErrorNotConnectedToInternet,
+                            NSURLErrorInternationalRoamingOff,
+                            NSURLErrorCallIsActive,
+                            NSURLErrorDataNotAllowed,
+                            NSURLErrorSecureConnectionFailed]
+  let MaxRetriesInQueue = 5
+
+  let serialQueue = DispatchQueue(label: "org.nypl.labs.SimplyE.networkQueue")
```
winniequinn: This is too picky perhaps, but I'd recommend the following instead:

```swift
let serialQueue = DispatchQueue(Bundle.main.bundleIdentifier!
                                + "."
                                + String(describing: NetworkQueue.self))
```

This code will continue to work correctly even if the bundle ID changes and it will allow refactoring tools to handle changes to the class name.

Simplified/NYPLNetworkQueue.swift
```diff
@@ -92,31 +108,29 @@ final class NetworkQueue: NSObject {
       Log.error(#file, "SQLite Error: Could not insert or update row")
     }
   }
-  
-  class func retryQueue()
+
+  private func retryQueue()
```
winniequinn: ❤️

Simplified/NYPLNetworkQueue.swift
```diff
@@ -92,31 +108,29 @@ final class NetworkQueue: NSObject {
...
   {
-    guard let db = startDatabaseConnection() else { return }
-    
-    do {
-      for row in try db.prepare(sqlTable) {
-        self.retry(db: db, requestRow: row)
+    serialQueue.async {
```
winniequinn: _Extremely_ minor, but you use `serialQueue.async` here and `self.serialQueue.async` below. I'd prefer the latter in both places but consistency would be nice either way.

Simplified/NYPLReachability.m
```diff
@@ -58,7 +60,7 @@ - (void) reachabilityChanged:(NSNotification *)note
     switch (netStatus) {
       case ReachableViaWiFi:
       case ReachableViaWWAN:
-        [NetworkQueue retryQueue];
+        [[NSNotificationCenter defaultCenter] postNotificationName:NYPLReachabilityHostIsReachableNotification object:self];
```
winniequinn: Very happy with this change!

gioneill:
Thanks 👍

PR #907, opened 2018-01-29
Refactor OPDS acquisitions

winniequinn:
This PR reworks the representation of acquisitions, indirect acquisitions, and availability information in the application, thus laying the groundwork for future content types and digital restriction management approaches.

The main highlights are as follows:

* `NYPLOPDSAcquisition`: This is the new class for representing acquisitions.
* `NYPLOPDSIndirectAcquisition`: No equivalent to this existed before.  This allows indirect acquisitions to be handled properly.
* `NYPLOPDSAcquisitionAvailability`: Before, availability was tracked on the book level. Now, it is correctly tracked on the acquisition level. Enjoy the visitor pattern fun in this one.
* `NYPLOPDSLink`: This class was pared back to what it was originally. Most information was moved elsewhere. Some that served no purpose at all was removed.
* `NYPLBook`: This was simplified greatly due to the new `NYPLOPDSAcquisition` class.
* `NYPLBookAcquisitionPath`: This flattens a tree of `NYPLOPDSAcquisition` and `NYPLOPDSIndirectAcquisition` objects into a set of linear acquisition paths the app can choose between. Filtering by content type and acquisition relation is possible.

Additional work will one day need to be done on `NYPLMyBooksDownloadCenter`, but it won't be in this PR. More unit tests should also be added to stress the new representations.

I will take one last proofreading pass at this before a merge, especially around serialization and deserialization—but please go ahead and review!


winniequinn:
@djds23: Feel free to review this if you wish, but it's not critical.

@gioneill: Please grab me when you have time so I can walk you through this!

Simplified/NYPLBook.m
```diff
@@ -329,9 +226,151 @@ - (instancetype)initWithDictionary:(NSDictionary *)dictionary
...
+    NSMutableArray<NYPLOPDSAcquisition *> *const mutableAcqusitions =
+      [NSMutableArray arrayWithCapacity:acquisitionsArray.count];
+
+    for (NSDictionary *const acquisitionDictionary in acquisitionsArray) {
+      assert([acquisitionDictionary isKindOfClass:[NSDictionary class]]);
+
+      NYPLOPDSAcquisition *const acquisition = [NYPLOPDSAcquisition acquisitionWithDictionary:acquisitionDictionary];
+      assert(acquisition);
+
+      [mutableAcqusitions addObject:acquisition];
```
djds23: if the assertion. fails in non-debug mode this could try and append `nil` here and 💣 , should we also check its presence in the app?

winniequinn: We do not define `NDEBUG` for release builds so assertions made with `assert` are always checked. In any case, the assertion here is just to check that my logic was correct: The assertion should _always_ be true, and if it isn't, it indicates a bug in the code that needs to be fixed. As such, I'd actually prefer the app crash so we can see it and fix it as opposed to just quietly bailing out of the method and logging something that'll never be noticed. I hope that makes sense!

PR #913, opened 2018-02-13
Support library-specific help desks

winniequinn:
Depends on NYPL-Simplified/Certificates#1.


gioneill:
@winniequinn approved

PR #924, opened 2018-03-06
Gracefully handle books the app does not support

### LINKED ISSUE #923: Books on loan or hold of an unsupported type erroneously appear in My Books
Reported by @ricardoom.

winniequinn:
This change allows the app to represent books it does not support due to content type, required DRM technologies, et cetera. By representing unsupported books explicitly, the app can hide them from the user as appropriate.

Closes #923

### DIFF
--- a/Simplified/NYPLBookDetailButtonsView.m
+++ b/Simplified/NYPLBookDetailButtonsView.m
@@ -207,15 +207,15 @@
                                 HintKey: hint}];
       }
       break;
+    }
     case NYPLBookButtonsStateDownloadInProgress:
+    {
+      if (self.showReturnButtonIfApplicable)
       {
-        if (self.showReturnButtonIfApplicable)
-        {
-          visibleButtonInfo = @[@{ButtonKey: self.cancelButton,
-                                  TitleKey: NSLocalizedString(@"Cancel", nil),
-                                  HintKey: [NSString stringWithFormat:NSLocalizedString(@"Cancels the download for the current book: %@", nil), self.book.title],
-                                  AddIndicatorKey: @(NO)}];
-        }
+        visibleButtonInfo = @[@{ButtonKey: self.cancelButton,
+                                TitleKey: NSLocalizedString(@"Cancel", nil),
+                                HintKey: [NSString stringWithFormat:NSLocalizedString(@"Cancels the download for the current book: %@", nil), self.book.title],
+                                AddIndicatorKey: @(NO)}];
       }
       break;
     }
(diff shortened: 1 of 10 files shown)


Simplified/NYPLBookDetailButtonsView.m
```diff
@@ -232,7 +232,12 @@ - (void)updateButtons
                                 HintKey: [NSString stringWithFormat:NSLocalizedString(@"Cancels the failed download for this book: %@", nil), self.book.title],
                                 AddIndicatorKey: @(NO)}];
       }
+      break;
     }
+    case NYPLBookButtonsStateUnsupported:
+      // The app should never show books it cannot support, but if it mistakenly does,
+      // no actions will be available.
+      visibleButtonInfo = @[];
```
gioneill: So it would appear as a book entry detail page without any buttons? I like it.

Simplified/NYPLBookRegistryRecord.m
```diff
@@ -47,6 +47,18 @@ - (instancetype)initWithBook:(NYPLBook *const)book
     self.bookmarks = [[NSMutableArray alloc] init];
   }
 
+  if (!book.defaultAcquisition) {
+    // Since the book has no default acqusition, there is no reliable way to
+    // determine if the book is on hold (although it may be), nor is there any
+    // way to download the book if it is available. As such, we give the book a
+    // special "unsupported" state which will allow other parts of the app to
+    // ignore it as appropriate. Unsupported books should generally only appear
+    // when a user has checked out a book in an unsupported format using another
```
gioneill: *checked out or reserved... Just in case someone reads that literally..

Simplified/NYPLBookRegistry.m
```diff
@@ -662,7 +662,12 @@ - (NSArray *)heldBooks
...
 - (NSArray *)myBooks
 {
-  return [self booksMatchingStateMask:~(NYPLBookStateHolding | NYPLBookStateUnregistered)];
+  return [self booksMatchingStateMask:
+          (NYPLBookStateDownloadNeeded
+           | NYPLBookStateDownloading
+           | NYPLBookStateDownloadFailed
+           | NYPLBookStateDownloadSuccessful
+           | NYPLBookStateUsed)];
 }
```
gioneill: So if I understand this correctly, accessing `myBooks` or `heldBooks` should reliably give me what I expect in each of those tabs in the UI, WITHOUT having to perform any further array filtering? When I was looking at the code inside the view controllers for those two classes, I think at the time I was less convinced that different code maintainers were interpreting things differently..

PR #937, opened 2018-03-14
Update numberOfLines to 0 for UILabel

### LINKED ISSUE #936: messageLabel does not expand on book detail view to fit text on all devices
The ellipsis at the end of the label indicates that apple is truncating the text.

screen shot 2018-03-14 at 12 56 11 pm

Should appear like this:
screen shot 2018-03-14 at 12 58 19 pm

djds23:
This indicates that the number of lines should fit the text

closes #936 

on 4 inch phones
screen shot 2018-03-14 at 3 51 21 pm

on 5.5 inch phones
screen shot 2018-03-14 at 3 54 51 pm

### DIFF
--- a/Simplified/NYPLBookDetailNormalView.m
+++ b/Simplified/NYPLBookDetailNormalView.m
@@ -33,7 +33,7 @@ typedef NS_ENUM (NSInteger, NYPLProblemReportButtonState) {
   self.messageLabel = [[UILabel alloc] init];
   self.messageLabel.font = [UIFont customFontForTextStyle:UIFontTextStyleBody];
   self.messageLabel.textColor = [NYPLConfiguration backgroundColor];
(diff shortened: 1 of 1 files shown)


gioneill:
I think this is a pretty acceptable loss of vertical spacing for gain in information.
Before the second string was added, this "status window" was more akin to a fixed "liquid crystal display" metaphor like the earliest versions of iTunes and Xcode.

It's just something to think about when we come back to the UX of the main app:
- Where should the user's attention be
- What should be given relative importance to other things on screen

djds23:
I think there is an argument that since this view is scrollable there is not much of a loss to the vertical space of the application. However it is unclear on 4 inch devices that the view is scrollable and it might make sense for us to shave a few lines off of the `description` cell so that the `information` cell is more clearly displayed

PR #961, opened 2018-04-11
Update rounded button view to show time remaining

### LINKED ISSUE #940: Time Remaining to Expiration of "Hold Available" not visible in download button
There should be a clock with an expiration on a Hold that has become availabe to checkout. After the expiration, the hold goes away for that user and goes to the next patron in the queue.

This should be the same button format as the expiration on a "Currently checked out loan."

This was likely inadvertently removed when the expiration clock view was taken out of certain views.

simulator screen shot - iphone 7 - 2018-03-19 at 10 11 43

djds23:
Before:
screen shot 2018-04-11 at 2 36 51 pm

After:
screen shot 2018-04-11 at 2 34 31 pm

Closes #940 

Note:

The "Remove" button on the the book detail page is still missing the clock, but I think thats okay because the red warning text shows when it will become available for checkout:

screen shot 2018-04-11 at 2 38 44 pm

### DIFF
--- a/Simplified/NYPLBookButtonsView.m
+++ b/Simplified/NYPLBookButtonsView.m
@@ -280,10 +280,14 @@ NYPLBookButtonsViewStateWithAvailability(id<NYPLOPDSAcquisitionAvailability> con
          }
        }
        unlimited:nil
-       reserved:nil
+       reserved:^(NYPLOPDSAcquisitionAvailabilityReserved * _Nonnull reserved) {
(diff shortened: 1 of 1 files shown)


gioneill:
And there's confidence, even in the absense of a screenshot for it, that this will take effect for a Reservation that moves into "Available for checkout" ? Meaning, the clock on the button should reflect the small window of time (usually a few days) that is available for a patron to _actually_ check out the book.

Books won't check out without explicit user interaction when a Reservation becomes available.

That's a nice touch to bring the clock back to the list views, too. Thanks!

PR #988, opened 2018-07-02
Integrate PDF Reader, No Local Data Save

vuinguyen:
We just want to render a PDF book in the PDF render. The bookmark and annotation features will still work, but no data of any kind will be saved locally: last page read, bookmarks, or annotations; and there will be no server syncing of data at this point either. Using the OpenTextbooks feed for the source of PDF books for this PR.

### DIFF
--- a/Simplified/Accounts.json
+++ b/Simplified/Accounts.json
@@ -52,6 +52,21 @@
  "mainColor" : "#497049"
  },
  {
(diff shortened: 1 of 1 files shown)


vuinguyen:
oops I created this PR on the NYPL branch by mistake. Going to close now

PR #990, opened 2018-07-03
Ignore ReaderClientCert.sig

winniequinn:
I'm surprised this was never done. Presumably we all had it in our personal ignore files.


winniequinn:
@gioneill It is relevant on both platforms. Merging as discussed in-person.

PR #993, opened 2018-07-16
Fix crash on invalid feed

winniequinn:
Co-authored-by: Greg O'Neill <gioneill@gmail.com>


gioneill:
When accessing `.type` of the OPDS Feed, on the condition of an invalid feed, the code was incorrectly returning a Boolean instead of the preferred type. Both are considered "Int" by the runtime, so was not caught by the compiler.

Prudence should be kept in the future to look out for "assignment" `=` versus "comparison" `==`. Usage of font ligations could exacerbate this by removing useful features of monospace.

PR #997, opened 2018-07-26
Fix Report A Problem Button #1033

vuinguyen:
This branch addresses JIRA issue: http://jira.nypl.org/browse/SIMPLY-1033 , where the Report A Problem button disappears from the Book Detail View when going from a feed that supports Report a Problem to a feed that does NOT support Report a Problem, and back to the supporting feed. 

With this change, the Report A Problem button appears in the Book Detail View for a feed that support Report A Problem, and the button does NOT appear in the Book Detail View for a feed that does not support Report A Problem, regardless of switching back and forth between supporting and non-supporting feeds.


gioneill:
The "Report a Problem" URL was not being cached to disk for a saved OPDS entry. When the app recreated the entry from the cache, it was missing that key in the dictionary. Thanks, Vui!

PR #998, opened 2018-07-26
Remove Search Icon For Non-Searchable Catalogs #1008

vuinguyen:
This branch addresses JIRA issue: http://jira.nypl.org/browse/SIMPLY-1008

where the search icon (magnifying glass) appeared in the navigation bar of a feed that doesn't have search support. Before, the search icon would sometimes be visible for a non-searchable feed, but nothing would happen if the user clicked on that icon. This created a poor user experience.

Now, with this fix, the search icon is visible and activated for a searchable feed; and not visible and not activated for a non-searchable feed.

### DIFF
--- a/Simplified/NYPLCatalogGroupedFeedViewController.m
+++ b/Simplified/NYPLCatalogGroupedFeedViewController.m
@@ -87,6 +87,12 @@ static CGFloat const sectionHeaderHeight = 50.0;
     self.navigationItem.rightBarButtonItem.enabled = NO;
     
     [self fetchOpenSearchDescription];
+  } else {
+    self.navigationItem.rightBarButtonItem = [[UIBarButtonItem alloc]
+                                              initWithTitle:@""
+                                              style:UIBarButtonItemStylePlain
+                                              target:nil
+                                              action:nil];
   }
   
   [self downloadImages];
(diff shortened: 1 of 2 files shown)


gioneill:
After some thinking about this, my guess is that the bar button item is not being removed when it should be. It has to do with how the underlying root view controller of the navigation controller forces out certain views on the stack during changes or refreshes.

We should try and see if we can remove them at the appropriate time instead of covering over the bug inside `viewDidLoad()`

As a step 1, I saw we just attempt to set the rightBarButtonItem to nil inside of `viewWillDissappear()`

gioneill:
As stated by Apple’s documentation of the `navigationItem` property, we’re not able to get a reference to an existing bar button item based on how `NYPLRemoteViewController` is adding and removing view controllers to this navigation stack...

vuinguyen:
@gioneill setting rightBarButtonItem to nil in viewDidDisappear() did not fix the problem. :( Any other suggestions I can try? Or is our other option to do a simple patch (with my code) for now, and maybe we can open another issue to do a refactor later (with NYPLRemoteViewController) or ...?

gioneill:
@winniequinn self-assigned herself, so we'll just keep it on hold until she has a chance to weigh in.

winniequinn:
Thank you for your investigation on this one, Vui!

PR #999, opened 2018-07-29
Fix inability to scroll to bottom of My Books

winniequinn:
This issue only affected devices running iOS 11 or later.


winniequinn:
This fix has been tested on iOS 8.4 and iOS 11.4.

gioneill:
We probably need the same change in My Reservations, yes?

winniequinn:
@gioneill I don't believe so. There are no facets in Reservations and thus no manual content inset adjustments.

PR #1001, opened 2018-08-02
Support iOS 12

winniequinn:
This fixes the following:

* Long-load detection failing to notice when a page is ready
* Page transitions not working correctly
* Tap-to-turn only working at the very edges of the screen
* Navigation toggling not working
* OpenDyslexic typeface selection not working
* Link-time failures


Simplified.xcodeproj/project.pbxproj
```diff
@@ -617,7 +616,6 @@
 				08A352271BDE91B80040BF1D /* libRDServices.a in Frameworks */,
 				08A352221BDE8E560040BF1D /* libicucore.tbd in Frameworks */,
 				5AEA9E011B947419009F71DB /* libc++.dylib in Frameworks */,
-				5A5B90181B946D8B002C53E9 /* libstdc++.6.0.9.dylib in Frameworks */,
```
gioneill: What was the purpose of this?

Simplified/Simplified.js
```diff
@@ -152,6 +158,10 @@ function Simplified() {
   this.linkOpenDyslexicFonts = function() {
     var id = 'simplified-opendyslexic';
     var innerDocument = window.frames['epubContentIframe'].document;
+    if (!innerDocument) {
```
gioneill: Should we also check the OS along with a lack of `innerDocument`? Is there any possibility of other changes to the app down the road causing this check to actually be ambiguous with the assumption that it's guaranteeing an iOS level?

winniequinn: I'm going to leave this as-is, as discussed!

Issue #706, opened 2017-08-15
Explicitly terminating app causing cpp exception (top crash in analytics)


gioneill:
Steps to reproduce:

1. Open any DRM book
2. Hit back button
3. Double tap and swipe up to terminate app

```libc++abi.dylib: terminating with uncaught exception of type std::__1::system_error: mutex lock failed: Invalid argument```

Notes:
- A crash will not occur without step 1.
- A crash will not occur if the user goes to another app, then later terminates the app.
- Seemingly any other action or length of time after Step 2 does not seem to prevent the crash.
- The Book Registry DOES seem to save successfully in my limited testing.

winniequinn:
This seems to be effectively harmless but I will give it a look when I get a chance!

PR #1002, opened 2018-08-02
Fix Xcode Analyze Warnings #706

vuinguyen:
This PR fixes a variety of Xcode warnings, and in particular, warnings from running Analyze in Xcode, as mentioned in this JIRA issue:

http://jira.nypl.org/browse/SIMPLY-706

There are a couple of "Dead store" Analyze warnings left that have not been resolved, as shown in the screenshot below:
screen shot 2018-08-02 at 1 33 19 pm

The "Dead store" warnings point to attributes that are stored in objects, that are never used or referenced again in code. Perhaps they were stored there for a reason, such as for future development?

I don't know how or if those warnings should be fixed in this PR, or if a new issue should be created to deal with those later. My recommendation would be to create a new issue to work on the "Dead store" warnings later. So for now, they will not be addressed in this current PR. 

In the same vein, there were a lot of non-Analyze "Deprecations" warnings, which I also recommend writing another issue to deal with those later for the same reasons -- the deprecated code may still be there for a reason and it may be disastrous to simply take them out.


Simplified/Accounts.swift
```diff
@@ -13,7 +13,7 @@ final class AccountsManager: NSObject
   static let shared = AccountsManager()
   
   // For Objective-C classes
-  @objc class func sharedInstance() -> AccountsManager
+  @objc @discardableResult class func sharedInstance() -> AccountsManager
```
gioneill: This is my fault based on how the method is being used in the app delegate. I'm not saving the singleton the first time we run it.
@winniequinn Let me know a better way to do this.. I see an example of running calling `initialize` but I'm not sure the behavior here would be equivalent.

Simplified/NYPLNetworkQueue.swift
```diff
@@ -8,7 +8,13 @@ import SQLite
  */
 final class NetworkQueue: NSObject {
 
-  static let shared = NetworkQueue()
+  static let sharedInstance = NetworkQueue()
+
+  // For Objective-C classes
+  @objc @discardableResult class func shared() -> NetworkQueue
+  {
+    return NetworkQueue.sharedInstance
```
gioneill: Same comment here as above. @winniequinn Your guidance on the proper usage in `didFinishLaunching..` would be appreciated!

Simplified/NYPLReaderTOCViewController.m
```diff
@@ -163,7 +163,7 @@ - (UITableViewCell *)tableView:(__attribute__((unused)) UITableView *)tableView
       }
       return cell;
     }
-    case 1:{
+    default: {
```
gioneill: Our TOC has two segments right now, and if we have the default being the "2nd" segment, then this would turn into a bug if that segment control ever changes by another developer. Since for now we essentially want to quickly find this as an error, we could actually invoke a crash in the default case by throwing an exception:
```
default:
  @throw NSInvalidArgumentException;
```

Simplified/Accounts.swift
winniequinn: @gioneill: It's not obvious to me that calling `AccountsManager.sharedInstance` has an effect. What is the reason we're presently invoking it here?

https://github.com/NYPL-Simplified/Simplified-iOS/blob/adaef5ec3ad0bf56889cfbf5dc2875bf9ecd2c2b/Simplified/NYPLAppDelegate.m#L42

winniequinn: In any case, we do not want to mark `sharedInstance` as discardable. Doing so may mask a bug in our code because the only reason to get the shared instance should be to do something with it.

Simplified/NYPLAccountSignInViewController.m
```diff
@@ -831,7 +831,7 @@ - (void)validateCredentials
   NSURLSessionDataTask *const task =
     [self.session
      dataTaskWithRequest:request
-     completionHandler:^(NSData *data,
+     completionHandler:^(NSData *data __attribute__((unused)),
```
winniequinn: This notation is more concise, so let's use this instead:

```Objective-C
NSURLSessionDataTask *const task =
  [self.session
   dataTaskWithRequest:request
   completionHandler:^(__unused NSData *data,

File: Simplified/NYPLBookDetailDownloadingView.m [hunk 5]
```diff
@@ -37,7 +37,7 @@ - (instancetype)init
   self.percentageLabel.font = [UIFont systemFontOfSize:14];
   self.percentageLabel.textColor = [NYPLConfiguration backgroundColor];
   self.percentageLabel.textAlignment = NSTextAlignmentRight;
-  self.percentageLabel.text = @"0%";
+  self.percentageLabel.text = NSLocalizedString(@"PercentageLabelZero", nil);
```
winniequinn: It probably doesn't make sense to localize numbers like this, especially since this percentage is going to be dynamically updated. Instead, I would suggest using a wrapper function that explicitly disables the localization warning via `__attribute__((annotate("returns_localized_nsstring")))`. We can write that function once—maybe call it `NYPLDoNotLocalize`—and reuse it. See [Clang's documentation](https://clang-analyzer.llvm.org/faq.html#unlocalized_string) for more information.

Simplified/NYPLBookDownloadingCell.m
```diff
@@ -63,7 +63,7 @@ - (void)layoutSubviews
                                            CGRectGetHeight(self.downloadingLabel.frame));
   
   NSString *const percentageLabelText = self.percentageLabel.text;
-  self.percentageLabel.text = @"100%";
+  self.percentageLabel.text = NSLocalizedString(@"PercentageLabelHundred", nil);
```
winniequinn: See above.

Simplified/NYPLBookDownloadingCell.m
```diff
@@ -122,7 +122,7 @@ - (void)setup
   self.percentageLabel.font = [UIFont systemFontOfSize:12];
   self.percentageLabel.textColor = [NYPLConfiguration backgroundColor];
   self.percentageLabel.textAlignment = NSTextAlignmentRight;
-  self.percentageLabel.text = @"0%";
+  self.percentageLabel.text = NSLocalizedString(@"PercentageLabelZero", nil);
```
winniequinn: See above.

Simplified/NYPLNetworkQueue.swift
winniequinn: @gioneill + @vuinguyen: See above.

Simplified/NYPLReaderTOCViewController.m
winniequinn: Seconding @gioneill on this one. Even better would be to have a function of type `(NSInteger) -> SegmentedControlType` where `SegmentedControlType` is an enum we defined. We could then switch on the result of applying the function to the integer and only have to handle cases explicitly defined, thus obviating the need for an unsafe default case here and making the integer check explicit in the conversation function.

Simplified/NYPLSettingsAccountDetailViewController.m
```diff
@@ -479,7 +479,7 @@ - (void)validateCredentials
   NSURLSessionDataTask *const task =
   [self.session
    dataTaskWithRequest:request
-   completionHandler:^(NSData *data,
+   completionHandler:^(NSData *data __attribute__((unused)),
```
winniequinn: See above.

Simplified/NYPLAccountSignInViewController.m
vuinguyen: so @winniequinn you're saying to use the unused attribute instead of discardable?

Simplified/NYPLBookDetailDownloadingView.m
vuinguyen: @winniequinn what file would be a good place to implement global functions like NYPLDoNotLocalize?

Simplified/NYPLAccountSignInViewController.m
winniequinn: I'm just suggesting we use `__unused` instead of `__attribute__((unused))` as per the example.

Simplified/NYPLBookDetailDownloadingView.m
winniequinn: Let's make `NYPLLocalization.m`  and `.h` with a global `NYPLLocalizationNotNeeded` function.

Simplified/NYPLNetworkQueue.swift
vuinguyen: @winniequinn my same question as above

Simplified/Accounts.swift
gioneill: You're right @winniequinn! Line 42 and any comment above it should be deleted.

Simplified/NYPLNetworkQueue.swift
vuinguyen: @gioneill How about I leave the discardableResult in this file, NYPLNetworkQueue.swift, and in Accounts.swift (above), and you can create a new issue / PR to remove them and take care of that?

Simplified/Accounts.swift
winniequinn: @vuinguyen: We do not want to mark it as discardable, no. Instea,d we want to change the interface of the class so it no longer requires that we discard the result of `sharedInstance` to initialize the class in the first place.

I spoke with @gioneill about this and I think we're both on the same page. Please work with him to get this fixed!

Simplified/NYPLNetworkQueue.swift
gioneill: For now just delete `@discardableResult` in both places, and I'll address it in another PR. Thanks Vui!

winniequinn:
Thanks so much, @vuinguyen! Merging with just a few small changes! Feel free to see the commits if you're curious.

vuinguyen:
Thanks @winniequinn ! I've created 2 new issues that came out of working on this one (mentioned in my first comment on this PR); just so they will get addressed eventually (they're outside the scope of this PR) and not be forgotten.

https://jira.nypl.org/browse/SIMPLY-1138
https://jira.nypl.org/browse/SIMPLY-1139

PR #1003, opened 2018-08-06
Rework account switch; pull accounts from Settings

winniequinn:
Previously, switching an account would reuse the top-level view controller in Catalog. This caused various bugs such as the search button not being correctly set after changing libraries. This commit reworks account switching so that `NYPLCatalogNavigationController` throws away its view controllers and starts fresh with a new one, thus eliminating various problems.

In addition, this commit also moves much accounts-related functionality from `NYPLSettings` to its rightful place in `AccountsManager`.

### DIFF
--- a/Simplified/NYPLSettings.m
+++ b/Simplified/NYPLSettings.m
@@ -5,8 +5,6 @@
 #import "NYPLConfiguration.h"
 #import "SimplyE-Swift.h"
 
-static NSString *const currentAccountIdentifierKey = @"NYPLCurrentAccountIdentifier";
-
 static NSString *const customMainFeedURLKey = @"NYPLSettingsCustomMainFeedURL";
 
 static NSString *const accountMainFeedURLKey = @"NYPLSettingsAccountMainFeedURL";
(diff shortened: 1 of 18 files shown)


Podfile.lock
```diff
@@ -58,6 +58,14 @@ DEPENDENCIES:
   - SQLite.swift (~> 0.11.4)
   - ZXingObjC (~> 3.2.1)
 
+SPEC REPOS:
```
gioneill: @winniequinn What is a spec repo?

winniequinn: Ah, don't worry about that. This is purely a work-in-progress PR so I can link to it from elsewhere. I will rebase and remove the changes to this file before submitting it for review.

Simplified/NYPLSettings.m
```diff
@@ -109,12 +98,13 @@ - (BOOL) userHasSeenFirstTimeSyncMessage
   return [[NSUserDefaults standardUserDefaults] boolForKey:userSeenFirstTimeSyncMessageKey];
 }
 
+// FIXME: This should be in `AccountsManager`, not `NYPLSettings`.
```
winniequinn: This will have to wait for another day. The main issue is that so many parts of the app use `Int` account IDs instead of `Account` objects.

winniequinn:
@gioneill This is now ready for review!

PR #1004, opened 2018-08-13
Integrate PDF Reader with No Data Saving/Synching

vuinguyen:
This PR addresses JIRA issue: https://jira.nypl.org/browse/SIMPLY-175
by enabling a PDF renderer to open and display PDF books within SimplyE. 

Data (ie bookmarks and annotations) is not being saved locally in between app restarts, nor is data being saved or synced to a server.


MalcolmMcFly:
@kyles-ep has it been discussed what we are doing with this?

kyles-ep:
@MalcolmMcFly I don't have the necessary historic knowledge to comment on this PR. It looks like this is for adding a PDF viewer, but we already have one in the app.

MalcolmMcFly:
@kyles-ep so I think I am going to close this PR if that is ok with you.

swandog30:
This was work that Minitex did last year integrating a commercial PDF renderer.  Go ahead and close it.

PR #1005, opened 2018-08-16
Fix issues with the reader's menu controller

winniequinn:
The following problems have been solved:

- Tapping when the menu is up no longer toggles the UI.
- Selecting text no longer toggles the UI.
- Selecting text from the edge of the screen no longer turns pages.
- Tapping now reliably clears the menu and text selection.
- Turning a page now dismisses the menu and text selection.
- Copying text now dismisses text selection.

### DIFF
--- a/Simplified/NYPLReaderReadiumView.m
+++ b/Simplified/NYPLReaderReadiumView.m
@@ -225,6 +225,17 @@ static void generateTOCElements(NSArray *const navigationElements,
    selector:@selector(didBecomeActive)
    name:UIApplicationDidBecomeActiveNotification
    object:nil];
+  
+  [[NSNotificationCenter defaultCenter]
+   addObserver:self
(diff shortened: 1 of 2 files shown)


Simplified/NYPLReaderReadiumView.m
```diff
@@ -343,6 +354,12 @@ - (void) openPageRight {
   [self sequentiallyEvaluateJavaScript:@"ReadiumSDK.reader.openPageRight()"];
 }
 
+/// Toggles user interaction to ensure text selections are cleared.
+- (void)clearTextSelection {
+  self.webView.userInteractionEnabled = !self.webView.userInteractionEnabled;
+  self.webView.userInteractionEnabled = !self.webView.userInteractionEnabled;
+}
```
gioneill: If you were repeating code in more than two places, I would refactor that into a method next time.

Issue #735, opened 2017-09-01
Split View: Glitch in UIKit delegate


gioneill:
By a certain sequence of rotations and split view changes, the system will swap height and width on the expected screen frame. It only seems occur when the BookDetailViewController is on screen from the main catalog, but also seems like a system-level issue that it cannot understand the correct size.

It can be alleviated by sliding the app to full screen, then rotating once more.

Steps to reproduce:
- Open SimplyE in landscape full screen
- Slide in another app from the right
- Tap the divider marker to fix both apps side by side at the 70/30 split
- Turn to portrait, keeping both apps split, but observe that SimplyE should transition to a compact layout
- Take the divider and slide the right-side alternate app all the way full screen (still in portrait)
- Now slide again from the right, and select SimplyE
- BUG: the system thinks the view's frame is "height x width" when it should actually be "width x height"

Example:
![screen shot 2017-08-31 at 7 06 33 pm copy](https://user-images.githubusercontent.com/6179695/29973928-34c2df70-8eff-11e7-8fa8-12846a6340e2.png)

PR #1007, opened 2018-08-24
Set Publication Date from dc:issued #735

vuinguyen:
This PR addresses the JIRA issue: https://jira.nypl.org/browse/SIMPLY-735

by setting the published date with the value from dc:issued instead of atom:published

### DIFF
--- a/Simplified/NSDate+NYPLDateAdditions.m
+++ b/Simplified/NSDate+NYPLDateAdditions.m
@@ -20,6 +20,24 @@
   return date;
 }
 
++ (NSDate *)dateWithDateOnlyString:(NSString *const)string
+{
+  NSDateFormatter *const dateFormatter = [[NSDateFormatter alloc] init];
+
+  dateFormatter.locale = [[NSLocale alloc] initWithLocaleIdentifier:@"en_US_POSIX"];
+  dateFormatter.dateFormat = @"yyyy'-'MM'-'dd";
+
+  NSDate *const date = [dateFormatter dateFromString:string];
+
+  // in case date came in as a year only format
+  if(!date) {
+    dateFormatter.dateFormat = @"yyyy";
+    return [dateFormatter dateFromString:string];
+  }
+
+  return date;
+}
+
 - (NSString *)RFC3339String
 {
   NSDateFormatter *const dateFormatter = [[NSDateFormatter alloc] init];
--- a/Simplified/NSDate+NYPLDateAdditions.m
+++ b/Simplified/NSDate+NYPLDateAdditions.m
@@ -20,17 +20,18 @@
   return date;
 }
 
-+ (NSDate *)dateWithDateOnlyString:(NSString *const)string
++ (NSDate *)dateWithISO8601DateString:(NSString *const)string
 {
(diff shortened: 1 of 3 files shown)


vuinguyen:
@winniequinn : I've pushed some changes up and am ready for you to take another look! Let me know if you have any questions

aslagle:
Is this done?

Simplified/NSDate+NYPLDateAdditions.m
```diff
@@ -20,6 +20,25 @@ + (NSDate *)dateWithRFC3339String:(NSString *const)string
...
++ (NSDate *)dateWithISO8601DateString:(NSString *const)string
+{
+  NSISO8601DateFormatter *const ISODateFormatter = [[NSISO8601DateFormatter alloc] init];
+  
+  ISODateFormatter.formatOptions = NSISO8601DateFormatWithFullDate;
+  ISODateFormatter.timeZone = [NSTimeZone timeZoneForSecondsFromGMT:0];
+
+  NSDate *const date = [ISODateFormatter dateFromString:string];
+
+  // in case date came in as a year only format
```
gioneill: So we try and make a date from the ISO formatter. If that ends up being nil it means we've failed. So we try again by guessting that we have a year format like "yyyy". That could also fail. Which is fine but let's make sure that a `nil` propogating down is tested. (The comment is misleading. I'd just remove it).

Simplified/NYPLOPDSEntry.m
```diff
@@ -135,9 +135,9 @@ - (instancetype)initWithXML:(NYPLXML *const)entryXML
   self.providerName = [entryXML firstChildWithName:@"distribution"].attributes[@"bibframe:ProviderName"];
   
   {
-    NSString *const dateString = [entryXML firstChildWithName:@"published"].value;
+    NSString *const dateString = [entryXML firstChildWithName:@"issued"].value;
     if(dateString) {
-      self.published = [NSDate dateWithRFC3339String:dateString];
+      self.published = [NSDate dateWithISO8601DateString:dateString];
```
gioneill: Test this code for me by changing this line to 
`self.published = nil` and make sure the UI still looks okay.

Simplified/NSDate+NYPLDateAdditions.m
vuinguyen: Easy enough. I've removed the comment!

Simplified/NYPLOPDSEntry.m
vuinguyen: When there is a valid published date, it will look like this:
simulator screen shot - iphone 8 plus - 2018-11-12 at 14 30 47

To perform that nil test, when the published date is nil, the screen for the same book looks like this:
nodate

Which is similar to what it would look like for a book that typically has no published date, such as this one from the publisher Project Guttenberg (no Published line gets displayed at all):
simulator screen shot - iphone 8 plus - 2018-11-12 at 14 31 33

@gioneill Let me know if you have any additional comments/questions. Thanks Greg!

gioneill:
Thank you, @vuinguyen

PR #1008, opened 2018-09-25
Update To Swift 4 #1109

vuinguyen:
This PR addresses the following JIRA issue: https://jira.nypl.org/browse/SIMPLY-1109 , 
where the SimplyE code base needs to be updated to Swift 4.

I mostly ran the automatic Swift 4 converter, which made many of the changes, including updating the project settings. However, because @objc inference has been reduced in Swift 4, there were some manual changes as well, in particular with the use of @objcMembers to allow functions and properties of Swift classes to be exposed to Objective-C. See the following stackoverflow:

https://stackoverflow.com/questions/45656671/unable-to-access-swift-4-class-from-objective-c-property-not-found-on-object-o/45656756


Simplified/NYPLBarcode.swift
```diff
@@ -55,7 +55,7 @@ final class NYPLBarcode: NSObject {
 
   class func presentScanner(withCompletion completion: @escaping (String?) -> ())
   {
-    AVCaptureDevice.requestAccess(forMediaType: AVMediaTypeVideo) { granted in
+    AVCaptureDevice.requestAccess(for: AVMediaType(rawValue: convertFromAVMediaType(AVMediaType.video))) { granted in
```
gioneill: I think this can all be rewritten as:
`AVCaptureDevice.requestAccess(for: .video) { granted in`
Let me know if that works.

Simplified/Accounts.swift
```diff
@@ -191,7 +191,7 @@ final class Account:NSObject
     }
   }
   
-  func getLicenseURL(_ type: URLType) -> URL? {
+  @objc func getLicenseURL(_ type: URLType) -> URL? {
```
gioneill: If I'm reading the proposal correctly, it seems like `@objcMembers` is a new keyword assigned to what `@obj` _used_ to do at the class level. Does this function work without the `@objc` marker? It's using an enum that has also been marked `@objc`
@winniequinn

Simplified/NYPLBarcode.swift
```diff
@@ -103,3 +103,8 @@ final class NYPLBarcode: NSObject {
     }
   }
 }
+
+// Helper function inserted by Swift 4.2 migrator.
```
gioneill: If the above works, you can remove this method from the migrator.

Simplified/NYPLBookDetailTableView.swift
```diff
@@ -105,7 +105,7 @@ final class NYPLBookDetailTableViewDelegate: NSObject, UITableViewDataSource, UI
     refresh()
   }
   
-  func updateFonts() {
+  @objc func updateFonts() {
```
gioneill: Same as above, I'm suspicious if `@objc` is needed if the entire class is marked `@objcMembers` Just let me know if it works after removing these on the functions.

Simplified/NYPLBookDetailTableView.swift
```diff
@@ -171,7 +171,7 @@ final class NYPLBookDetailTableViewDelegate: NSObject, UITableViewDataSource, UI
     }
   }
 
-  func moreBooksTapped(sender: UIButton) {
+  @objc func moreBooksTapped(sender: UIButton) {
```
gioneill: Same as above.

Simplified/NYPLReportIssueViewController.swift
```diff
@@ -83,7 +83,7 @@ class NYPLReportIssueViewController: UIViewController, UITableViewDataSource, UI
   }
   
   
-  func submitPressed(sender:Any)
+  @objc func submitPressed(sender:Any)
```
gioneill: Same as above.

Simplified/RemoteHTMLViewController.swift
```diff
@@ -79,7 +79,7 @@ final class RemoteHTMLViewController: UIViewController, WKNavigationDelegate {
         decisionHandler(.cancel)
       } else {
         if #available(iOS 10.0, *) {
-          UIApplication.shared.open(url, options: [:], completionHandler: nil)
+          UIApplication.shared.open(url, options: convertToUIApplicationOpenExternalURLOptionsKeyDictionary([:]), completionHandler: nil)
```
gioneill: Wow I'm not sure what the migrator had in mind here 😄 
I checked the method docs for UIApplication [https://developer.apple.com/documentation/uikit/uiapplication/1648685-open](here) and it looks like both `options` and `completionHandler` have the same default values, so what SHOULD work is:
```
UIApplication.shared.open(url)
```
Let me know if that works.

Simplified/RemoteHTMLViewController.swift
```diff
@@ -94,3 +94,8 @@ final class RemoteHTMLViewController: UIViewController, WKNavigationDelegate {
     activityView(false)
   }
 }
+
+// Helper function inserted by Swift 4.2 migrator.
```
gioneill: Remove this after testing the above.

vuinguyen:
@gioneill Greg, thanks for the review! I was able to implement all of your suggested changes without any problems, except for the enum in Accounts.swift. The @objc keyword had to stay there, or else it wouldn't compile.

Simplified.xcodeproj/project.pbxproj
```diff
@@ -1747,7 +1747,8 @@
 				PRODUCT_BUNDLE_IDENTIFIER = org.nypl.SimplyETests;
 				PRODUCT_NAME = "$(TARGET_NAME)";
 				SWIFT_OPTIMIZATION_LEVEL = "-Onone";
-				SWIFT_VERSION = 3.0;
+				SWIFT_SWIFT3_OBJC_INFERENCE = On;
```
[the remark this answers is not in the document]
vuinguyen: I changed the ```SWIFT3_OBJC_INFERENCE``` to "Default" for the SimplyETests target, just now. But I only see 2 targets, the other being SimplyE, which I've already changed. Am I missing something? Pushing up my changes right after this comment here ...

gioneill:
I made a small commit to put everything consistent with the usage of `@objMembers` to accomplish what `@objc` _used_ to do. I built and ran with the updated APIKeys.swift just to make sure everything worked okay; a quick smokescreen.

Thank you for your work on this, Vui! I'm approving and merging.

PR #1009, opened 2018-10-05
Entry Point support in SimplyE Catalogs (Audiobooks)

gioneill:


Simplified/NYPLCatalogUngroupedFeedViewController.m
```diff
@@ -76,15 +78,27 @@ - (void)updateActivityIndicator
...
   [super viewDidLoad];
-  
+
+  [self configureEntryPoints:self.feed.entryPoints];
+
   self.facetBarView = [[NYPLFacetBarView alloc] initWithOrigin:CGPointZero width:0];
   self.facetBarView.facetView.dataSource = self;
   self.facetBarView.facetView.delegate = self;
   [self.view addSubview:self.facetBarView];
+  [self.facetBarView autoPinEdgeToSuperviewEdge:ALEdgeLeading];
```
gioneill: This just happened to be an existing bug that kept a blank toolbar on screen if an ungrouped feed vc didn't have any facet groups. The special init method is kept for its usage relevant to the My Books VC.

PR #1012, opened 2018-10-14
Simply-1228: remove email char filter

jrryhrtn:
Removed code that prevented the entry of the "@" sign and the "." required to use email addresses as a username.


jrryhrtn:
Closing and resubmitting...

PR #1013, opened 2018-10-14
Simply-1228: remove email char filtering code

jrryhrtn:
Removed filter  that prevents user from entering "@" and "." required to use email address as username


gioneill:
Eventually we'll need to support emails for login to SimplyE. Tracked here:

PR #1016, opened 2018-11-16
Delete audiobooks on sign-out

winniequinn:


gioneill:
I've verified that audiobook content gets successfully deleted even from an account that is currently not set as the "active" library account.

PR #1017, opened 2018-11-23
add objcMembers to fix swift 4/xcode 10 build errors

elindsey:
@objc inference was (thankfully) dropped in swift 4; now must be explicit. This fixes build errors I hit when doing a fresh checkout of the project.


winniequinn:
Thanks, @elindsey!

PR #1020, opened 2018-12-21
Send Notifications For Reserved Items #1345

vuinguyen:
This PR addresses the JIRA issue:
https://jira.nypl.org/browse/SIMPLY-1345

which will send a notification when a reserved item is first available to checkout, and a second (final) notification when the patron has 24 hours left to checkout the reserved item.

This feature was briefly tested on an iOS 9 simulator (iPhone 6s plus) and did not crash for notifications (which works only for iOS 10 and above). It simply does not send any notifications at all for iOS 9.

Since the BookRegistryRecord is being modified to save the notification sent state, some kind of data migration plan must be in place before this can go out to production.

@gioneill I will be out for the next couple of weeks, and will address any comments/feedback on this PR when I get back. Thanks! Happy holidays and happy new year, Greg!


gioneill:
Notifications merged in this PR:
https://github.com/NYPL-Simplified/Simplified-iOS/pull/1025

PR #1028, opened 2019-01-17
Support smart invert colors

winniequinn:
This prevents cover images from being inverted when using iOS 11's smart invert color feature.

**NOTE:** I was not able to test this change, although I'd be happy to do so if I could receive an up-to-date `Accounts.json`. I'm especially unsure about `blurCoverImageView`.


winniequinn:
Happy to test this myself if I can receive an up-to-date `Accounts.json`.

PR #1038, opened 2019-02-15
Display Helpful Error Msg When Borrowing Already Checked Out Book #1613

vuinguyen:
This PR addresses the issue raised in JIRA #1613: 
https://jira.nypl.org/browse/SIMPLY-1613

where, when a patron tries to check out a book from multiple devices, they will receive an unhelpful error message on the 2nd device.

I try to address this problem by displaying a more useful error message to the patron.

Screenshots for before and after the fix, are in the JIRA issue.

### DIFF
--- a/Simplified/NYPLMyBooksDownloadCenter.m
+++ b/Simplified/NYPLMyBooksDownloadCenter.m
@@ -609,7 +609,7 @@ didDismissWithButtonIndex:(NSInteger const)buttonIndex
         // create an alert to display for error, feed, or feed count conditions
         NYPLAlertController *alert = [NYPLAlertController alertWithTitle:@"BorrowFailed" message:@"BorrowCouldNotBeCompletedFormat", book.title];
 
-        // display different message for special type of error or just add document message for generic error
+        // set different message for special type of error or just add document message for generic error
         if (error) {
           if([error[@"type"] isEqualToString:NYPLProblemDocumentTypeLoanAlreadyExists]) {
             alert = [NYPLAlertController alertWithTitle:@"BorrowFailed"
@@ -621,6 +621,7 @@ didDismissWithButtonIndex:(NSInteger const)buttonIndex
           }
         }
 
+        // display the alert
         [alert presentFromViewControllerOrNil:nil animated:YES completion:nil];
       }];
       return;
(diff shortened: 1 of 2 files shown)


Simplified/NYPLMyBooksDownloadCenter.m
```diff
@@ -605,9 +605,20 @@ - (void)startBorrowForBook:(NYPLBook *)book
...
+          BOOL displayDocumentMessage = YES;
+          if([error[@"type"] isEqualToString:NYPLProblemDocumentTypeLoanAlreadyExists]) {
+            message = NSLocalizedString(@"You have already checked out this loan. You may need to refresh your My Books list to download the title.",
+                                        comment: @"When book is already checked out on patron's other device(s), they will get this message");
+            displayDocumentMessage = NO;
+          }
+          NYPLAlertController *const alert = [NYPLAlertController alertWithTitle:@"BorrowFailed" message:message, book.title];
+          [alert setProblemDocument:[NYPLProblemDocument problemDocumentWithDictionary:error] displayDocumentMessage:displayDocumentMessage];
+          [alert presentFromViewControllerOrNil:nil animated:YES completion:nil];
+        }
```
gioneill: I think you probably want it like this:
```
            [alert setProblemDocument:[NYPLProblemDocument problemDocumentWithDictionary:error] displayDocumentMessage:displayDocumentMessage];
        }
        [alert presentFromViewControllerOrNil:nil animated:YES completion:nil];
```
In Objective-C, a one line if statement ONLY applies to that line, so:
`if (error) someThing()`
is equivalent to:
```
if (error) {
    someThing();
}
```

vuinguyen: The code I have above is exactly what I want to happen though.

What I'm doing there is:

1. If we have an error (line 609), set default values for the error message and whether or not to display the document message
(lines 610 and 611)

2. inside the if block for error, if we have a specific error (line 612), change the values for the error message and boolean for displaying the document message (lines 613-615)
3. Create an alert object, and display the proper error message and possibly also the  document message depending on if it was a general error or the specific error and then display the alert (lines 617-619).

Let me know if this all makes sense! We can also do a quick phone chat if that would be easier.

vuinguyen: Thanks @gioneill  for doing a quick phone chat with me. I think I understand your comment better now.

What I did was modify the code to ensure that an alert still displays for error, as well as incorrect feed and feed count conditions. Also, I added some comments to the code to make it clearer what's going on now.

PR #1044, opened 2019-04-26
Fixed logic in WebViewDelegate to show html content

swandog30:
This fixes a bug where HTML content wasn't shown properly because of the incorrect filtering of links in the NavigationDelegate for the view.  Links in the content will still open in the devices web browser and not in the application.  This replicates the logic that is used in RemoteHTMLViewController.swift.  In other words, the Software Licenses page will show properly.

Alternatives are to render software-licenses.html from an online source and not bundle it with the app.


leonardr:
Resolves https://jira.nypl.org/browse/SIMPLY-1896

PR #1045, opened 2019-05-08
Don't show DRM PDFs

benanderman:
Also, don't crash on book with no valid acquisition paths, and show beta button on Simulator.

On removing the log / throw: I think it's the wrong place to log, since semantically it makes sense that we could ask for the default content type of a book that we don't have a valid acquisition path for. It's also inconsistent, since we would just silently return `NYPLBookContentTypeUnsupported` if we had a path but it was of an unsupported type (which should have already been filtered out before that point). I think if we wanted to log (to bugsnag for instance), it should happen somewhere where we actually don't expect to have an unsupported book.

### DIFF
--- a/Simplified/NYPLBook.m
+++ b/Simplified/NYPLBook.m
@@ -563,11 +563,11 @@ static NSString *const UpdatedKey = @"updated";
     [finalTypes addObject:path.types.lastObject];
   }
 
+  NYPLBookContentType defaultType = NYPLBookContentTypeUnsupported;
   if (finalTypes.count == 1) {
-    return NYPLBookContentTypeFromMIMEType(finalTypes.firstObject);
+    defaultType = NYPLBookContentTypeFromMIMEType(finalTypes.firstObject);
   } else if (finalTypes.count > 1) {
     // Defualt to epub if it exists, else assign a random supported type if one exists.
-    NYPLBookContentType defaultType = NYPLBookContentTypeUnsupported;
     for (NSString *const type in finalTypes) {
       NYPLBookContentType const contentType = NYPLBookContentTypeFromMIMEType(type);
       if (contentType != NYPLBookContentTypeUnsupported) {
@@ -577,11 +577,9 @@ static NSString *const UpdatedKey = @"updated";
         }
       }
     }
-    return defaultType;
   }
-
-  NYPLLOG(@"Invalid argument: No mime type in acquisition path/s");
-  @throw NSInvalidArgumentException;
+  
+  return defaultType;
 }
 
 @end
--- a/Simplified/NYPLBook.m
+++ b/Simplified/NYPLBook.m
@@ -564,17 +564,13 @@ static NSString *const UpdatedKey = @"updated";
   }
 
   NYPLBookContentType defaultType = NYPLBookContentTypeUnsupported;
-  if (finalTypes.count == 1) {
-    defaultType = NYPLBookContentTypeFromMIMEType(finalTypes.firstObject);
-  } else if (finalTypes.count > 1) {
-    // Defualt to epub if it exists, else assign a random supported type if one exists.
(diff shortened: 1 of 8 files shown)


KSops:
Looks fine to me. I guess the only question is if we need to capture the case when there are no acquisition paths. I'm not sure under what circumstances this can happen. Is that from the feed data being bad/malformed?

leonardr:
It's valid OPDS for a book to have no acquisition paths -- the server need only provide a link. And sometimes we don't even provide a link. Right now, NYPL's production server knows not to tell the client about any PDFs, but this knowledge is hacked in and it does this by serving the OPDS entry for a book and omitting all of the links that would end up in a PDF.

If the book is only available as a PDF, this means the OPDS entry is served with no links at all. That's _not_ valid OPDS but the client should be prepared to handle it.

Simplified/NYPLBook.m
```diff
@@ -563,11 +563,11 @@ - (NYPLBookContentType)defaultBookContentType
     [finalTypes addObject:path.types.lastObject];
   }
 
+  NYPLBookContentType defaultType = NYPLBookContentTypeUnsupported;
   if (finalTypes.count == 1) {
-    return NYPLBookContentTypeFromMIMEType(finalTypes.firstObject);
+    defaultType = NYPLBookContentTypeFromMIMEType(finalTypes.firstObject);
```
leonardr: The logic for a single item in finalTypes seems the same as the logic for multiple items. Is it necessary to treat the one-item case differently?

leonardr:
While talking to Mark I discovered that iOS and Android have inconsistent rules about how to choose from multiple supported acquisition paths. I don't know what the right answer is so I filed a ticket (https://jira.nypl.org/browse/SIMPLY-1940) to cover it rather than asking that you fix it in this PR.

Simplified/NYPLBook.m
benanderman: > It's valid OPDS for a book to have no acquisition paths -- the server need only provide a link. And sometimes we don't even provide a link. Right now, NYPL's production server knows not to tell the client about any PDFs, but this knowledge is hacked in and it does this by serving the OPDS entry for a book and omitting all of the links that would end up in a PDF.
> 
> If the book is only available as a PDF, this means the OPDS entry is served with no links at all. That's _not_ valid OPDS but the client should be prepared to handle it.

In this method, it's valid for there to be no _supported_ acquisition paths, because they got filtered out by `supportedAcquisitionPathsForAllowedTypes:...`. That's why I think it doesn't make sense to throw an exception (or even log anything), since this method was already sometimes returning `NYPLBookContentTypeUnsupported`. The only reason this didn't crash the app before is that we happened not to call `defaultBookContentType` if there weren't any valid acquisition paths.

leonardr: That makes sense.

benanderman: I think you're right about the single / multiple item cases. Thanks for pointing it out; it makes the code much simpler!

leonardr: The Android implementation also has separate but equivalent code for single-item lists, so it's possible that there used to be a difference that has since been removed.

benanderman:
> But I have no idea what our test coverage looks like on iOS. Do we already have tests for these methods? How difficult would it be to write some basic test cases where we run the algorithm on some common acquisition-path scenarios and verify that it chooses the right answer?

As far as I can tell, there is very little test coverage. However, this is one of the few things that IS covered by a test, so I'll update that test. The test wasn't broken by this change, but obviously it also wasn't testing the behavior that was changed. I agree this is a particularly good thing to write a unit test for!

KSops:
I seem to get a different value for CODE_SIGN_IDENTITY. I wonder if perhaps we should omit that change if it will continually be different for each person.

PR #1046, opened 2019-05-13
Update version to 3.1.1

kyles-ep:
Updated version string and updated Carthage dependencies, which were done as part of the Carthage update command


Cartfile.resolved
```diff
@@ -4,6 +4,6 @@ github "NYPL-Simplified/NYPLAEToolkit" "102d0453d1a351de0d3457fc85381e96f8369fe9
 github "NYPL-Simplified/NYPLAudiobookToolkit" "ef856a0c0a96d438311d87b352c23d9ae39eab6c"
 github "NYPL-Simplified/PDFRendererProvider-iOS" "b75f1d24915cba8b255039df951a072d62abdd94"
 github "PureLayout/PureLayout" "v3.1.4"
-github "TheLevelUp/ZXingObjC" "3.5.0"
-github "bugsnag/bugsnag-cocoa" "v5.17.3"
-github "stephencelis/SQLite.swift" "0.11.5"
+github "TheLevelUp/ZXingObjC" "3.6.4"
+github "bugsnag/bugsnag-cocoa" "v5.21.0"
```
leonardr: Does upgrading bugsnag-cocoa fix the Bugsnag issues you mentioned earlier today?

kyles-ep: No, the bugsnag problem is simply that the symbols weren't uploaded. It might be a race condition on when the upload script is run or perhaps just a one-off error in sending them. I have some improvements in mind there but the Bugsnag symbol problem is unrelated to this version update. This is just dependencies that happened to get updated along with the Carthage update command. I think staying up to date with our dependencies is a good idea in general, but it always introduces risk that there's a regression.

leonardr: Joe will perform a general automated regression test before release, so we should catch any problems of that sort.

PR #1048, opened 2019-05-30
[SIMPLY-1952] Ensure PDFs get deleted

kyles-ep:
Currently there's no switch option for deleting PDF books. This is a simple copy-pasta for the EPUB case.

### DIFF
--- a/Simplified/NYPLMyBooksDownloadCenter.m
+++ b/Simplified/NYPLMyBooksDownloadCenter.m
@@ -444,6 +444,7 @@ didDismissWithButtonIndex:(NSInteger const)buttonIndex
            error:&error]){
         NYPLLOG_F(@"Failed to remove local content for download: %@", error.localizedDescription);
       }
+      break;
     }
     case NYPLBookContentTypeAudiobook: {
       NSData *const data = [NSData dataWithContentsOfURL:
@@ -454,6 +455,15 @@ didDismissWithButtonIndex:(NSInteger const)buttonIndex
       id const json = NYPLJSONObjectFromData([NSData dataWithContentsOfURL:
                                               [self fileURLForBookIndentifier:book.identifier account:account]]);
       [[AudiobookFactory audiobook:json] deleteLocalContent];
+      break;
+    }
+    case NYPLBookContentTypePDF: {
+      NSError *error = nil;
+      if (![[NSFileManager defaultManager]
+          removeItemAtURL:[self fileURLForBookIndentifier:identifier account:account]
+          error:&error]) {
+        NYPLLOG_F(@"Failed to remove local content for download: %@", error.localizedDescription);
+      }
       break;
     }
     case NYPLBookContentTypeUnsupported:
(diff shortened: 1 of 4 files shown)


Simplified/NYPLMyBooksDownloadCenter.m
```diff
@@ -444,6 +444,7 @@ - (void)deleteLocalContentForBookIdentifier:(NSString *const)identifier account:
            error:&error]){
         NYPLLOG_F(@"Failed to remove local content for download: %@", error.localizedDescription);
       }
+      break;
```
benanderman: Good catch!

kyles-ep: I added an extra break here because it didn't make sense to me that fallthrough should be allowed. Hopefully this can be sanity-checked.

Simplified/NYPLMyBooksDownloadCenter.m
```diff
@@ -456,6 +457,15 @@ - (void)deleteLocalContentForBookIdentifier:(NSString *const)identifier account:
       [[AudiobookFactory audiobook:json] deleteLocalContent];
       break;
     }
+    case NYPLBookContentTypePDF: {
+      NSError *error = nil;
+      if (![[NSFileManager defaultManager]
+          removeItemAtURL:[self fileURLForBookIndentifier:identifier account:account]
```
benanderman: As long as the PDFs are indeed saved here, it seems good. Ideally (as we talked about), they would have a `.pdf` file extension instead, but that's a much bigger thing to change if we already have PDFs saved this way.

kyles-ep:
> How hard is it to write test coverage for this method?

Good question. I'll look into what I can do for this

SimplifiedTests/NYPLMyBooksDownloadCenterTests.swift
```diff
@@ -0,0 +1,69 @@
...
+    ]
+    for config in configs {
+      let fakeAcquisition = NYPLOPDSAcquisition.init(
+        relation: .generic,
+        type: config["type"]!,
+        hrefURL: emptyUrl,
+        indirectAcquisitions: [NYPLOPDSIndirectAcquisition](),
+        availability: NYPLOPDSAcquisitionAvailabilityUnlimited.init()
+      )
+      let fakeBook = NYPLBook.init(
```
leonardr: Something to consider for the future: in the server code we have a [testing.py file](https://github.com/NYPL-Simplified/server_core/blob/master/testing.py#L306) which has a lot of helper methods for setting up fake data model objects for use in tests.

Right now you're building the iOS test base more or less from scratch, so you don't have anything to build on, but I see the same needs here, and this code looks like it could go into a similar test helper module.

kyles-ep: I thought about a data setup helper, but I was hesitant to make one until I had more test cases to get an idea of what's needed of such a utility class. But there's no harm in making it anyways and modifying it as we go.

leonardr: I agree it's better to generalize from a few cases rather than one and I think this is fine for now.

PR #1050, opened 2019-06-04
Fixed PDFs not being marked as read

kyles-ep:
One-liner for SIMPLY-1979


kyles-ep:
Regarding testing, this would mean we would have to flesh out actual book download/creation helpers in the test suite, since the PDF delegate gets called after the book opens, and I'm sure it'll cause a crash if I use a fake file.

There actually isn't a mirrored structure that exists easily, but if there are more commonalities on a book open in the future, we should certainly look into centralizing those common blocks of code.

PR #1051, opened 2019-06-06
Fixing warnings on splitview controller

kyles-ep:
Minor fix to resolve some issues regarding the view being shown before the viewControllers have been assigned. This was causing warnings to appear primarily when the settings menu was being displayed.


kyles-ep:
@MalcolmMcFly Looks like we take the opportunities as they come. Primarily new files, but I suspect if something requires extensive overhaul, it would be re-written in swift.

PR #1052, opened 2019-06-07
Update submodule

kyles-ep:


kyles-ep:
6041c56 is part of a fix for "No Connection" screen being displayed when it shouldnt

PR #1053, opened 2019-06-13
Updating app version

kyles-ep:


kyles-ep:
I usually save this until I'm sure there's no further changes to the release

PR #1056, opened 2019-06-20
Updating build process

kyles-ep:
Adding `Accounts.json` to the repo as a commit. This is required for 3.1.x -> 3.2.0 migrations and can be phased out later since we no longer use hardcoded lists.  
The Accounts list has been stripped of most information, leaving only enough for id->uuid mapping.  
The build notes were written with effort to keep enough sensitive info out of the process while making things easier to identify and understand.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,5 +11,4 @@ ReaderClientCert.sig
 Simplified/APIKeys.swift
 AudioEngine.json
 Carthage
-Accounts.json
 bugsnag-dsym-upload.rb
--- a/README.md
+++ b/README.md
@@ -24,20 +24,20 @@
 
 In theory, following the instructions in "adobe-rmsdk/RMSDK_User_Manual(obj).pdf", you should be able to build OpenSSL (section 12.1) and cURL (section 12.3) since Adobe provides this package to their developers.
 The following are some smoother steps to achieving this:
-- From `adobe-rmsdk/thirdparty/openssl` run `mkdir public; mv iOS-openssl/ public/ios/`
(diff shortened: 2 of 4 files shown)


malcolmleeo:
@kyles-ep  Going through the process today, will let you know how it goes.

MalcolmMcFly:
@kyles-ep We should consider how we plan to handle step 6. **DRM_Connector_Prerelease** is on a private Google Drive. I have requested access to this file so that I may continue the build process. Should this be apart of the solution we came up with step 5? 

> We can get with Risa to put them in a shared location, then we could put a note in there saying like *contact the current mobile lead for access to these files*

MalcolmMcFly:
More feedback @kyles-ep:

1. Along with adding the notes about contacting the team lead for step 5, lets do that for step 6 too.
2. Please add link for open ssl, I got it from here: https://www.openssl.org/source/old/1.0.1/
3. For the section labeled **Building OpenSSL and cURL** please add the numbers like you have for **Building the Application**, I think that makes it easier to follow.
4. Please add link for curl, I got mine here: https://curl.haxx.se/download/
5. Please add note that both times you run `bash ./build.sh` that it is going to to take quite a bit of time to complete.
6. In this step: 

> Replace adobe-rmsdk/thirdparty/curl/ios/build.sh with build_curl.sh, a fixed version of the build script

Perhaps this should be **_rename_** instead of replace?

Following all of the steps, my build ended in the following error:
iOS Build Error

MalcolmMcFly:
@kyles-ep Now building successfully. Change **Building OpenSSL and cURL** to a numbered list and you have my approval 👍

PR #1060, opened 2019-07-05
SIMPLY-1620

kyles-ep:
Converting some classes to swift and adding problem document description to the alerts that display.


Simplified/NYPLBugsnagLogs.swift
```diff
@@ -170,4 +170,20 @@ fileprivate let tabName = "Extra Data"
       report.addMetadata(metadata, toTabWithName: tabName)
     })
   }
+
```
MalcolmMcFly: Function explaining what this function does and maybe short one-liner about the parameters?

Simplified/NYPLAlertUtils.swift
```diff
@@ -0,0 +1,86 @@
...
+    } else {
+      if document.title != nil {
+        controller.title = document.title
+      }
+      if document.detail != nil {
+        controller.message = document.detail
+      }
+    }
+  }
+
```
MalcolmMcFly: Function explaining what this function does and maybe short one-liner about the parameters?

Simplified/NYPLProblemDocument.swift
```diff
@@ -0,0 +1,34 @@
...
+  fileprivate init(_ dict: [String : Any]) {
+    self.type = dict["type"] as? String
+    self.title = dict["title"] as? String
+    self.status = dict["status"] as? Int
+    self.detail = dict["detail"] as? String
+    self.instance = dict["instance"] as? String
+    super.init()
+  }
+  
+  @objc static func fromData(_ data: Data) throws -> NYPLProblemDocument {
```
MalcolmMcFly: Function explaining what this function does and maybe short one-liner about the parameters?

Simplified/NYPLProblemDocument.swift
```diff
@@ -0,0 +1,34 @@
...
+    let jsonDecoder = JSONDecoder()
+    jsonDecoder.keyDecodingStrategy = .convertFromSnakeCase
+    
+    return try jsonDecoder.decode(NYPLProblemDocument.self, from: data)
+  }
+  
+  @objc static func fromDictionary(_ dict: [String : Any]) -> NYPLProblemDocument {
```
MalcolmMcFly: Function explaining what this function does and maybe short one-liner about the parameters?

Simplified/NYPLAlertUtils.swift
```diff
@@ -0,0 +1,86 @@
...
+    let alertController = UIAlertController.init(
+      title: alertTitle,
+      message: alertMessage,
+      preferredStyle: .alert
+      )
+    alertController.addAction(UIAlertAction.init(title: NSLocalizedString("OK", comment: ""), style: style, handler: nil))
+    return alertController
+  }
+  
+  class func setProblemDocument(controller: UIAlertController?, document: NYPLProblemDocument?, append: Bool) {
```
MalcolmMcFly: Function explaining what this function does and maybe short one-liner about the parameters?

Simplified/NYPLRemoteViewController.m
```diff
@@ -149,19 +147,25 @@ - (void)connectionDidFinishLoading:(__attribute__((unused)) NSURLConnection *)co
...
   self.activityIndicatorLabel.hidden = YES;
-  
-  if ([(NSHTTPURLResponse *)self.response statusCode] != 200
-      && ([self.response.MIMEType isEqualToString:@"application/problem+json"]
-          || [self.response.MIMEType isEqualToString:@"application/api-problem+json"])) {
-    NYPLProblemDocument *problem = [NYPLProblemDocument problemDocumentWithData:self.data];
-    NYPLAlertController *alert = [NYPLAlertController alertWithTitle:problem.title message:problem.detail];
-    [alert setProblemDocument:problem displayDocumentMessage:NO];
+  BOOL mimeTypeMatches = [self.response.MIMEType isEqualToString:@"application/problem+json"] ||
+    [self.response.MIMEType isEqualToString:@"application/api-problem+json"];
```
leonardr: I don't understand the details of this PR, but I think this method is the crucial part of the change, so I'm focused on it and I don't see the change. Previously you were using `problem.detail` -- the piece of text I'd expect to see but wasn't seeing -- and now you're using `pDoc.detail` in the same place. Is there something different about the _surrounding_ code that makes the detail text get rendered?

If the software crashes, I don't think you'll get a problem document -- just the string "Internal Server Error". It would be nice to notice that, but I realize that's not a lot to work with. We have a ticket to improve our handling of crashes (https://jira.nypl.org/browse/SIMPLY-1840), and maybe we could send you a problem detail document even when there's an ISE.

kyles-ep: The primary change lies in `setProblemDocument` whereby the old version omitted the details and the updated one includes it. That particular place you mentioned will not see any change, as you noted.

leonardr: Great, got it, thanks.

PR #1061, opened 2019-07-09
SIMPLY-2098

kyles-ep:


Simplified/NYPLBugsnagLogs.swift
```diff
@@ -170,4 +170,15 @@ fileprivate let tabName = "Extra Data"
       report.addMetadata(metadata, toTabWithName: tabName)
     })
   }
+
```
MalcolmMcFly: Comment

kyles-ep:
@leonardr I'll certainly be making changes to the naming of the class.
There were two references in error messages erroneously mentioning loans feeds which should have been user profile. Aside from that, there are two places where loans feed is referenced:

- BookRegistry for tracking loans
- bugsnag reporting for when checking the loans feed fails

PR #1063, opened 2019-08-01
SIMPLY-2142 - Unify error messages for library fetch errors

kyles-ep:
I couldn't find any instances where I could swap an error message for "disk crashes" or "generic errors" as per https://jira.nypl.org/browse/SIMPLY-2142

### DIFF
--- a/Simplified/NYPLBookRegistry.m
+++ b/Simplified/NYPLBookRegistry.m
@@ -383,7 +383,7 @@ static NSString *const RecordsKey = @"records";
       [self save];
     } else {
       UIAlertController *alert = [NYPLAlertUtils alertWithTitle:@"SyncFailed"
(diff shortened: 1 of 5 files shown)


Simplified/NYPLBookRegistry.m
```diff
@@ -383,7 +383,7 @@ - (void)syncWithStandardAlertsOnCompletion
       [self save];
     } else {
       UIAlertController *alert = [NYPLAlertUtils alertWithTitle:@"SyncFailed"
-                                                        message:@"CheckConnection"];
+                                                        message:@"LibraryLoadError"];
```
leonardr: Since this is getting a loans feed, not a list of libraries, the error message to use is DataError, not LibraryLoadError.

PR #1067, opened 2019-10-08
[SIMPLY-2331] Fixing some colors for Dark Mode

kyles-ep:
Also refactored a class into swift from ObjC


kyles-ep:
@MalcolmMcFly Yes. Dark Mode was introduced in iOS13.

kyles-ep:
Let's keep this PR open for now. I'd like to push a build for people to test the dark colours and see what they think. I have a feeling there may be things to tweak before pushing to production.

PR #1070, opened 2019-10-30
Application Refactor for Merging OpenEBooks

kyles-ep:
**What's this do?**
[Description of what this PR does goes here]

[Quick blurb about why the code is needed and Jira link goes here / Do these changes meet the business requirements of the story?]

[Description of any tests that need to be performed once merged goes here]

[Description of any watchouts, dependencies, or issues we should be aware of goes here]


OpenEbooks-Info.plist
```diff
@@ -0,0 +1,89 @@
...
+	<true/>
+	<key>NSAppTransportSecurity</key>
+	<dict>
+		<key>NSAllowsArbitraryLoads</key>
+		<true/>
+	</dict>
+	<key>NSCameraUsageDescription</key>
+	<string>Open eBooks requires access to the camera in order to scan your library card.</string>
+	<key>NSLocationWhenInUseUsageDescription</key>
+	<string>This service is available to New York State residents only. In order to determine your location, please allow access</string>
```
leonardr: Presumably you need to provide a value for this string, but this particular string is inaccurate for Open eBooks. If you do need a value for this string, it would be better to have the equivalent of "this can't happen" so we can treat it as a bug if it ever shows up.

In the medium term we'll hopefully be removing the native card creator code and this can go away when that does.

Simplified/OpenEBooks/OpenEBooks_OPDS2_Catalog_Feed.json
```diff
@@ -0,0 +1,62 @@
...
+          "type": "application/atom+xml;profile=opds-catalog;kind=acquisition",
+          "rel": "http://opds-spec.org/shelf"
+        },
+        {
+          "href": "https://circulation.openebooks.us/USOEI/patrons/me/",
+          "type": "vnd.librarysimplified/user-profile+json",
+          "rel": "http://librarysimplified.org/terms/rel/user-profile"
+        },
+        {
+          "href": "https://openebooks.net/",
```
leonardr: The HTTPS version of this URL isn't working -- there's a cert problem. I filed https://jira.nypl.org/browse/SEB-1604 to fix the cert and it's probably best to leave this alone rather than switching to the HTTP URL until it's fidxed.

leonardr:
Just so we have a link, this is for https://jira.nypl.org/browse/SIMPLY-2062.

Obviously I'm not qualified to review this whole thing but what I do understand brings up one major concern. The SimplyE mobile app asks for permission to check the patron's location, and the Open eBooks mobile app shouldn't ask for that permission. I noticed that you've got strings talking about the mobile permission in some kind of Open eBooks resource file, and I want to make sure that we don't build an app that asks for a new permission we don't need, potentially freaking out patrons.

OpenEbooks-Info.plist
kyles-ep: Can confirm they require this string even if we don't actually use location services. As per your suggestion, how about "Oops! Looks like OpenEBooks is accidentally requesting location services. Please report a bug!"

killerpuppytails: Since there's no way to submit a bug within the iOS interface that I can find, I vote "Open eBooks never tracks or uses your location. Please take a screenshot and email it to info@openebooks.net so we can fix this error."

ettore:
This PR was open more than 5 months ago and SimplyE has changed a lot since then. For instance the app delegate has been updated, but I see here it has been rewritten in swift. So we would need to integrate all those changes in. Same thing for Accounts.swift, NYPLConfiguration and probably others. 

I also see large sections of code being removed from Navigation controllers, but I don't know why. 

So I don't think I have enough context to review this. We should rebase this branch but since so much time has passed we would really need to revisit all these changes carefully. Otherwise we'll be in regressions hell.

I would leave the Swift rewrites outside of this PR. The reason is it's extremely hard to review changes between 2 classes in different languages, and on top of that across different apps. We can probably use some of the code you wrote here and do those changes individually in a separate PR, and merge those first. Or, do the swift rewrite after the merge. 

The goal here should be to have as few changes as possible in order to minimize merge conflicts and more importantly regressions.

ettore:
Closing since we shipped the refactored OE.

PR #1075, opened 2020-01-22
[SIMPLY-2491] Improving error handling + logging for account login

kyles-ep:


killerpuppytails:
I underline @leonardr 's comment about not sending patron info to Bugsnag. This is super important.

kyles-ep:
We've had two login paths for the longest time and it's been a pain to ensure they're mirroring behaviour. It's possible that there's a discrepancy.

Bugsnag is potentially getting loan URLs and profile URLs to debug what we're trying to hit. Can this contain sensitive or user-specific data that we shouldnt be sending?

leonardr:
No, there are no URLs that SimplyE would know about that contain sensitive patron information.

PR #1077, opened 2020-01-31
Add scripts to build 3rd party dependencies

ettore:
**What's this do?**
Updates steps for building-from-scratch (onboarding scenario) and translates most of them from English to shell scripts. This also removes AudioEngine framework from Carthage management. All dependencies can now be built by simply doing:

$ build-3rd-parties-dependencies.sh <Debug | Release>

The previous build steps no longer worked on Xcode 11.3.1 and Carthage 0.34. They also involved a lot of manual steps, which are error prone and confusing. This could also be useful for CI automation in the future. Moreover, incompatibilities between the current AudioEngine binary and Carthage exist. 

I repeated these steps on a clean checkout of all the involved code. I was able to build the app in both Debug and Release configurations, which one gotcha: the AudioEngine zipfile specified in Certificates repo includes a Release build that does not contain any slice for x86_64/Simulator architectures, causing a Carthage build failure. Therefore, only the `build-3rd-parties-dependencies.sh Debug` invocation works; note that the Debug parameter ONLY means that the debug build of AudioEngine will be used: all other dependencies are built according to Carthage / Adobe build scripts.

This currently require using https://github.com/ettore/Certificates and https://github.com/ettore/NYPLAEToolkit. I have PR open on the related NYPL-Simplified repos.

Yes.

@ettore


README.md
```diff
@@ -1,49 +1,40 @@
...
-    - `SDK_VERSION` reflects the iOS SDK you're using (you can check what you have installed using `xcodebuild -showsdks`)
-    - `MIN_VERSION` is "9.0"
-08. From the directory of the build script (this will take a while): `bash ./build.sh` (or whatever you named the copied script)
-
-Be sure to note the following:
-
-* You will need to verify and edit the "build.sh" scripts for both OpenSSL and cURL to reflect the correct version numbers and local directory names (lines 11–24).
-* You must add `--enable-ipv6` to line 80 of Adobe's "build.sh" script used for building cURL. This necessary both due to Apple's requirements for IPv6 support and because the library may not build against recent iOS SDKs otherwise.
-* cURL 7.57.0 is known _not_ to work and later versions are unlikely to work either. 7.50.0 is recommended.
+01. Contact project lead and ensure you have repo access to all required submodules, including private ones. Also request a copy of the Adobe RMSDK archive, which is currently not on Github, unzip it and place it in `~/Documents/AdobeRMSDK`.
```
kyles-ep: We should just tell the developer it can be wherever and reference this as `<rmsdk_path>` in subsequent instructions, since the location of this archive is arbitrary.

ettore: Sounds good, will fix.

build-openssl-curl.sh
```diff
@@ -0,0 +1,41 @@
...
+cp build_curl.sh adobe-rmsdk/thirdparty/curl/iOS-libcurl/
+SDKVERSION=`xcodebuild -version -sdk iphoneos | grep SDKVersion | sed 's/SDKVersion[: ]*//'`
+
+echo "======================================="
+echo "Building OpenSSL..."
+cd adobe-rmsdk/thirdparty/openssl
+mkdir public
+mv iOS-openssl/ public/ios/
+cd public/ios
+curl -O https://www.openssl.org/source/old/1.0.1/openssl-1.0.1u.tar.gz
```
kyles-ep: I was reluctant to make these steps into a script for fear that the download URL may change in the future or become outdated.
To make it clear to the developer that this should be configurable in the future, I would extract the target URL for OpenSSL and the OpenSSL version ID into defined variables at the top of the file with a comment saying "edit as required if OpenSSL needs updating or retargeting" and use those variables in the actual downloading and extracting process.

build-carthage.sh
```diff
@@ -0,0 +1,75 @@
...
+# https://github.com/Carthage/Carthage#archive-prebuilt-frameworks-into-one-zip-file .
+# Occasionally Carthage may be still able to build, but it is not certain which
+# build of AudioEngine is actually going to use. To control this better, especially
+# when preparing app Release builds) we are manually managing this dependency by
+# copying the pre-built framework binary into the location Carthage expects it:
+
+cd Carthage
+AE_BUILD_CONFIG=$1
+chmod u+x ../../Certificates/SimplyE/iOS/AudioEngineZipURLExtractor.swift
+AUDIOENGINE_ZIP_URL=$( ../../Certificates/SimplyE/iOS/AudioEngineZipURLExtractor.swift ../../Certificates/SimplyE/iOS/AudioEngine.json )
```
kyles-ep: I'm not sure it's fair to make an assumption about where a developer has their Certificates repo relative to the Simplified-iOS repo

ettore: I made that assumption / indication in the README

build-openssl-curl.sh
ettore: good point, will change

build-carthage.sh
ettore: Let me rephrase: I indicated to clone Simplified-iOS and Certificates as siblings as a requirement in the manual steps, so it seems like a correct assumption here. While I could avoid that completely by parametrizing, the error checking involved and extra work etc didn't seem worth the time to me. At the end of the day placing Certificates next to Simplified-iOS is just as good as other placements AFAIK. Let me know if you disagree or if I can resolve this.

README.md
```diff
@@ -1,49 +1,40 @@
...
-Be sure to note the following:
-
-* You will need to verify and edit the "build.sh" scripts for both OpenSSL and cURL to reflect the correct version numbers and local directory names (lines 11–24).
-* You must add `--enable-ipv6` to line 80 of Adobe's "build.sh" script used for building cURL. This necessary both due to Apple's requirements for IPv6 support and because the library may not build against recent iOS SDKs otherwise.
-* cURL 7.57.0 is known _not_ to work and later versions are unlikely to work either. 7.50.0 is recommended.
+01. Contact project lead and ensure you have repo access to all required submodules, including private ones. Also request a copy of the Adobe RMSDK archive, which is currently not on Github, unzip it and place it in a place of your choice.
+02. Then run:
+```bash
+git clone git@github.com:NYPL-Simplified/Simplified-iOS.git
+git clone git@github.com:NYPL-Simplified/Certificates.git
```
kyles-ep: Just raising a concern over security by obscurity. While it's weak, it's still something.
It might be worth considering just referencing it by something like "the secrets repo"
And perhaps in the future we can make the installation location more arbitrary rather than making the relative pathing set in stone.

ettore: IMO this is ok: while security by obscurity is a practice, a serious hacker would not be stopped by that in any way. They would still need to break github's security, which is really what defines our security. 

Besides, the android repos do something similar in publicly available repos to automate their builds too, e.g.:
https://github.com/NYPL-Simplified/Simplified-Android-SimplyE/blob/e9d9d1b9353e54b428e0cbb6301e54c0c324674c/.travis-pre.sh
https://github.com/NYPL-Simplified/Simplified-Android-Core/blob/b76381f60cfcb47c113dc3554509018897712641/.travis-pre.sh

@MalcolmMcFly thoughts?

MalcolmMcFly: I'm ok either way here. If they can get past GitHub security, we don't stand a chance anyway. I think `Clone Certificate repo` would work too.

PR #1078, opened 2020-02-04
Add full NYPLBookRegistry.h nullability

ettore:
**What's this do?**
Adds nullability annotations to NYPLBookRegistry.h, eliminating a lot of warnings.

Nullability annotations are essential for proper objc-swift communication. This header file already specified if for a couple members, and when you do that Xcode expects the whole file to deal with annotations.

There are no tests added but I verified every single change is safe.
- i looked at the usage in source files to determine what made sense at runtime. E.g. passing a nil bookmark to `addReadiumBookmark:forIdentifier:` would crash the app.
- i checked usages in swift code of every api/property I modified, in particular `nonnull`s.
- When Swift was not involved, I opted for what made more sense semantically. E.g. `cachedThumbnailImageForBook:` is not used by swift, so nullability is irrelevant, but it makes sense for the book parameter to be nonnull (there's no thumbnail for a nil book).
- A few cases were ambiguous: I used existing comments in the code to decide (e.g. location may be nil when adding a book) and/or exercised code in the debugger (e.g. forcing a nil).

No dependencies

N/A

@ettore

### DIFF
--- a/Simplified/NYPLBookRegistry.h
+++ b/Simplified/NYPLBookRegistry.h
@@ -7,22 +7,22 @@
 @class NYPLReadiumBookmark;
 
 // This is broadcast whenever the book registry is modified.
-static NSString *const NYPLBookRegistryDidChangeNotification =
+static NSString *const _Nonnull NYPLBookRegistryDidChangeNotification =
   @"NYPLBookRegistryDidChangeNotification";
 
-static NSString *const NYPLBookProcessingDidChangeNotification =
+static NSString *const _Nonnull NYPLBookProcessingDidChangeNotification =
   @"NYPLBookProcessingDidChangeNotification";
 
 @interface NYPLBookRegistry : NSObject
 
 // Returns all registered books.
-@property (atomic, readonly) NSArray *allBooks;
+@property (atomic, readonly, nonnull) NSArray *allBooks;
 
 // Returns all books that are on hold
-@property (atomic, readonly) NSArray *heldBooks;
+@property (atomic, readonly, nonnull) NSArray *heldBooks;
 
 // Returns all books not on hold (borrowed or kept)
-@property (atomic, readonly) NSArray *myBooks;
+@property (atomic, readonly, nonnull) NSArray *myBooks;
 
 // Returns the number of books currently registered.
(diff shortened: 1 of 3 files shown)


Simplified/NYPLBookRegistry.h
```diff
@@ -7,37 +7,37 @@
...
 // Returns YES if the registry is currently syncing, else NO;
 @property (atomic, readonly) BOOL syncing;
 
-+ (id)new NS_UNAVAILABLE;
-- (id)init NS_UNAVAILABLE;
++ (nonnull id)new NS_UNAVAILABLE;
+- (nonnull id)init NS_UNAVAILABLE;
 
-+ (NYPLBookRegistry *)sharedRegistry;
++ (nonnull NYPLBookRegistry *)sharedRegistry;
```
kyles-ep: If we're going to make this claim we should actually make a hard exit when failing to create this and perform error reporting.

ettore: So from the implementation this method calls `init` and so I followed the same assumptions from the docs, i.e. "In terms of nullability, callers can assume that the NSObject implementation of init does not return nil." -- I suppose there is one difference here where the call to init is performed only once: so if it fails the first time, it will continue to fail. But in that case it would mean that the app is having serious memory issues because it cannot perform a alloc-init close to startup, so not much will work anyways. It seems like a pretty remote edge case to handle, but I could change this to `nullable` to reflect that. Let me know.

The current code is also not handling that case anyway... and honestly I am not sure it should: usually singletons have the assumption that they are not nil. Here it would only be nil for severe system reasons.

In any case, the implementation does perform error logging already.

kyles-ep: I agree it's edge and realistically means it's a memory problem.
I just thought we should explicitly exit and not let the method return and keep the nullable decorator.
I had Bugsnag logging in mind but I suppose if there's memory issues there's not much chance that it'll go through anyways.

ettore: oh you mean call a `@throw NSMallocException;` and keep the `nonnull` decorator? I'd be ok with that.

ettore: Done

PR #1079, opened 2020-02-07
Added Crashlytics

MalcolmMcFly:
What's this do?
Added Crashlytics

Why are we doing this? (w/ JIRA link if applicable)
No Jira task for this feature.

How should this be tested? / Do these changes have associated tests?
Simply running the app and forcing a crash will show in Firebase Console

Dependencies for merging? Releasing to production?

Has the application documentation been updated for these changes?

Did someone actually run this code to verify it works?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,3 +12,4 @@ Simplified/APIKeys.swift
 AudioEngine.json
 Carthage
 bugsnag-dsym-upload.rb
+GoogleService-Info.plist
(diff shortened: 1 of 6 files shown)


ettore:
one more thing: no matter what we choose, we should also update the build scripts and README to handle this extra setup

ettore:
if we just need Crashlytics without the Firebase stuff, these steps should work and are much simpler (skip the Firebase section): https://medium.com/@bgtspen/firebase-crashlytics-carthage-fc4753d827cf

PR #1081, opened 2020-02-13
SIMPLY-1276 Explicit Back button for pages leading to Search

ettore:
**What's this do?**
It changes the Back button title to say "Back" for pages that contain a Search button. It also removes some related logic in NYPLBookDetailViewController that was changing the navigationItem.backBarButton of the presenting VC, breaking encapsulation.

https://jira.nypl.org/browse/SIMPLY-1276
The search button in NYPLCatalogSearchViewController expands in place in the nav bar but only up to the rightmost edge of the Back button. Since this button by default indicates the full previous page, it may lead hardly any space to the user to type and see their search terms. Always using "Back" has been deemed acceptable: while this is not the most exact solution, it is quick / effective implementation-wise. 

For any tab (Catalog, My Books, Reservations):
- tap on search icon: the next page should show a nav bar with "< Back  [empty search field extending throughout the rest of the nav bar]"
- repeat above by tapping on any child lane (e.g. "Young Adult Nonfiction")
- repeat the above by tapping on a book and then on one of the related "More..." links on the right.

No dependencies.

I only added a few comments inline. Not sure where I should update the app documentation.

@ettore


ettore:
> If I understand correctly, this standardizes the back button text to "Back"
> Are the previously used strings still in use elsewhere? If not we should purge them.

Those previously used strings are still used for other things, so I left them in.

PR #1084, opened 2020-02-22
SIMPLY-2491 Fix retain cycles related to URLSession in Settings tab

ettore:
**What's this do?**
Fixes retain cycles caused by URLSession setup and when executing validating credentials.

https://jira.nypl.org/browse/SIMPLY-2491
While this is not a fix for 2491, memory issues related to network request handling could be cause of hard-to-reproduce bugs. E.g. if the completion closure is living longer that it should, it could keep alive any object that it captured, with unknown side effects.
Since we haven't been able to reproduce the bug on our dev environments, releasing this fix would rule out these known peripheral problems in the code.

This is affecting sign-in flow from the Settings tab only. So hammering on the Settings page signing in and out of various accounts / libraries would be good.
NB: Similar cycles still exist in the pop-up sign in panel, but that is not the flow described in SIMPLY-2491. We are thinking to address these issues with a larger refactor of the code in these 2 VCs.

None. 

I will update the JIRA ticket.

@ettore

### DIFF
--- a/Simplified/NYPLSettingsAccountDetailViewController.m
+++ b/Simplified/NYPLSettingsAccountDetailViewController.m
@@ -13,7 +13,6 @@
 #import "NYPLConfiguration.h"
 #import "NYPLLinearView.h"
 #import "NYPLMyBooksDownloadCenter.h"
-
 #import "NYPLSettingsEULAViewController.h"
 #import "NYPLRootTabBarController.h"
 #import "UIFont+NYPLSystemFontOverride.h"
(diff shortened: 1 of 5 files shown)


Simplified/NYPLSettingsAccountDetailViewController.m
```diff
@@ -525,115 +543,131 @@ - (void)validateCredentials
...
   [self.session
    dataTaskWithRequest:request
    completionHandler:^(NSData *data,
                        NSURLResponse *const response,
                        NSError *const error) {
-      NSInteger const statusCode = ((NSHTTPURLResponse *) response).statusCode;
-     
-      if (statusCode == 200) {
-#if defined(FEATURE_DRM_CONNECTOR)
-        NSError *pDocError = nil;
```
ettore: no changes here: I just copied and pasted this code below.

PR #1085, opened 2020-02-25
SIMPLY-2583 Refactor error reporting to Bugsnag wrapper

ettore:
**What's this do?**
Replaces all direct calls to Bugsnag with a wrapper call, making NYPLBugsnagLogs (now renamed to NYPLErrorLogger) the only class that depends on Bugsnag.

https://jira.nypl.org/browse/SIMPLY-2583
This is an interim change to ease the upcoming work for actually replacing Bugsnag. This way it's simpler to see how we are using Bugsnag and all we'll need to do to replace it would be to rewrite the calls in NYPLErrorLogger.
It also future-proofs replacing any other future bug/crash reporting tool we want to may use.

At app startup there's a call to report the app launch. Verifying that that even was recorded before and after this change would verify that Bugsnag is still configured correctly.

None. Releasable to prod if needed.

I updated the in-code docs.

@ettore


ettore:
> Can we rename this wrapper class now instead of later to something generic? This way we can abstract the bug reporting framework layer so that the app is agnostic to the framework backend.

I have already renamed it to `NYPLErrorLogger` but I have that in my current branch for actually removing Bugsnag. I can see if it's easy to cherry pick that commit without too many conflicts

PR #1089, opened 2020-02-26
SIMPLY-2559, SIMPLY-2561 Merge 3.3.5 hotfix branch to develop

ettore:
**What's this do?**
We had already merged to develop the changes regarding the About Us/EULA links some time ago (see PR 1083), so this is just making sure to merge any additional changes that were done on that branch, which just means the app version change.

Avoid losing any changes done on hotfix branches

Already been tested

Already released to prod.

yes.

@ettore


Simplified.xcodeproj/project.pbxproj
```diff
@@ -1938,6 +1938,7 @@
 				IPHONEOS_DEPLOYMENT_TARGET = 9.3;
 				LD_RUNPATH_SEARCH_PATHS = "$(inherited) @executable_path/Frameworks";
 				MARKETING_VERSION = 3.3.4;
+				MARKETING_VERSION = 3.3.5;
```
ettore: this is fixed in my subsequent MR for Crashlytics. Not sure why Xcode did that.

ettore: it's not causing issues as-is either.

PR #1090, opened 2020-02-28
SIMPLY-2583 Bugsnag removal, Crashlytics integration

ettore:
**What's this do?**
Extirpates Bugsnag and replaces it with Crashlytics. This covers crash reporting (with symbolication) as well as non-fatal error/log reporting. Also, since we are starting fresh, I refactored a bit the way non-fatal errors are being reported, reorganizing the error codes and related contextual info.

https://jira.nypl.org/browse/SIMPLY-2583
Crashlytics is the de-facto standard for crash reporting on iOS.

If you know a reproducible crash, do that and after a while (could be 1-15 min) you should see it in the Firebase console. Ask Malcolm to be invited. There's one non-fatal event (app launch) which you should be able to see: https://console.firebase.google.com/project/simplye-nypl/crashlytics/app/ios:org.nypl.labs.SimplyE/issues/44371e0b41226d541f2bbcaacfba8c96
There are also some other manually generated crashes. For most of them you should see stack trace, custom keys, logs and device data, all of them anonymous.

Currently the dSYMs are uploaded only for Release builds, because it might be overkill / time consuming to do that for every build. That's easy to change though from the related Build Phase in Xcode.

I added docs and removed all mentions of Bugsnag.

@ettore


kyles-ep:
Also, did we test each codepath for triggering these errors? Obviously doing it properly in QA is a bit much, but I mean something like forcibly manipulating the codepath to go into the error paths and verifying the full callstack works at each location to get something registered in firebase.

ettore:
> Is there a difference between the verb usage on the reporting functions in NYPLErrorLogger?
> If not we should normalize the verbs. I see "log__" "report__" and "record__".

I think they are all the same. I kinda left it alone to contain the changes, but it's not a big deal to find and replace. I'll use `log` since it's shorter.

ettore:
> Also, did we test each codepath for triggering these errors? Obviously doing it properly in QA is a bit much, but I mean something like forcibly manipulating the codepath to go into the error paths and verifying the full callstack works at each location to get something registered in firebase.

So, I did verify one (the app launch event) and saw that the user info dictionary is being reported nicely, as well as the call stack and custom logs. I don't have reasons to believe that  other event reporting will behave differently: they all use just 2 Crashlytics apis, and one seems like a sub case of the other.  So I could spend some more for those errors that are more involved, but probably I need to time box it to a hour or so. 

Does this sound good @kyles-ep ?

ettore:
I verified that catalog load errors (e.g. 404) and no-network errors are being reported correctly. The logged keys include iOS provided info as well as our own additions, as you can see this in the Firebase Console. Other errors would behave the same, since they share the same code path for adding keys and logs.

PR #1092, opened 2020-03-03
SIMPLY-2590 Replace deprecated method for autorotation

ErnestFan:
**What's this do?**
Replace deprecated method with modern method

Those methods are deprecated in iOS 6.0

Open barcode scanner in account setting, rotate device and make sure the orientation stays portrait

No

No

@ErnestFan ran this in simulator

### DIFF
--- a/Simplified/NYPLBarcodeScanningViewController.m
+++ b/Simplified/NYPLBarcodeScanningViewController.m
@@ -85,18 +85,13 @@
   [self dismissViewControllerAnimated:YES completion:nil];
 }
 
-// TODO: remove after fixing SIMPLY-2590
-#pragma clang diagnostic push
-#pragma clang diagnostic ignored "-Wdeprecated-implementations"
-- (BOOL)shouldAutorotateToInterfaceOrientation:(UIInterfaceOrientation)toInterfaceOrientation {
-  return toInterfaceOrientation == UIInterfaceOrientationPortrait;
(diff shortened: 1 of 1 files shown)


Simplified/NYPLBarcodeScanningViewController.m
```diff
@@ -85,18 +85,13 @@ - (void)didSelectCancel
...
-#pragma clang diagnostic ignored "-Wdeprecated-implementations"
-- (BOOL)shouldAutorotateToInterfaceOrientation:(UIInterfaceOrientation)toInterfaceOrientation {
-  return toInterfaceOrientation == UIInterfaceOrientationPortrait;
+- (UIInterfaceOrientationMask)supportedInterfaceOrientations {
+    return UIInterfaceOrientationMaskPortrait;
 }
 
-- (void)didRotateFromInterfaceOrientation:(UIInterfaceOrientation)fromInterfaceOrientation {
-	[super didRotateFromInterfaceOrientation:fromInterfaceOrientation];
-	[self applyOrientation];
```
ettore: just checking, we lost the call to `applyOrientation` here. Did you verify this is still being called?

ErnestFan: Yeah it's being called in `viewWillTransitionToSize:withTransitionCoordinator:`

PR #1094, opened 2020-03-05
SIMPLY-2489 Prep work for R2

ettore:
**What's this do?**
Various minor fixes and some project reorg while preparing to work on R2. Does NOT introduce any R2 dependencies.

https://jira.nypl.org/browse/SIMPLY-2489
It's difficult to browse source files in our Xcode project, so this groups a few related ones together (ereader, logging, utilities) instead of having them in the ungrouped list. The Italian localization was also badly broken, so I fixed a few major issues while I was at it.

No tests needed except for Italian localization, these are just build / org changes.

This should be merged after https://github.com/NYPL-Simplified/Simplified-iOS/pull/1093

Not needed

@ettore


ettore:
> I take it the strings removed from the EN localization file were not used?

yep, exactly.

PR #1095, opened 2020-03-05
[SIMPLY-2491] Handle error when logging in, report error detail to Crashlytics

ErnestFan:
**What's this do?**
Throw custom NSError when parsing UserProfileDocument, with unit tests
Log error detail to Crashlytics

[SIMPLY-2491](https://jira.nypl.org/browse/SIMPLY-2491?filter=-1)

Manually tested with dummy data since we are unable to reproduce the error

No

No

@ErnestFan

### DIFF
--- a/Simplified/NYPLErrorLogger.swift
+++ b/Simplified/NYPLErrorLogger.swift
@@ -64,6 +64,12 @@ fileprivate let nullString = "null"
 
     // ereader
     case deleteBookmarkFail = 500
+    
+    // Parse failure
+    case parseProfileDataCorrupted = 600
+    case parseProfileTypeMismatch = 601
+    case parseProfileValueNotFound = 602
+    case parseProfileKeyNotFound = 603
   }
 
   // MARK:- Generic helpers
--- a/Simplified/NYPLErrorLogger.swift
+++ b/Simplified/NYPLErrorLogger.swift
@@ -450,12 +450,13 @@ fileprivate let nullString = "null"
     let err = error ?? NSError.init(domain: simplyeDomain,
                                     code: ErrorCode.userProfileDocFail.rawValue,
                                     userInfo: nil)
-    var metadata = [AnyHashable : Any]()
+    var metadata: [AnyHashable : Any] = error?.userInfo ?? [AnyHashable : Any]()
     metadata["errorDescription"] = error?.localizedDescription ?? nullString
     addAccountInfoToMetadata(&metadata)
     reportLogs()
 
     let userInfo = additionalInfo(severity: .error, metadata: metadata)
+    print("Sending record to Crashlytics")
     Crashlytics.sharedInstance().recordError(err,
                                              withAdditionalUserInfo: userInfo)
   }
--- a/Simplified/NYPLErrorLogger.swift
+++ b/Simplified/NYPLErrorLogger.swift
@@ -450,7 +450,7 @@ fileprivate let nullString = "null"
     let err = error ?? NSError.init(domain: simplyeDomain,
                                     code: ErrorCode.userProfileDocFail.rawValue,
                                     userInfo: nil)
-    var metadata: [AnyHashable : Any] = error?.userInfo ?? [AnyHashable : Any]()
+    var metadata = [AnyHashable : Any]()
     metadata["errorDescription"] = error?.localizedDescription ?? nullString
     addAccountInfoToMetadata(&metadata)
     reportLogs()
--- a/Simplified/NYPLErrorLogger.swift
+++ b/Simplified/NYPLErrorLogger.swift
@@ -456,7 +456,6 @@ fileprivate let nullString = "null"
     reportLogs()
 
     let userInfo = additionalInfo(severity: .error, metadata: metadata)
-    print("Sending record to Crashlytics")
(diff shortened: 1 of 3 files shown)


Simplified/NYPLErrorLogger.swift
```diff
@@ -450,12 +450,13 @@ fileprivate let nullString = "null"
...
                                     code: ErrorCode.userProfileDocFail.rawValue,
                                     userInfo: nil)
-    var metadata = [AnyHashable : Any]()
+    var metadata: [AnyHashable : Any] = error?.userInfo ?? [AnyHashable : Any]()
     metadata["errorDescription"] = error?.localizedDescription ?? nullString
     addAccountInfoToMetadata(&metadata)
     reportLogs()
 
     let userInfo = additionalInfo(severity: .error, metadata: metadata)
+    print("Sending record to Crashlytics")
```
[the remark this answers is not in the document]
ErnestFan: fixed

Simplified/NYPLErrorLogger.swift
```diff
@@ -450,7 +450,7 @@ fileprivate let nullString = "null"
     let err = error ?? NSError.init(domain: simplyeDomain,
                                     code: ErrorCode.userProfileDocFail.rawValue,
                                     userInfo: nil)
-    var metadata = [AnyHashable : Any]()
+    var metadata: [AnyHashable : Any] = error?.userInfo ?? [AnyHashable : Any]()
```
ettore: the error.userInfo is already reported by the subsequent call to `recordError` and its values merged with the additional metadata hash, so I think this isn't necessary.

Simplified/NYPLUserProfileDocument.swift
```diff
@@ -1,6 +1,13 @@
 import Foundation
 
 @objcMembers public class UserProfileDocument : NSObject, Codable {
+  static let parseErrorKey: String = "NYPLParseProfileErrorKey"
+  static let parseErrorDescription: String = "NYPLParseProfileErrorDescription"
+  static let parseErrorCodingPath: String = "NYPLParseProfileErrorCodingPath"
+
+  static let decodingErrorCodeInvalid: Int = 4864
```
ettore: are these cocoa error codes? seems to match this:  https://developer.apple.com/documentation/foundation/1448136-nserror_codes/nscoderreadcorrupterror?language=objc

if so we should use their values. 
If not, I wonder if it would make sense to centralize our error codes in NYPLErrorLogger (or other file).

Same for NYPLParseProfileErrorKey below.

Simplified/NYPLUserProfileDocument.swift
```diff
@@ -67,6 +81,32 @@ import Foundation
     dateFormatter.dateFormat = "yyyy-MM-dd'T'HH:mm:ss'Z'"
     jsonDecoder.dateDecodingStrategy = .formatted(dateFormatter)
     
-    return try jsonDecoder.decode(UserProfileDocument.self, from: data)
+    do {
+      return try jsonDecoder.decode(UserProfileDocument.self, from: data)
+    } catch let DecodingError.dataCorrupted(context) {
+      throw NSError(domain: NSCocoaErrorDomain,
```
ettore: shouldn't we use the domain and code of the caught DecodingError instead of changing it ? that way you don't even need to define new constants.

SimplifiedTests/UserProfileDocumentTests.swift
```diff
@@ -62,4 +198,68 @@ class UserProfileDocumentTests: XCTestCase {
...
+    XCTAssertNotNil(data)
+    
+    do {
+      let _ = try UserProfileDocument.fromData(data!)
+      XCTAssert(false)
+    } catch {
+      let err = error as NSError
+      XCTAssertEqual(err.code, UserProfileDocument.decodingErrorCodeNotFound)
+        
+      let customErrorCode = err.userInfo[UserProfileDocument.parseErrorKey] as? Int ?? -1
```
ettore: is the `-1` indicative of a failure? If so I would prefer to have a guard-let instead and XCTFail inside its else so that it's more explicit.

Simplified/NYPLErrorLogger.swift
ErnestFan: fixed

Simplified/NYPLUserProfileDocument.swift
ErnestFan: Its using NSErrorCodes, also moved the parsing error code into NYPLErrorLogger.

SimplifiedTests/UserProfileDocumentTests.swift
ErnestFan: Added guard let instead of -1

Simplified/NYPLUserProfileDocument.swift
ErnestFan: Should I also move those strings into NYPLErrorLogger?

ettore: that's a good question -- it seems like these keys are particular to this class? If so they should stay here (you could even make them private to prevent external usage).  The error codes in NYPLErrorLogger I think could be more generic. For example another class that has a parse error could reuse your same error codes but a different userInfo keys for a different context.

PR #1096, opened 2020-03-06
[Simply -2624] Fix text color on sign in view

ErnestFan:
**What's this do?**
Change text color on Dark Mode when signing in

[SIMPLY-2624](https://jira.nypl.org/browse/SIMPLY-2624?filter=-1)

Open a book without signing in, the signInViewController should pop up and the text color on Barcode and Pin field should be white

No

No

@ErnestFan


ettore:
It would be better to refactor this into a UIColor extension (in swift!) that can do that check in a reusable manner, using a semantic name like `defaultLabelColor` or something like that. For example I noticed that identical check exists in NYPLSettingsAccountDetailViewController. That way updating the color palette would be much simpler if we have to do it across the app. And less error prone.

You could also do this in a follow up PR if you prefer btw.

ErnestFan:
> It would be better to refactor this into a UIColor extension (in swift!) that can do that check in a reusable manner, using a semantic name like `defaultLabelColor` or something like that. For example I noticed that identical check exists in NYPLSettingsAccountDetailViewController. That way updating the color palette would be much simpler if we have to do it across the app. And less error prone.
> 
> You could also do this in a follow up PR if you prefer btw.

Since Malcolm want to release a new build asap, I think it's better to do that on a separate ticket/PR.

ettore:
> > It would be better to refactor this into a UIColor extension (in swift!) that can do that check in a reusable manner, using a semantic name like `defaultLabelColor` or something like that. For example I noticed that identical check exists in NYPLSettingsAccountDetailViewController. That way updating the color palette would be much simpler if we have to do it across the app. And less error prone.
> > You could also do this in a follow up PR if you prefer btw.
> 
> Since Malcolm want to release a new build asap, I think it's better to do that on a separate ticket/PR.

gotcha -- ok, I'll approve this so you can merge but if you can do the refactor soon (or open a ticket if you can't right now) that would be great. Thanks!

PR #1097, opened 2020-03-06
[SIMPLY-2630] Change auth document cache refresh frequency

ErnestFan:
**What's this do?**
Change auth document cache refresh frequency, from 1 day to 1 hour

[SIMPLY-2630](https://jira.nypl.org/browse/SIMPLY-2630?filter=-1)

Didn't test since there is no changes to the validation mechanism
Inspected the variable to make sure the expiry date is one hour different from the modified date

No

No

@ErnestFan


ErnestFan:
So the loadCacheWithData actually also handle other cache, eg. catalogs. I added some parameter with default value to the method, so only auth doc will expiry in an hour now, default will expiry in a day.

Also, the original logic will load from cache if load options contain .preferCache regardless of the cache expired or not. It is now removed and the expiry date should work, tested locally like @ettore suggested

Simplified/AccountsManager.swift
```diff
@@ -19,10 +19,10 @@ let prodUrlHash = prodUrl.absoluteString.md5().base64EncodedStringUrlSafe().trim
  cacheOnly - only fetch from cache, unless `noCache` is specified
  @param completion callback method when this is complete, providing the data or nil if unsuccessful
  */
-func loadDataWithCache(url: URL, cacheUrl: URL, options: AccountsManager.LoadOptions, completion: @escaping (Data?) -> ()) {
+func loadDataWithCache(url: URL, cacheUrl: URL, expiryComponent: Calendar.Component = .day, expiryValue: Int = 1, options: AccountsManager.LoadOptions, completion: @escaping (Data?) -> ()) {
```
ettore: expiryComponent is kind of unclear. Maybe stick with the same terminology apple uses in `Calendar::date(byAdding:value:to:options:)`, so `expiryUnit`.

ErnestFan: Updated the param name to `expiryUnit` and removed `preferringCache` param from `loadAuthenticationDocument`

PR #1098, opened 2020-03-07
[Refactor] UIColor+LabelColor extension

ErnestFan:
**What's this do?**
Refactor the usage of UIColor.label into an extension

See [PR 1096](https://github.com/NYPL-Simplified/Simplified-iOS/pull/1096)

Checked the text color in simulator in both normal and dark mode, nothing changed

No

No

@ErnestFan

### DIFF
--- a/Simplified/UIColor+LabelColor.swift
+++ b/Simplified/UIColor+LabelColor.swift
@@ -1,0 +1,11 @@
+import UIKit
+
+@objc extension UIColor {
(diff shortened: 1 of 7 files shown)


Simplified/UIColor+LabelColor.swift
```diff
@@ -0,0 +1,11 @@
+import UIKit
+
+@objc extension UIColor {
```
ettore: nit: this @objc is not needed

PR #1100, opened 2020-03-11
SIMPLY-2191 Migration - NYPLBookState

ErnestFan:
**What's this do?**
Migrate NYPLBookState from Objective-C to Swift

[SIMPLY-2191](https://jira.nypl.org/browse/SIMPLY-2191?filter=-1&jql=assignee%20%3D%20currentUser()%20AND%20resolution%20%3D%20Unresolved%20ORDER%20BY%20updated%20DESC)

Access a book in different state to make sure it's behaving like intended

No

No

@ErnestFan


Simplified/NYPLBookState.swift
```diff
@@ -0,0 +1,85 @@
+//
+//  .swift
```
kyles-ep: mising filename

Simplified/NYPLBookRegistryRecord.m
```diff
@@ -141,7 +141,7 @@ - (NSDictionary *)dictionaryRepresentation
   
   return @{BookKey: [self.book dictionaryRepresentation],
            LocationKey: NYPLNullFromNil([self.location dictionaryRepresentation]),
-           StateKey: NYPLBookStateToString(self.state),
+           StateKey: @"NYPLBookstatetodo",
```
kyles-ep: This doesn't look like one of the enum's defined keys. What does this do?

Simplified/NYPLBookState.swift
ErnestFan: fixed

Simplified/NYPLBookCell.h
```diff
@@ -2,10 +2,9 @@
 // simply to encapsulate dequeueing cells from a UICollectionView so that the logic does not need to
 // be repeated for every part of the application that displays book cells.
 
-#import "NYPLBookState.h"
-
 @class NYPLBook;
 @class NYPLBookCell;
+typedef NS_ENUM(NSInteger, NYPLBookState);
```
ettore: you should be able to remove this forward declaration: I don't see `NYPLBookState` being used in this header file

Simplified/NYPLBookRegistry.m
```diff
@@ -492,6 +492,12 @@ - (void)setState:(NYPLBookState)state forIdentifier:(NSString *const)identifier
   }
 }
```
kyles-ep: Please comment with TODO so we know that this is *because* of ObjC and can be removed when 100% swift

Simplified/NYPLBookDetailView.h
```diff
@@ -1,10 +1,12 @@
-#import "NYPLBookState.h"
+//#import "NYPLBookState.h"
+//#import "SimplyE-Swift.h"
```
kyles-ep: If imports aren't needed, let's just remove them

Simplified/NYPLBookState.swift
```diff
@@ -0,0 +1,77 @@
...
+        return UsedKey;
+      case .Unsupported:
+        return UnsupportedKey;
+    }
+  }
+}
+
+// For Objective-C, since Obj-C enum is not allowed to have methods
+@objcMembers class NYPLBookStateHelper : NSObject {
+  static func getString(from state: NYPLBookState) -> String {
```
ettore: this name is not very swift-y / objc-y: perhaps `stringValue(from:)` would be better. And I would add a `@objc(stringValueFromBookState:)` above the declaration for a more common objc name.

Simplified/NYPLBookRegistryRecord.m
ErnestFan: Was trying to get the code compile, replaced with the method call

Simplified/NYPLBookRegistry.m
ErnestFan: Added comment

Simplified/NYPLBookState.swift
```diff
@@ -0,0 +1,77 @@
...
+    return state.stringValue()
+  }
+    
+  static func getState(from string: String) -> Int {
```
ettore: same thing here, maybe `bookState(fromString string:)` for swift, and `@objc(bookStateFromString:)` for objc? In general `get` is rarely used in objc or swift.

Simplified/NYPLBookState.swift
```diff
@@ -0,0 +1,77 @@
...
+        self = .DownloadSuccessful
+      case UnregisteredKey:
+        self = .Unregistered
+      case HoldingKey:
+        self = .Holding
+      case UsedKey:
+        self = .Used
+      case UnsupportedKey:
+        self = .Unsupported
+      default:
```
ettore: do we need the default `nil` case if we have Unsupported? If not, the initializer doesn't need to be optional, which would be nice.  But I am just wondering.

Simplified/NYPLBookState.swift
kyles-ep: @ettore is that some sort of decorator that can explicitly set the name used in ObjC?

Simplified/NYPLBookState.swift
kyles-ep: I believe "Unsupported" is an actual book state, not a statement about the enum itself.

Simplified/NYPLBookState.swift
ettore: yep that's it... see for example NYPLBackgroundExecutor::setUpWorkItem(wrapping:). Very useful to keep objc code non-swifty. It's basically the longer form of the usual `@objc func ...`.

I also tend to avoid using @objcMembers because of that, and to avoid exposing too much to objc: I find it pushes me to write more swift rather than going the easy way of adding to existing objc sources.

Simplified/NYPLBookState.swift
ettore: gotcha, never mind then.

Simplified/NYPLBookState.swift
ErnestFan: Updated the method name

Simplified/NYPLBookState.swift
ErnestFan: Updated the method name

ErnestFan:
@ettore @kyles-ep I missed some of the usage NYPLBookState before, the latest commit has covered all the them. Please review it again, sorry about that.

Simplified/NYPLBookRegistry.m
```diff
@@ -802,7 +808,7 @@ - (NSArray *)booksMatchingStateMask:(NSUInteger)mask
      enumerateKeysAndObjectsUsingBlock:^(__attribute__((unused)) NSString *identifier,
                                          NYPLBookRegistryRecord *const record,
                                          __attribute__((unused)) BOOL *stop) {
-       if (record.state & mask) {
+       if (states.count == 0 || [states containsObject:@(record.state)]) {
```
ettore: hmm I don't fully understand the reason for checking `states.count == 0`... if we have no states coming in we still add the book?

2nd thing: from the old code it seems like `record.state` could be nil, we should still continue to check for that to avoid a possible crash (I think... I'm saying that because NSArray::containsObject wants a non-optional in swift, although I know we are in objc here).

ErnestFan: I am using an empty array as a convention of all enum cases, so that we don't need to compare the `record.state` to every enum case

Simplified/NYPLBookState.swift
```diff
@@ -0,0 +1,80 @@
...
+        return UsedKey;
+      case .Unsupported:
+        return UnsupportedKey;
+    }
+  }
+}
+
+// For Objective-C, since Obj-C enum is not allowed to have methods
+// TODO: Remove when migration to Swift completed
+@objc class NYPLBookStateHelper : NSObject {
```
ettore: you can remove `@objc` from the class since this class in inheriting from NSObject

Simplified/NYPLBookRegistry.m
ErnestFan: Added statement to check `record.state`

Simplified/NYPLBookRegistry.m
```diff
@@ -347,7 +347,7 @@ - (void)syncWithCompletionHandler:(void (^)(BOOL success))completion
          }
          for (NSString *identifier in identifiersToRemove) {
            NYPLBookRegistryRecord *record = [self.identifiersToRecords objectForKey:identifier];
-           if (record.state & (NYPLBookStateDownloadSuccessful | NYPLBookStateUsed)) {
+           if (record.state == NYPLBookStateDownloadSuccessful || record.state == NYPLBookStateUsed) {
```
ettore: there's an insidious *potential* bug here: if `record` happens to be nil, `record.state` will evaluate to 0 in objc. This means that depending on which rawValue the enum cases listed here evaluate to, this `if` condition could evaluate to true. This is not the case now, because the 0 case is mapped to Unregistered, but it would be more future-proof to add a nil check on record ANDed to what you have here.

Simplified/NYPLBookRegistry.m
```diff
@@ -392,7 +392,7 @@ - (void)syncWithStandardAlertsOnCompletion
 
 - (void)addBook:(NYPLBook *const)book
        location:(NYPLBookLocation *const)location
-          state:(NYPLBookState)state
+          state:(NSInteger)state
```
ettore: hmm not sure why this changed, but we lost some expressiveness here. If you need an Integer, can't you cast it inside the method instead?

ErnestFan: Having NYPLBookState as one of it's parameter caused this method not being compiled to the bridging header (not able to be called in Swift).
Now there is only called in one Swift file which is a test, but there might be more in the future.

Simplified/NYPLBookState.swift
```diff
@@ -0,0 +1,98 @@
...
+    return NYPLBookState.init(string)?.rawValue ?? -1
+  }
+    
+  @objc(bookStateFromCode:)
+  static func bookState(fromCode code: Int) -> Int {
+    return NYPLBookState.init(rawValue: code)?.rawValue ?? -1
+  }
+    
+  @objc static func allBookStates() -> [NYPLBookState.RawValue] {
+    return [
```
ettore: nit: another way to do this is make NYPLBookState conform to the CaseIterable protocol and here do `return NYPLBookState.allCases.map { $0.rawValue }`

Simplified/NYPLBookDetailButtonsView.m
```diff
@@ -259,7 +259,7 @@ - (void)updateButtons
   
   BOOL fulfillmentIdRequired = NO;
   NYPLBookState state = [[NYPLBookRegistry sharedRegistry] stateForIdentifier:self.book.identifier];
-  BOOL hasRevokeLink = (self.book.revokeURL && state & (NYPLBookStateDownloadSuccessful | NYPLBookStateUsed));
+  BOOL hasRevokeLink = (self.book.revokeURL && state && (state == NYPLBookStateDownloadSuccessful || state == NYPLBookStateUsed));
```
ettore: sorry I missed this before. You don't need the `&& state` because state is an integer

Simplified/NYPLBookRegistryRecord.m
```diff
@@ -79,13 +79,7 @@ - (instancetype)initWithBook:(NYPLBook *const)book
...
   if (!actuallyOnHold) {
     // Set the correct non-holding state.
-    if (!((NYPLBookStateDownloadFailed |
-           NYPLBookStateDownloading |
-           NYPLBookStateDownloadNeeded |
-           NYPLBookStateDownloadSuccessful |
-           NYPLBookStateUsed)
-          & self.state)
-        && self.state != NYPLBookStateUnregistered)
+    if (self.state && (self.state == NYPLBookStateHolding || self.state == NYPLBookStateUnsupported))
```
ettore: `self.state &&` is not needed here because it's essentially an integer, which in objc evaluates to false if it's 0 and true in all other cases.

Simplified/NYPLBookRegistry.h
```diff
@@ -74,10 +74,10 @@ static NSString *const _Nonnull NYPLBookProcessingDidChangeNotification =
 // Adds a book to the book registry until it is manually removed. It allows the application to
 // present information about obtained books when offline. Attempting to add a book already present
 // will overwrite the existing book as if |updateBook:| were called. The location may be nil. The
-// state provided must not be |NYPLBookStateUnregistered|.
+// state provided must be one of NYPLBookState and must not be |NYPLBookStateUnregistered|.
```
kyles-ep: Need to add a TODO to ensure we dont forget to revert this to an enum after swift migration

Simplified/NYPLBookRegistry.m
```diff
@@ -174,8 +174,12 @@ - (void)loadWithoutBroadcastingForAccount:(NSString *)account
     for(NSDictionary *const recordDictionary in dictionary[RecordsKey]) {
       NYPLBookRegistryRecord *const record = [[NYPLBookRegistryRecord alloc]
                                               initWithDictionary:recordDictionary];
+      // If record doesn't exist, proceed to next record
+      if (!record) {
+        break;
```
kyles-ep: Shouldn't this be `continue`?

ErnestFan: Sorry my bad, fixed

PR #1104, opened 2020-03-24
SIMPLY-2655 Secure secrets loading systems

ErnestFan:
**What's this do?**
Add a NYPLSecrets.swift with variables and functions declaration without actual details
Update README with instruction to generate keys into NYPLSecrets.swift

[SIMPLY-2655](https://jira.nypl.org/browse/SIMPLY-2655)

Update your local Certificates repo, follow the fourth step of `Building the application` in README
Please also see if the difference in NYPLSecrets is untracked in git after running the script

No

Updated build instruction

@ErnestFan

### DIFF
--- a/Simplified/NYPLSecrets.swift
+++ b/Simplified/NYPLSecrets.swift
@@ -1,0 +1,25 @@
+import Foundation
+
+enum NYPLSecrets {
+  private static let salt: [UInt8] = [0]
+
+  static var feedbooks:String {
+    let encoded: [UInt8] = [0]
+    return decode(encoded, cipher: salt)
+  }
+
+  static var cardCreator:String {
+    let encoded: [UInt8] = [0]
+    return decode(encoded, cipher: salt)
+  }
+
+  static var cardCreatorInfo:[String:Any] {
+    return [:]
+  }
+
+  static func decode(_ encoded: [UInt8], cipher: [UInt8]) -> String {
+    var decrypted = [UInt8]()
+
+    return String(bytes: decrypted, encoding: .utf8)!
+  }
+}
--- a/Simplified/NYPLSecrets.swift
+++ b/Simplified/NYPLSecrets.swift
@@ -1,24 +1,30 @@
 import Foundation
 
+enum AudioBookVendors: String {
+  case cantook = "cantook"
+}
+
 enum NYPLSecrets {
   private static let salt: [UInt8] = [0]
-
-  static var feedbooks:String {
-    let encoded: [UInt8] = [0]
-    return decode(encoded, cipher: salt)
-  }
-
-  static var feedbooksInfo:[String:Any] {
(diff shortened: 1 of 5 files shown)


Simplified/NYPLSecrets.swift
```diff
@@ -0,0 +1,25 @@
+import Foundation
+
+enum NYPLSecrets {
+  private static let salt: [UInt8] = [0]
+
+  static var feedbooks:String {
+    let encoded: [UInt8] = [0]
+    return decode(encoded, cipher: salt)
+  }
```
kyles-ep: There should be associated info here. Namely the profile url.

ErnestFan: Added

Simplified/NYPLSecrets.swift
```diff
@@ -0,0 +1,29 @@
+import Foundation
```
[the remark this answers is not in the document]
ettore: I just tried to add a random file already under git to `.gitignore`, but once I modify that file git will show me that that file has changes. So that's a problem. I think once it's under source control, .gitignore doesn't matter anymore.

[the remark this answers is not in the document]
ErnestFan: Try `git update-index --skip-worktree Simplified/NYPLSecrets.swift`
The line above should ignore any changes but still manage to keep the file in source control
Let me know if it works or not

Simplified/NYPLSecrets.swift
```diff
@@ -0,0 +1,35 @@
...
+
+  static func feedbookInfo(forVendor name: AudioBookVendors) -> [String:Any] {
+    let info : [String: [String:Any]] = [:]
+    return info[name.rawValue] ?? [:]
+  }
+
+  static func decode(_ encoded: [UInt8], cipher: [UInt8]) -> String {
+    var decrypted = [UInt8]()
+
+    return String(bytes: decrypted, encoding: .utf8)!
```
ettore: as I was saying in the other PR, returning an optional string would be my preference. I could live with returning "" if decoding fail, but I am not sure it makes sense.

Simplified/NYPLSecrets.swift
```diff
@@ -27,9 +27,8 @@ enum NYPLSecrets {
     return info[name.rawValue] ?? [:]
   }
 
-  static func decode(_ encoded: [UInt8], cipher: [UInt8]) -> String {
+  static func decode(_ encoded: [UInt8], cipher: [UInt8]) -> String? {
     var decrypted = [UInt8]()
-
-    return String(bytes: decrypted, encoding: .utf8)!
+    return String(bytes: decrypted, encoding: .utf8) ?? nil
```
ettore: no need for `?? nil`, all good otherwise!

PR #1107, opened 2020-03-31
SIMPLY-2473 Refactor reader settings view for R2 usage

ettore:
**What's this do?**
It refactors our reader settings view so that it can be used with the R2 reader. It also makes it so that the settings view opens as a drop-down menu from the top navigation bar on all platforms, and synchronizes font size changes between R1 and R2.

https://jira.nypl.org/browse/SIMPLY-2473

- Open a book from the SimplyE collection that you previously downloaded with R1 (using 3.3.7).
- The book should open with the previously set font size.
- Tapping on the "TT" user settings button on the top right should display our usual UI instead of the one from the R2 test app. 
- Brightness and font size adjustments should work and persist once you close the reader or the app.
- Font typeface and color scheme do not currently work (this is tracked in separate tickets).

This will merge into our Readium2 feature branch.

Only inside the app.

@ettore

### DIFF
--- a/Simplified/NYPLReaderSettings.h
+++ b/Simplified/NYPLReaderSettings.h
@@ -29,43 +29,36 @@ typedef NS_ENUM(NSInteger, NYPLReaderSettingsMediaOverlaysEnableClick) {
   NYPLReaderSettingsMediaOverlaysEnableClickFalse = 1
 };
 
-static NSString *const NYPLReaderSettingsColorSchemeDidChangeNotification =
-  @"NYPLReaderSettingsColorSchemeDidChange";
-
-static NSString *const NYPLReaderSettingsFontFaceDidChangeNotification =
-  @"NYPLReaderSettingsFontFaceDidChange";
-
-static NSString *const NYPLReaderSettingsFontSizeDidChangeNotification =
-  @"NYPLReaderSettingsFontSizeDidChange";
-
-static NSString *const NYPLReaderSettingsMediaClickOverlayAlwaysEnableDidChangeNotification =
-@"NYPLReaderSettingsMediaClickOverlayAlwaysEnableDidChangeNotification";
+extern NSString * _Nonnull const NYPLReaderSettingsColorSchemeDidChangeNotification;
+extern NSString * _Nonnull const NYPLReaderSettingsFontFaceDidChangeNotification;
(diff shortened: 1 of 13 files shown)


Simplified/NYPLReaderSettings.h
```diff
@@ -21,51 +21,45 @@ typedef NS_ENUM(NSInteger, NYPLReaderSettingsFontSize) {
   NYPLReaderSettingsFontSizeLarge = 4,
   NYPLReaderSettingsFontSizeXLarge = 5,
   NYPLReaderSettingsFontSizeXXLarge = 6,
-  NYPLReaderSettingsFontSizeXXXLarge = 7
+  NYPLReaderSettingsFontSizeXXXLarge = 7,
+  NYPLReaderSettingsFontSizeLargest = 7
```
kyles-ep, line 25: Is this value duplication intentional?

Simplified/NYPLReaderSettings.h
```diff
@@ -21,51 +21,45 @@ typedef NS_ENUM(NSInteger, NYPLReaderSettingsFontSize) {
...
-
-static NSString *const NYPLReaderSettingsFontSizeDidChangeNotification =
-  @"NYPLReaderSettingsFontSizeDidChange";
-
-static NSString *const NYPLReaderSettingsMediaClickOverlayAlwaysEnableDidChangeNotification =
-@"NYPLReaderSettingsMediaClickOverlayAlwaysEnableDidChangeNotification";
+extern NSString * _Nonnull const NYPLReaderSettingsColorSchemeDidChangeNotification;
+extern NSString * _Nonnull const NYPLReaderSettingsFontFaceDidChangeNotification;
+extern NSString * _Nonnull const NYPLReaderSettingsFontSizeDidChangeNotification;
+extern NSString * _Nonnull const  NYPLReaderSettingsMediaClickOverlayAlwaysEnableDidChangeNotification;
```
kyles-ep, line 36: spacing

Simplified/NYPLReaderSettings.h
ettore: yeah it is. I use it to make it clear that this is meant to be the "largest" value here and so if we were to add a NYPLReaderSettingsFontSizeXXXXLarge (not that we would hopefully!), "largest" will need to be updated but it will still be semantically correct wherever it's used.   

On the other side (i.e. for "small" sizes) we have "smallest" which already expresses the above.

PR #1109, opened 2020-04-01
[SIMPLY-2677] Fixing incomplete implementation from SIMPLY-2339

kyles-ep:
**What's this do?**
Fixes up a missing alert when attempting to report a problem doc.

https://jira.nypl.org/browse/SIMPLY-2677

Somehow trigger a problem document for a book and then attempt to view and report the problem document.

n/a

n/a

@kyles-ep  for the alert but not for the email sending dialog because I don't have a physical device.

### DIFF
--- a/Simplified/NYPLBookDetailsProblemDocumentViewController.swift
+++ b/Simplified/NYPLBookDetailsProblemDocumentViewController.swift
@@ -96,6 +96,7 @@
     self.submitButton = submitButton
     submitButton.translatesAutoresizingMaskIntoConstraints = false
     submitButton.setTitle("Send to Support", for: .normal)
+    submitButton.isEnabled = AccountsManager.shared.currentAccount?.supportEmail != nil
     submitButton.addTarget(self, action: #selector(submitButtonWasPressed), for: .touchDown)
     
     scrollView.addSubview(label)
--- a/Simplified/NYPLBookDetailsProblemDocumentViewController.swift
+++ b/Simplified/NYPLBookDetailsProblemDocumentViewController.swift
@@ -196,6 +196,6 @@
         body: body
       )
     }))
-    self.present(alert, animated: false, completion: nil)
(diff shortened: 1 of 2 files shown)


Simplified/NYPLBookDetailsProblemDocumentViewController.swift
```diff
@@ -171,6 +172,11 @@
   }
   
   func submitButtonWasPressed() {
+    guard let supportEmail = AccountsManager.shared.currentAccount?.supportEmail else {
```
ettore, line 175: did we get confirmation that we should not use a default email anymore -- that gethelp@nypl.org for example, or another one Risa can provide maybe? For support purposes it would be good not to silently fail here.

Simplified/NYPLBookDetailsProblemDocumentViewController.swift
```diff
@@ -185,10 +191,11 @@
         BookIdentifier:\n\(self.book?.identifier ?? "n/a")\n\n
       """
       ProblemReportEmail.sharedInstance.beginComposing(
-        to: AccountsManager.shared.currentAccount?.supportEmail ?? "gethelp@nypl.org",
+        to: supportEmail,
         presentingViewController: self,
         body: body
       )
     }))
+    self.present(alert, animated: false, completion: nil)
```
ettore, line 199: why no animation?

kyles-ep: No particular reason. I'll put it in.

Simplified/NYPLBookDetailsProblemDocumentViewController.swift
kyles-ep: This is supposed to be a highly exceptional case and Risa did not want to send non-NYPL users to that email address. We the submit button is supposed to be disabled if there is no support email.

PR #1112, opened 2020-04-03
SIMPLY-2671 Prevent crash when signing up without card creator

ettore:
**What's this do?**
Prevents a crash currently reported by Crashlytics, caused by failure in creating a registration URL object. It also tightens the logic that determines when card creator can be used or not: we now require both the card creator URL prefix and a valid sign up URL after the prefix. If there's no prefix, we still validate the URL upstream in the Account initialization instead than in the VCs.

https://jira.nypl.org/browse/SIMPLY-2671

This crash happens during sign up when the account does NOT support the card creator. The condition for crash is when the `rel` field of the OPDS2 link in the authentication document is a non-nil string that does not represent a valid URL.
Since this code changes some of the sign up logic, I would recommend QA'ing all the sign up flows (with/without card creator, using different libraries that support or not sign up, etc).

no dependencies

will update the ticket with QA info.

@ettore


Simplified/Account.swift
```diff
@@ -149,7 +149,7 @@ private let accountSyncEnabledKey        = "NYPLAccountSyncEnabledKey"
     mainColor = authenticationDocument.colorScheme
     
     let registerUrl = authenticationDocument.links?.first(where: { $0.rel == "register" })?.href
-    if let url = registerUrl, url.hasPrefix("nypl.card-creator:") == true {
+    if let url = registerUrl, url.trimmingCharacters(in: .whitespacesAndNewlines).lowercased().hasPrefix("nypl.card-creator:") {
       supportsCardCreator = true
       cardCreatorUrl = String(url.dropFirst("nypl.card-creator:".count))
     } else {
```
kyles-ep, line 155: We should be able to save a lot of peripheral logic if we set `cardCreatorUrl` to `nil` on the condition that it does not form a proper URL. This would prevent the need for a URL check later down the line and cause `registrationIsPossible` to be false.

ettore: Do you mean that if `registrationIsPossible` is false, the registration cell won't be there at all?

ettore: so it seems like if we set `cardCreatorUrl` to nil, the code I changed in the VC won't be triggered. Which is good in a way (no crash). 

However is this ok functionality/product wise? The code looks like if the registerURL is not a card creator URL BUT is a valid URL, the app would still be able to register the user. It seems like the VC code was written with this use case in mind, so this would be a functionality change. Let me know if I'm wrong though, I don't know this code as well as you.

ettore: In the way it's used now essentially, the `Account::cardCreatorUrl` really means `Account::signUpUrl`. It can be a "card creator URL" or a normal URL.

ettore: Unit test `testAccountSetAuthenticationDocument` also suggests that the _"card creator not supported, but cardCreatorUrl not nil and valid"_ scenario is possible. Brooklyn library I think falls under this case, since it provides `https://go.bklynlibrary.org/ecard` as cardCreatorURL, which is a valid url, but that would set `supportsCardCreator = false`. So I pushed a new commit to make this explicit.

PR #1113, opened 2020-04-03
Feature/dpla audiobooks

kyles-ep:
**What's this do?**
Updates NYPLAudiobookToolkit and NYPLAEToolkit versions for DPLA audiobook support. Also adds support for mp4/m4a formats.

https://jira.nypl.org/browse/SIMPLY-2302

RBDigital, Findaway, DPLA, and "none-of-the-above" audiobooks should be tessted for downloading and playback.

n/a

I don't think so. Where is it?

@kyles-ep


Cartfile.resolved
```diff
@@ -1,8 +1,8 @@
 binary "Crashlytics.json" "3.14.0"
 binary "https://dl.google.com/dl/firebase/ios/carthage/FirebaseAnalyticsBinary.json" "6.18.0"
 github "NYPL-Simplified/CardCreator-iOS" "v1.1.2"
-github "NYPL-Simplified/NYPLAEToolkit" "27c08e76cb46985f7861cce50782972b8e05e6fa"
-github "NYPL-Simplified/NYPLAudiobookToolkit" "010c2b066c51672fc0bbb4abac485b0cdde1025e"
+github "NYPL-Simplified/NYPLAEToolkit" "74925bb79dc256cfb7d96be22039667039aa70dd"
```
ettore, line 4: do you have more work planned for these 2 frameworks? If not, can you add release tags for these commits?

PR #1114, opened 2020-04-08
SIMPLY-2691 More Crashlytics logging and fix a couple minor regressions

ettore:
**What's this do?**
This adds user barcode (one-way hashed) tracking to Crashlytics to more easily diagnose hard-to-reproduce errors happening in the wild. Also adds a lot more logging related to account management. Fixes a couple differences accidentally introduced in 3.3.7 related to BookState handling.

https://jira.nypl.org/browse/SIMPLY-2691

It's unlikely this change actually fixes the issue, plus we never got a clear set of reproducible steps. To test the additional logging and user tracking:
- sign in to a library with your barcode, browse around, sign into different libraries etc.
- ask me to compute the hash for your barcode(s)
- Load the Firebase console, click on Crashlytics, filter by 3.4.0, make sure you're not only looking at crashes, there's a lot of non-fatal events logged: https://console.firebase.google.com/project/simplye-nypl/crashlytics/app/ios:org.nypl.labs.SimplyE/issues?state=open&time=last-thirty-days&type=all
- click on Search by user ID, and enter the hash. You should get some results.
- Each event now should show some info, the most useful stuff is under "Keys". This should not be empty, depending on the event code.

no dependencies

Added inline.

@ettore

### DIFF
--- a/Simplified/NYPLAccount.h
+++ b/Simplified/NYPLAccount.h
@@ -60,8 +60,6 @@ static NSString *const NYPLAccountLoginDidChangeNotification =
 
 - (void)removeAll;
 
-- (void)removeObject:(NSString *const)key;
-
 - (NSDictionary *)licensor;
 
 @end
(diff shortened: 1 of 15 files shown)


Simplified/NYPLAccount.h
```diff
@@ -54,14 +54,8 @@ static NSString *const NYPLAccountLoginDidChangeNotification =
 
 - (BOOL)hasBarcodeAndPIN;
 
-- (BOOL)hasCredentials;
```
kyles-ep, line 57: Are these being removed because they are unused? If so, can we put them back? iirc this may break OpenEBooks merge.

ettore: Yeah, they are unused. Are they still used by OpenEBooks or is it just to avoid merge conflicts?

Also, I was wondering about that merge: is the current open PR still ok to review? I was wondering what's the plan there.

killerpuppytails: Please do review that open PR!

ettore: I readied the methods still used by OpenEBooks.

PR #1115, opened 2020-04-09
Fixing Audiobooks

kyles-ep:
**What's this do?**
- Updates to use new NYPLAudiobookToolkit version
- Refactors logging to be continuous so that crashes can record the logging timeline, as opposed to just events

- We need to verify logging words for both events and crashes. I verified logging for crashes in simulator but can't find the app launch event.
- We need to verify DPLA downloading is fixed in a real device.


Simplified/Log.swift
```diff
@@ -44,25 +45,10 @@ final class Log: NSObject {
...
-    var overwrite = false
-    if let size = try? FileManager.default.attributesOfItem(atPath: logUrl.path)[FileAttributeKey.size] as! Int {
-      if size > 1048576 {
-        overwrite = true
-      }
-    }
-    if overwrite {
-      try? formattedMsg.write(to: logUrl, atomically: false, encoding: .utf8)
+    if level != .debug {
+      CLSNSLogv("%@", getVaList([formattedMsg]))
```
ettore, line 49: According to the docs in CLSLogging.h  it seems like we should actually not use this in production. I had not seen this before. Can you add a `#if DEBUG` to use that only when debugging and use `CLSLogv` otherwise?

PR #1119, opened 2020-04-13
SIMPLY-2675 Fix retain cycle in NYPLRemoteViewController

ettore:
**What's this do?**
Fixes a retain cycle afflicting NYPLRemoteViewController and its subclass NYPLCatalogFeedViewController. This could lead to memory / performance issues down the line. The crash report itself does not point to this, but it does come from a async callback from code paths originating in either NYPLCatalogFeedViewController or NYPLCatalogSearchViewController.

https://jira.nypl.org/browse/SIMPLY-2675

Unfortunately I was not able to reproduce this crash. I suggest exploratory testing around catalog and search screens, trying to move fast between one another (e.g. search and immediately go back to catalog, then back to search).

no dependencies

Added better code docs inline.

@ettore


Simplified/en.lproj/Localizable.strings
```diff
@@ -74,6 +75,7 @@
 "SyncFailed" = "Sync Failed";
 "TimedOut" = "The server did not respond. Please check your connection or try again later.";
 "TryAgain" = "Try Again";
+"Unknown error parsing problem document" = "Unknown error parsing problem document";
```
kyles-ep, line 78: It seems like the other string keys aren't space-separated. Maybe we should follow suit?

Simplified/NYPLRemoteViewController.h
```diff
@@ -22,11 +21,16 @@
...
-/// is only called if the data was downloaded successfully. The handler may return nil to indicate
-/// that there is something wrong with the data.
+/**
+ This is the designated initializer.
+ @param URL The URL where to fetch the data at.
+ @param handler Must not be nil. It is strongly retained. The @p data parameter
+ will never be @p nil as this handler is only called if the download was
+ successful. This may return nil to indicate that there's something wrong
+ with the downloaded data.
+ */
```
kyles-ep, line 31: Is `@p` specific to some markup system?

Simplified/en.lproj/Localizable.strings
ettore: we are sort of inconsistent there, for some strings we do use the space-separated English version as the key. I sort of prefer that approach because for incomplete localizations like our case, if a key is missing in a given localization you would at least see see the English version. Otherwise you see something like "UnknownErrorParsingProblemDocument" but in some cases it can be something really cryptic like "SettingsAccountViewControllerLogoutMessageDefault". I've seen this happening often for Italian for example.

Simplified/NYPLRemoteViewController.h
ettore: yep, it's the objc header-doc formatting :)

ettore: it just highlights the following word.

PR #1121, opened 2020-04-15
Refreshed the guide for building a project without Adobe DRM

Neckk:
**What's this do?**
Updates guide for building project without Adombe DRM and fixes build errors

To help users build the project.

I made some changes to build settings and one of the swift files. In order to test them, compile the project with Adobe DRM, and if it succeeds compile it without Adobe. If any errors were to show, they'd do it during build.

none

Yes

No


README.md
```diff
@@ -43,6 +43,7 @@ Both scripts must be run from the Simplified-iOS repo root.
 
 # Building Without Adobe DRM
 
+**Note:** In order to build dependencies, you need Xcode 11.4 or higher.
```
ettore, line 46: can you mention the Xcode 11.4 requirement above where System Requirements is? This applies to both DRM and non-DRM builds.

README.md
```diff
@@ -51,16 +52,19 @@ Both scripts must be run from the Simplified-iOS repo root.
...
 08. `carthage bootstrap --platform ios --use-ssh`
 09. `cp APIKeys.swift.example Simplified/APIKeys.swift` and edit accordingly.
 10. `cp Accounts.json.example Simplified/Accounts.json`.
-11. `(cd readium-sdk; sh MakeHeaders.sh Apple)` (parentheses included) to generate the headers for Readium.
-12. `open Simplified.xcodeproj`
-13. Remove import of "Simplified+RMSDK.xcconfig" from "Simplified.xcconfig".
-14. Delete `NYPLAEToolkit.framework` and `AudioEngine.framework` from "Link Binary with Libraries", and remove input and output filepaths for `AudioEngine.framework` and `NYPLAEToolkit.framework` from `Copy Frameworks (Carthage)`.
-15. Note: For now, we recommend keeping any unstaged changes as a single git stash until better dynamic build support is added.
-16. Build.
+11. `cp GoogleService-Info.plist.example GoogleService-Info.plist` and edit with you firebase project config.
```
ettore, line 59: if you don't edit anything for this file, does the project build and run anyway?

Simplified/NYPLSettingsAccountDetailViewController.m
```diff
@@ -466,15 +466,15 @@ - (void)logOut
...
+    [self presentViewController:[NYPLAlertUtils
                                  alertWithTitle:@"SettingsAccountViewControllerCannotLogOutTitle"
                                  message:@"SettingsAccountViewControllerCannotLogOutMessage"]
                        animated:YES
                      completion:nil];
   } else {
-    [[NYPLMyBooksDownloadCenter sharedDownloadCenter] reset:self.selectedAccountType];
-    [[NYPLBookRegistry sharedRegistry] reset:self.selectedAccountType];
-    [[NYPLAccount sharedAccount:self.selectedAccountType] removeAll];
+    [[NYPLMyBooksDownloadCenter sharedDownloadCenter] reset:self.selectedNYPLAccount.userID];
```
ettore, line 475: what are these changes for? they seem unrelated to this PR

Simplified.xcodeproj/project.pbxproj
```diff
@@ -2089,7 +2089,7 @@
 				PRODUCT_NAME = "$(TARGET_NAME)";
 				PROVISIONING_PROFILE = "2e185b6c-271e-4b02-a05e-860b8c3831f6";
 				PROVISIONING_PROFILE_SPECIFIER = "SimplyE Development (Keychain Migration)";
-				SWIFT_ACTIVE_COMPILATION_CONDITIONS = "DEBUG FEATURE_DRM_CONNECTOR";
+				SWIFT_ACTIVE_COMPILATION_CONDITIONS = DEBUG;
```
ettore, line 2092: can you remove these changes and mention them in the list of manual steps instead? Otherwise the DRM build will be affected and build instructions will need to change there. I'd rather not add them there because we have automated the steps already for DRM build.

Simplified.xcodeproj/project.pbxproj
```diff
@@ -2139,7 +2139,7 @@
 				PRODUCT_NAME = "$(TARGET_NAME)";
 				PROVISIONING_PROFILE = "b3d9154d-70e1-48d6-a0c5-869431277a5c";
 				PROVISIONING_PROFILE_SPECIFIER = "SimplyE Distribution 2019";
-				SWIFT_ACTIVE_COMPILATION_CONDITIONS = FEATURE_DRM_CONNECTOR;
+				SWIFT_ACTIVE_COMPILATION_CONDITIONS = "";
```
ettore, line 2142: same here

Simplified/NYPLSettingsAccountDetailViewController.m
```diff
@@ -578,6 +578,7 @@ - (void)validateCredentials
   [task resume];
 }
 
+#if defined(FEATURE_DRM_CONNECTOR)
```
ettore, line 581: this also seems like it should stay?

README.md
Neckk: yep, it builds and runs

Simplified/NYPLSettingsAccountDetailViewController.m
Neckk: Without this changes, project didn't build.
This hunk of code lies in else part of #if FEATURE_DRM_CONNECTOR check (which means it wasn't compiled during development with Adobe).
Variable self.selectedAccountType no longer exists.I noticed that a piece of code inside FEATURE_DRM_CONNECTOR part uses self.selectedNYPLAccount.userID as a parameter in those methods.

Simplified/NYPLSettingsAccountDetailViewController.m
Neckk: Function in this block is used only for Adobe builds and doesn't compile otherwise. From history of the file I see, that this method was refactored from other larger method, and it was embedded in #if check back then.

Simplified/NYPLSettingsAccountDetailViewController.m
ettore: oh I see, gotcha. BTW selectedNYPLAccount has been renamed recently, make sure to rebase on top of develop. 

But I also wonder if you should use `self.selectedAccountId` which refers to the library currently selected. Judging from the `reset:` method in NYPLMyBooksDownloadCenter that's probably what you want. 

Some background here: `account` as terminology is a bit confusing. There's an Account class that essentially identifies a library. Then there's a `NYPLAccount` class that means "User account". @ErnestFan is rewriting that in swift and renaming it to NYPLUserAccount for more clarity.

Simplified/NYPLSettingsAccountDetailViewController.m
ettore: I see, makes sense.

Simplified/NYPLSettingsAccountDetailViewController.m
Neckk: @ettore of course you were right, I changed `self.selectedNYPLAccount.userID` to `self.selectedAccountId`

ettore:
Looks good!

PR #1123, opened 2020-04-23
SIMPLY-2191 NYPLAccount migration

ErnestFan:
**What's this do?**
Migrate NYPLAccount to Swift
Rename NYPLAccount to NYPLUserAccount
Add NSRecursiveLock to NYPLUserAccount

[SIMPLY-2191](https://jira.nypl.org/browse/SIMPLY-2191)

Log in to an account and make sure the account info is stored into keychain

No

No

@ErnestFan


Simplified/AccountsManager.swift
```diff
@@ -26,7 +26,7 @@ private let prodUrlHash = prodUrl.absoluteString.md5().base64EncodedStringUrlSaf
     static let strict_offline: LoadOptions = [.preferCache, .cacheOnly]
   }
 
-  static let NYPLAccountUUIDs = [
+  static let NYPLUserAccountUUIDs = [
```
kyles-ep, line 29: These are account IDs

Simplified/NYPLWelcomeScreen.swift
```diff
@@ -284,7 +284,7 @@ import PureLayout
 final class NYPLWelcomeScreenAccountList: UIViewController, UITableViewDelegate, UITableViewDataSource {
   
   var accounts: [Account]!
-  var nyplAccounts: [Account]!
+  var nyplUserAccounts: [Account]!
```
kyles-ep, line 287: Also not NYPLUserAccount objects in this array

Simplified/AccountsManager.swift
ErnestFan: Thanks for the catch, fixed

Simplified/NYPLWelcomeScreen.swift
ErnestFan: fixed

Simplified/NYPLUserAccount.swift
```diff
@@ -0,0 +1,402 @@
...
+    sharedKeychain.setObject(token, forKey: adobeTokenKey)
+    sharedKeychain.setObject(patron, forKey: patronKey)
+    
+    NotificationCenter.default.post(
+      name: Notification.Name.NYPLUserAccountDidChange,
+      object: self
+    )
+  }
+  
+  @objc(setAdobeVender:)
```
ettore, line 199: spelling: it's `vendor` not `vender`

Simplified/NYPLUserAccount.swift
```diff
@@ -0,0 +1,402 @@
...
+  }
+  
+  func removeAll() {
+    guard let sharedKeychain = NYPLKeychain.shared() else {
+      return
+    }
+    
+    NYPLUserAccount.accountInfoLock.lock()
+    defer {
+        NYPLUserAccount.accountInfoLock.unlock()
```
ettore, line 389: where you able to test this code at runtime? this `removeAll` function will unlock the lock twice. Not sure if that's undefined behavior. You can also enable the undefined behavior sanitizer in the scheme run diagnostics settings to verify.

Simplified/NYPLUserAccount.swift
```diff
@@ -0,0 +1,402 @@
...
+}
+
+@objc extension NSNotification {
+  public static let NYPLUserAccountDidChange = Notification.Name.NYPLUserAccountDidChange
+  public static let NYPLUserAccountLoginDidChange = Notification.Name.NYPLUserAccountLoginDidChange
+}
+
+@objcMembers class NYPLUserAccount : NSObject {
+  static private let shared = NYPLUserAccount()
+  static private let accountInfoLock = NSRecursiveLock()
```
ettore, line 15: did you make this static to protect access to keychain because `NYPLKeychain` is not thread safe? routing all NYPLUserAccount instances through this shared lock makes me think of potential deadlocks. It would be safer to have it as a normal instance var unless I'm missing something. 

If `NYPLKeychain` is not thread-safe it'd be better to have that class handle its own threadsafety internally instead of fixing it here. This is of course out of scope for this work.

Simplified/NYPLUserAccount.swift
```diff
@@ -0,0 +1,402 @@
...
+    )
+  }
+  
+  @objc(setLicensor:)
+  func setLicensor(licensor: [String : Any]) {
+    guard let sharedKeychain = NYPLKeychain.shared() else {
+      return
+    }
+    
+    NYPLUserAccount.accountInfoLock.lock()
```
ettore, line 243: not sure you need to lock here... all this method is actually doing is issuing a call to NYPLKeychain. Ideally that class should handle its own thread-safety.

Same thing for `setAuthorizationIdentifier` below.

PR #1124, opened 2020-04-28
SIMPLY-2656 Remove unused reader settings code

ettore:
**What's this do?**
Removes some unused code that was used for user settings.
Also removes a duplicate reference to NYPLAttributedString.m (reported by Xcode)

https://jira.nypl.org/browse/SIMPLY-2656

This code was compiled in but not connected to any live code paths.

Merging on R2 feature branch.

no need

@ettore


ettore:
> I don't know enough to comment here, but I'll approve if the code has been demonstrated to build and run.

yep, it does... This is a removal I forgot to do in my previous PR for user settings.

PR #1125, opened 2020-04-28
SIMPLY-2607 Readium2 TOC functionality

ettore:
**What's this do?**
Implements TOC functionality using R2. It only partially refactors the R1 logic because to fully do that we would need to finish the work for Bookmarks, which is slated for the near future. Then we should be able to just have one storyboard and one VC.

https://jira.nypl.org/browse/SIMPLY-2607

Open a book and access table of contents. Functionally this should be identical to what we currently have using R1.

Merging to feature branch.

In-code comments.

@ettore


vladimirfedorov:
Probably NYPLReaderPositions.storyboard should be added to "SimplyE" target as well, as that target crashes otherwise

ettore:
> Probably NYPLReaderPositions.storyboard should be added to "SimplyE" target as well, as that target crashes otherwise

oh good catch, I totally forgot. Will fix and verify if anything else is missing from that target.

PR #1129, opened 2020-05-07
Separate some business logic from SignIn VCs

ettore:
**What's this do?**
Refactors some business logic out of the Sign In VCs. Also rewrites that logic in Swift.

To improve separation of concerns, code readability and reusability. This is also some prep work for working on juvenile card creator.
Note that there's much more logic that could be moved out, but I decided to stop because (1) it's not code I need to work on for card creator, and (2) the card creator project is time sensitive. The idea is that we can move code to the NYPLSignInBusinesLogic object whenever we need to change the logic currently in the VC.

All sign-in and sign-up flows should be tested, from both the Settings tab and the modal Sign In page that pops up when trying to get a book as a signed-out user.

Can be merged into develop.

This PR adds a lot more documentation for how various functions of the sign-in/up flows work together.

@ettore


Simplified/NYPLSettingsAccountDetailViewController.m
```diff
@@ -45,66 +45,81 @@ typedef NS_ENUM(NSInteger, CellKind) {
...
+#pragma mark - NSObject
 
-- (instancetype)initWithAccount:(NSString *)account
+// overriding superclass's designated initializer
+- (instancetype)initWithStyle:(__unused UITableViewStyle)style
 {
-  self.selectedAccountId = account;
-  self.selectedAccount = [[AccountsManager sharedInstance] account:self.selectedAccountId];
-  return [self init];
+  return [self initWithAccount:nil];
```
ErnestFan, line 113: Is this intended? Cause the `nil` value will be passed into `[[NYPLSignInBusinessLogic alloc] initWithLibraryAccountID:libraryUUID]` and it's not optional on the Swift side

ettore: that's a good point, I'll take a look at what I can do. Maybe I'll have to disable that initializer, although I'm supposed to override it per Cocoa conventions.

ettore: I pushed a change where I ended up continuing the initialization as it was done in other parts of the code (see NYPLSettingsAccountsList.swift). This should be fine because there's no code path actually invoking `initWithStyle:` for this VC. However, I also added an assert to catch this situation if we incur in it in the future.
Thanks for catching this.

PR #1131, opened 2020-05-21
Use NSLocalizedDescriptionKey for alert messages

ettore:
**What's this do?**
Use NSLocalizedDescriptionKey for alert messages instead of custom/undocumented "message" userInfo key. Also added localization for some related missing unknown error keys.

Also some more sign-in logic cleanup and using the obfuscated secrets instead of Certificates/APIKeys.

The alert messaging logic could present some unlocalized / not-so-userfriendly error messages in some code paths, so I cleaned that. Also, since we have obfuscation logic let's use for other existing keys/secrets. 

- Slowing down the internet connection to a crawl should cause the app to error out much more, which will trigger alerts. 
- Test the signin flow.
- Test the card creator flow.

Depends on a PR in Certificates repo: https://github.com/NYPL-Simplified/Certificates/pull/11
Once that's merged you'll need to run, from the repo root: ` swift ../Certificates/SimplyE/iOS/KeyObfuscator.swift`

I don't plan to include this change in the upcoming release 3.4.0, unless the release process stagnates.

n/a

@ettore


Simplified/NYPLAlertUtils.swift
```diff
@@ -38,13 +39,17 @@ import UIKit
...
-    } else if message.isEmpty { // Handle unassigned message
-      message = "UnknownError"
+
+    if message.isEmpty {
+      // since it wasn't a networking or Adobe DRM error, show the error
+      // description if present
+      if let errorDescription = error?.localizedDescription, !errorDescription.isEmpty {
+        message = errorDescription
+      } else {
+        message = "UnknownError"
```
kyles-ep, line 49: We should consider logging cases when this happens since it is unexpected. This way we can be warned via Crashlytics when we have missing localizations for errors. Can we fall back to english errors if localization is missing or does `localizedDescription` do that already?

ettore: If the NSLocalizedDescriptionKey is populated in `userInfo`, it will be used no matter what. And if it's constructed with NSLocalizedString("key"...) (as it should) then if "key" is not localized it will default to "key" as the not-localized value.

I'll see if I can add a logging call there since it would be nice to avoid that UnknownError.

PR #1132, opened 2020-05-22
SIMPLY-2710 Integrate auth and eligibility API calls for juvenile card creation

ettore:
**What's this do?**
It adds a button in the Account sign-in VC under Settings tab to add a dependent card for a signed-in NYPL user. Tapping this button performs a first auth call to get an access token which is then used to determine the user eligibility to create dependent cards.

This PR also create a new workspace to easily work on the card creator framework.

https://jira.nypl.org/browse/SIMPLY-2710
This is the full flow:
https://docs.google.com/drawings/d/1ZbddLroRGSgsnRACqoPF0uOx1z_q-sJGClKLqXjZWNw/edit?ts=5eb43c0c
This is the BRD:
https://docs.google.com/document/d/1IexiL8yhfHLTsSNG3D7E5rL-E_ph0HY4_aTyizbxk4M/edit

Currently this can be tested only by hitting the `qa-platform.nypl.org` endpoints. This requires a special build since the URL is hardcoded in. The flow and the BRD explain the possible error situations to be tested. There are the usual error situations related to networking, those can be easily tested with a proxy or by simulating lack of connectivity.

This is merging into a feature branch.

Yes

@ettore


Simplified/en.lproj/Localizable.strings
```diff
@@ -213,6 +211,10 @@
 "LocationRequiredMessage" = "Location information is required for Sign Up. Please allow access to location information in the iOS settings";
 "Error parsing user profile document." = "Error parsing user profile document. Please try again.";
 "Trouble locating DRM in profile document." = "Trouble locating DRM in profile document. Please try again.";
+"SigningInAgree" = "By signing in, you agree to the End User License Agreement.";
+"Want a card for your child?" = "Want a card for your child?";
+"We are unable to read your library card, which is necessary in order to create a dependent card." = "We are unable to read your library card, which is necessary in order to create a dependent card.";
+"Try to sign out, sign back in, then try again." = "Try to sign out, sign back in, then try again.";
```
kyles-ep, line 217: I take it all these strings were run by or came from @killerpuppytails ?

ettore: not yet, but I am compiling a list of strings for these edge error cases and I was thinking to send it over closer to the end.

PR #1134, opened 2020-05-30
Use authorization_identifier for juvenile APIs

ettore:
**What's this do?**
Uses the authorization_identifier to source the barcode to provide to the juvenile-creation APIs.

The barcode field in NYPLUserAccount may contain either a barcode or the username, but the Platform APIs for juvenile accounts require to specify what you're passing in (either a barcode or username), but we on the client don't know what the user has signed in with.

For NYPL accounts, and only for NYPL Accounts, the NYPLUserAccount::authorization_identifier instead always contains a barcode: this can be a 14-digit NYPL-issued barcode, or a 16-digit NYC ID (issued by the city of NY). This was confirmed by @leonardr.

Assuming you verified the juvenile flow in other preceding tickets by signing in to SimplyE via a NYPL barcode, gather the username for that same user, and sign in via the username.
Make sure that user didn't already create 3 juvenile accounts.
Then start the juvenile flow: you should still be eligible to create a juvenile account. I don't think you need to complete the full flow to verify this ticket, since a first eligibility api call is made before even showing the first screen.
It would be good to repeat this by signing in to SimplyE with a NYC ID.

Merging to feature branch.

Yes. I added comments in this PR based on the discussion with @leonardr and @EdwinGuzman.

@ettore

### DIFF
--- a/Cartfile
+++ b/Cartfile
@@ -1,6 +1,6 @@
 binary "Crashlytics.json"
 binary "https://dl.google.com/dl/firebase/ios/carthage/FirebaseAnalyticsBinary.json"
-github "NYPL-Simplified/CardCreator-iOS" ~> 1.1.2
+github "NYPL-Simplified/CardCreator-iOS" ~> 1.2
 github "NYPL-Simplified/NYPLAEToolkit" "master"
 github "NYPL-Simplified/NYPLAudiobookToolkit" "master"
 github "PureLayout/PureLayout" ~> 3.1.2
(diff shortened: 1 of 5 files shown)


leonardr:
To be clear, different libraries can set the `authorization_identifier` to whatever they want -- for Internet Archive it's an email address. But for NYPL it will always be a numeric barcode.

ettore:
> To be clear, different libraries can set the `authorization_identifier` to whatever they want -- for Internet Archive it's an email address. But for NYPL it will always be a numeric barcode.

Thanks for the clarification @leonardr  -- I have updated the PR and code comments/docs.

PR #1141, opened 2020-06-23
SIMPLY-224 Support Overdrive Audiobook

ErnestFan:
**What's this do?**
Support Overdrive Audiobook

https://jira.nypl.org/browse/SIMPLY-224

Play Overdrive Audiobook
Test audiobook after download expired

Right now this is using a test branch of NYPLAEToolkit and NYPLAudiobookToolkit
We should update the master branch of both frameworks and use those after QA review

No

@ErnestFan


Simplified/NYPLMyBooksDownloadCenter.m
```diff
@@ -774,6 +799,36 @@ - (void)startDownloadForBook:(NYPLBook *const)book
...
+                              book:(NYPLBook *)book {
+  NSURLSessionDownloadTask *const task = [self.session downloadTaskWithRequest:request];
+  
+  self.bookIdentifierToDownloadInfo[book.identifier] =
+    [[NYPLMyBooksDownloadInfo alloc]
+     initWithDownloadProgress:0.0
+     downloadTask:task
+     rightsManagement:NYPLMyBooksDownloadRightsManagementUnknown];
+  
+  self.taskIdentifierToBook[@(task.taskIdentifier)] = book;
```
ettore, line 812: wondering if we should have a safety guard here to check if `book` is not nil to prevent a possible crash.

Simplified/NYPLMyBooksDownloadCenter.m
```diff
@@ -774,6 +799,36 @@ - (void)startDownloadForBook:(NYPLBook *const)book
...
+   state:NYPLBookStateDownloading
+   fulfillmentId:nil
+   readiumBookmarks:nil
+   genericBookmarks:nil];
+  
+  // It is important to issue this immediately because a previous download may have left the
+  // progress for the book at greater than 0.0 and we do not want that to be temporarily shown to
+  // the user. As such, calling |broadcastUpdate| is not appropriate due to the delay.
+  [[NSNotificationCenter defaultCenter]
+   postNotificationName:NYPLMyBooksDownloadCenterDidChangeNotification
```
ettore, line 828: not sure I understand the reason to post this notification. At this point we just started a download, right? I thought this notification was for when the download has concluded....

ErnestFan: I believe it's to update the information shown to user if previous download failed. 
This part of code originally existed and I am only refactoring them into a function to avoid duplication.

Simplified/NYPLBookAcquisitionPath.h
```diff
@@ -12,6 +12,8 @@ static NSString * const _Nonnull ContentTypeFindaway = @"application/vnd.library
 static NSString * const _Nonnull ContentTypeOpenAccessAudiobook = @"application/audiobook+json";
 static NSString * const _Nonnull ContentTypeOpenAccessPDF = @"application/pdf";
 static NSString * const _Nonnull ContentTypeFeedbooksAudiobook = @"application/audiobook+json;profile=\"http://www.feedbooks.com/audiobooks/access-restriction\"";
+static NSString * const _Nonnull ContentTypeOverdriveCatalog = @"application/atom+xml;type=entry;profile=opds-catalog";
```
ettore, line 15: I'm really sorry I didn't catch this sooner, I apologize for the back and forth. But we should avoid putting static definitions in a header file, because header files can be included anywhere, and therefore you will continue to _define_ identically named symbols wherever you include the file.

The right way is to move these to the .m if they are only used there. If they used somewhere else, as I suspect, then you should declare them as (e.g.)
```
extern NSString * const _Nonnull ContentTypeOverdriveCatalog;
```
and in the .m:
```
NSString * const _Nonnull ContentTypeOverdriveCatalog = @"application/atom+xml;type=entry;profile=opds-catalog";
```

Simplified/NYPLBookContentType.m
```diff
@@ -5,7 +5,8 @@ NYPLBookContentType NYPLBookContentTypeFromMIMEType(NSString *const string)
 {
   if ([string isEqualToString:ContentTypeFindaway] ||
       [string isEqualToString:ContentTypeOpenAccessAudiobook] ||
-      [string isEqualToString:ContentTypeFeedbooksAudiobook]) {
+      [string isEqualToString:ContentTypeFeedbooksAudiobook] ||
+      [string isEqualToString:ContentTypeOverdriveAudiobook]) {
```
kyles-ep, line 9: If we have this many types, we might want to consider checking if the string matches an element in an array, where the array contains audiobook types.

Simplified/NYPLMyBooksDownloadCenter.m
```diff
@@ -723,11 +742,38 @@ - (void)startDownloadForBook:(NYPLBook *const)book
...
       [self startBorrowForBook:book attemptDownload:YES borrowCompletion:nil];
+    } else if ([book.distributor isEqualToString:OverdriveDistributorKey]) {
+      NSURL *URL = book.defaultAcquisition.hrefURL;
+        
+      [[OverdriveAPIExecutor shared] fulfillBookWithUrlString:URL.absoluteString
+                                                     username:[[NYPLUserAccount sharedAccount] barcode]
+                                                          PIN:[[NYPLUserAccount sharedAccount] PIN]
+                                                   completion:^(NSDictionary<NSString *,id> * _Nullable responseHeader, NSError * _Nullable _) {
+        if (!responseHeader[@"X-Overdrive-Scope"] || !responseHeader[@"Location"]) {
+          return;
```
kyles-ep, line 753: Does this warrant logging?

Simplified/NYPLMyBooksDownloadCenter.m
```diff
@@ -723,11 +742,38 @@ - (void)startDownloadForBook:(NYPLBook *const)book
...
+        }
+          
+        if ([[OverdriveAPIExecutor shared] patronToken] && ![[[OverdriveAPIExecutor shared] patronToken] isExpired]) {
+          // Use existing Patron Token
+          NSURLRequest *request = [[OverdriveAPIExecutor shared] getManifestRequestWithUrlString:responseHeader[@"Location"]];
+          [self addDownloadTaskWithRequest:request book:book];
+        } else {
+          [[OverdriveAPIExecutor shared] refreshPatronTokenWithKey:NYPLSecrets.overdriveClientKey secret:NYPLSecrets.overdriveClientSecret username:[[NYPLUserAccount sharedAccount] barcode] PIN:[[NYPLUserAccount sharedAccount] PIN] scope:responseHeader[@"X-Overdrive-Scope"] completion:^(NSError * _Nullable error) {
+            if (error) {
+              return;
```
kyles-ep, line 763: Is this error already logged?

PR #1145, opened 2020-07-01
SIMPLY-2843 Don't load catalog if missing URL + extensive logging

ettore:
**What's this do?**
Performs checks to verify we have a non-nil URL in NYPLRemoteViewController before issuing the network request. This prevents a situation where the catalog page would appear completely blank (with no Try Again button) because the request failed since it was nil.
This also adds extensive logging for catalog handling, hopefully logging every code paths there.

https://jira.nypl.org/browse/SIMPLY-2843

The way I reproduce the problem of having a blank + unresponsive catalog page was to:
- install SimplyE 3.4.0, download a book in Brooklyn lib, open ereader.
- background the app, do not force quit.
- use the phone for a day or so, making sure not to force-quit SimplyE.
- the next day, open SimplyE. At this point I was able to see the blank catalog error. 
- Repeat the above with 3.4.1. With this build I did not see it anymore.

Merging into hot fix branch based off of `master` since  I think we want to release a fix for SIMPLY-2843 and SiMPLY-2853 quickly.

added docs in the codebase

@ettore


Simplified/NYPLCatalogFeedViewController.m
```diff
@@ -82,8 +95,17 @@ + (UIViewController *)navigationFeedWithData:(NYPLXML *)data remoteVC:(NYPLRemot
...
+      [vc loadWithURL:url];
+    } else {
+      [NYPLErrorLogger logErrorWithCode:NYPLErrorCodeNoURL
+                                context:NSStringFromClass([self class])
+                                message:@"Server response for age verification lacks a URL to load."
+                               metadata:@{
+                                 @"ageAboveLimit": @(ageAboveLimit),
+                                 @"gateElementXMLAttributes": gatedXML.attributes,
+                               }];
+    }
```
kyles-ep, line 108: What is the user experience if we don't load the URL?

Simplified/NYPLCatalogGroupedFeedViewController.m
```diff
@@ -373,8 +373,15 @@ - (void)previewingContext:(__unused id<UIViewControllerPreviewing>)previewingCon
...
-  [self.remoteViewController load];
+
+  if (newURL != nil) {
+    [self.remoteViewController loadWithURL:newURL];
+  } else {
+    [NYPLErrorLogger logErrorWithCode:NYPLErrorCodeNoURL
+                              context:NSStringFromClass([self class])
+                              message:@"Catalog facet missing href URL"
+                             metadata:nil];
+  }
```
kyles-ep, line 384: Is there feedback to the user on what happens if the URL is nil?

ettore: Good catch, I added the Reload UI here.

Simplified/NYPLCatalogFeedViewController.m
ettore: also added the Reload UI here. The copy was validated by Risa.

PR #1150, opened 2020-07-10
SIMPLY-2843, SIMPLY-2853, SIMPLY-2857: Recover from errors on Catalog and Accounts pages

ettore:
**What's this do?**
SIMPLY-2843: In case the catalog fails to load because of a nil URL, the catalog screen will now try to reload both the library accounts list and the current library authentication document. 
SIMPLY-2843, SIMPLY-2853: In case the library accounts list fails, the Settings screen would appear empty and the user would be stuck in the water. This PR adds a Reload button to allow the user to recover from that.
SIMPLY-2857: Adds extensive logging for book download errors.

https://jira.nypl.org/browse/SIMPLY-2843
https://jira.nypl.org/browse/SIMPLY-2853
https://jira.nypl.org/browse/SIMPLY-2857

The easiest way would be to use a proxy and rewrite the responses. I've used Charles, where one can do the following:
1. enable SSL proxying on device per https://www.charlesproxy.com/documentation/using-charles/ssl-certificates
2. once you are able to see decrypted requests, from the Tools menu choose "Rewrite..."
3. Click "Add" on the left of the dialog, then add these 2 locations:
https://libraryregistry.librarysimplified.org/libraries
https://circulation.librarysimplified.org/*/authentication_document
4. on the bottom right of the dialog, click on Add and enable:
- Response Status --> Replace with "500 Internal Server Error"
- Body for Response --> Replace with any string

response-status-rewrite
body-rewrite
rewrite-summary

Once you have that, make sure to clear the cache of the app from the Settings screen. Otherwise the network won't even be hit for the 2 libraries and authentication document requests, since we cache those. 

charles-main

Then launch the app and you should see a "Try Again" button in both the catalog screen and on the Settings > Accounts page. If you then disable response rewrites in Charles and hit "Try Again", the pages should reload correctly.

merging to hot fix branch, since this is fairly urgent.

Yes. I added a ton of comments in the code.

@ettore

### DIFF
--- a/Simplified/NYPLMyBooksDownloadCenter.m
+++ b/Simplified/NYPLMyBooksDownloadCenter.m
@@ -162,7 +162,7 @@ totalBytesExpectedToWrite:(int64_t const)totalBytesExpectedToWrite
 
 - (void)URLSession:(__attribute__((unused)) NSURLSession *)session
       downloadTask:(NSURLSessionDownloadTask *const)downloadTask
-didFinishDownloadingToURL:(NSURL *const)location
+didFinishDownloadingToURL:(NSURL *const)tmpSavedFileURL
 {
   NYPLBook *const book = self.taskIdentifierToBook[@(downloadTask.taskIdentifier)];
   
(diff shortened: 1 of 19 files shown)


Simplified/NYPLMyBooksDownloadCenter.m
```diff
@@ -266,21 +287,32 @@ - (void)URLSession:(__attribute__((unused)) NSURLSession *)session
...
+
         [[NSFileManager defaultManager]
-         removeItemAtURL:[self fileURLForBookIndentifier:book.identifier]
-         error:NULL];
-        
+         removeItemAtURL:finalFileURL
+         error:&removeError];
+
         success = [[NSFileManager defaultManager]
-                   moveItemAtURL:location
```
ErnestFan, line 276: Can't we just use `replaceItemAtURL:withItemAtURL:backupItemName:options:resultingItemURL:error:` instead of having two different operations?

ettore: good point, probably... but I don't want to change this code too much in the hot fix branch unless I have to. Right now I don't know if this is actually causing any issues.

ErnestFan: Sure, not an important change anyways

PR #1153, opened 2020-07-21
SIMPLY-2846 Access shared accounts dictionary in a thread-safe way

ettore:
**What's this do?**
Puts all accesses to the `AccountManager::accountSets` dictionary behind a concurrent queue using a barrier for write operations to ensure thread-safety for that dictionary.

https://jira.nypl.org/browse/SIMPLY-2846
This crash happened while _reading_ from the dictionary, with a BAD_ACCESS exception reading from 0x0 address. Since that dictionary is on AccountManager, which is a big Singleton class, it can be accessed in R/W mode from many threads, leading to possible memory corruption.

This crash was hard to reproduce so I don't have clear steps on how to do that. I think QA should focus on doing a testing on the sign-in/sign-out flows, account management, switching different libraries. All the section of the regression suite involving accounts should be executed.
I'll see if I can add some unit tests, but it might be difficult because AccountManager is a singleton also interfacing with the keychain.

merging to develop

in code

@ettore

### DIFF
--- a/Simplified/AccountsManager.swift
+++ b/Simplified/AccountsManager.swift
@@ -27,11 +27,17 @@ private let prodUrlHash = prodUrl.absoluteString.md5().base64EncodedStringUrlSaf
     return shared
   }
   
-  var accountSet: String
+  private var accountSet: String
   private var accountSets = [String: [Account]]()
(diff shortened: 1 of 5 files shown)


ettore:
@kyles-ep I know you're busy these days but this is not urgent... whenever you have some time in the coming week I'd appreciate your eyes on this

PR #1163, opened 2020-08-08
SIMPLY-2502 Feedbook DRM

ErnestFan:
**What's this do?**
Strip the header of the PEM key and then store it into Keychain
Update dependencies branch and hash
Bump iOS version to 10

[SIMPLY-2502](https://jira.nypl.org/browse/SIMPLY-2502?filter=-1)

Open an audiobook in DPLA test library

N/A

N/A

@ErnestFan


ErnestFan:
@ettore I removed `GoogleDataTransportCCTSupport.framework` because I had issue building the project with it.
Let me know if any actions is needed.

Simplified/Utilities/RSAUtils.swift
```diff
@@ -0,0 +1,9 @@
+import CommonCrypto
+
+class RSAUtils {
```
ettore, line 3: didn't we have this also in NYPLAudiobookToolkit? If so can't we just expose and use that?

Simplified/NYPLKeychainManager.swift
```diff
@@ -160,38 +162,67 @@ import Foundation
...
-        }
+        logKeychainError(for: status, with: "FeedbookKeyManagement Error:")
+      }
+    }
+  }
+    
+  private class func manageFeedbookDrmPrivateKey() {
+    for vendor in AudioBookVendors.allCases {
+      guard let privateKeyString = NYPLSecrets.drmCertificate(forVendor: vendor),
+        let privateKeyData = Data(base64Encoded: RSAUtils.stripPEMKeyHeader(privateKeyString)),
```
ettore, line 173: can you add a comment on why we need to strip the PEM key header?

ettore:
> @ettore I removed `GoogleDataTransportCCTSupport.framework` because I had issue building the project with it.
> Let me know if any actions is needed.

I think this was caused by the bump in version number for google analytics, but I would leave that upgrade for the future.

Simplified/NYPLKeychainManager.swift
```diff
@@ -160,38 +162,67 @@ import Foundation
...
+      case errSecUserCanceled:
+        errMsg = "errSecUserCanceled"
+      case errSecBadReq:
+        errMsg = "errSecBadReq"
+      default:
+        errMsg = "Unknown OSStatus: \(status)"
       }
     }
+    
+    Log.error(#file, "\(message) \(errMsg)")
```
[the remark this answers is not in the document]
ErnestFan: It calls a Crashlytics log function `CLSLogv` down the line but not a report, but we would know if the client fails to retrieve item from keychain

Simplified/NYPLKeychainManager.swift
```diff
@@ -160,38 +162,67 @@ import Foundation
...
-          }
-        }
+        logKeychainError(for: status, with: "FeedbookKeyManagement Error:")
+      }
+    }
+  }
+    
+  private class func manageFeedbookDrmPrivateKey() {
+    for vendor in AudioBookVendors.allCases {
+      guard let privateKeyString = NYPLSecrets.drmCertificate(forVendor: vendor),
```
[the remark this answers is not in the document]
ErnestFan: I already pushed the update since it's a rather small change, sorry I should have submitted a PR

Simplified.xcodeproj/project.pbxproj
```diff
@@ -2779,7 +2781,7 @@
 				GCC_PRECOMPILE_PREFIX_HEADER = YES;
 				GCC_PREFIX_HEADER = "Simplified/Simplified-Prefix.pch";
 				INFOPLIST_FILE = "Simplified/Simplified-Info.plist";
-				IPHONEOS_DEPLOYMENT_TARGET = 9.3;
+				IPHONEOS_DEPLOYMENT_TARGET = 10.0;
```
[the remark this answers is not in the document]
ErnestFan: [here](https://jira.nypl.org/browse/SIMPLY-2502?focusedCommentId=44636&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-44636)

Simplified/Utilities/RSAUtils.swift
ErnestFan: Sure, should I just update the NYPLAudiobookToolkit? Cause it's just adding two `public` mark on the class and function

Simplified/NYPLKeychainManager.swift
ErnestFan: added

Simplified/NYPLKeychainManager.swift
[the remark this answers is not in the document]
ettore: if it calls NYPLErrorLogger then that's fine, otherwise we should add one. Stuff logged by `Log.error` / `NYPLLog` is kind of unsearchable in Crashlytics.

Simplified/NYPLKeychainManager.swift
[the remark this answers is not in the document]
ettore: yeah, if it's just a new constant being added, sure, no problem, but any new code no matter how small I think should be reviewed. At least that's what I've been doing.  Nothing personal, just best practice. :) Also this way everybody is on the same page (e.g. I didn't know I had to update Certificates).

Anyway, for this change I'm taking a look in Certificates so no worries.

Simplified/Utilities/RSAUtils.swift
ettore: I think making that public in NYPLAudiobookToolkit sounds great. Less code duplication.

Long term, we should think about creating a more generic NYPLUtilities framework... or something that allows us to share general utilities across repos.

Simplified/NYPLKeychainManager.swift
```diff
@@ -160,38 +162,72 @@ import Foundation
...
         
-        Log.error(#file, "FeedbookKeyManagement Error: \(errMsg)")
+      let status = SecItemAdd(addQuery as CFDictionary, nil)
+      if status != errSecSuccess && status != errSecDuplicateItem {
+        logKeychainError(for: status, with: "FeedbookDrmPrivateKeyManagement Vendor: \(vendor.rawValue) Error:")
+      }
+    }
+  }
+    
+  private class func logKeychainError(for status: OSStatus, with message: String) {
```
ettore, line 197: I would also pass in the `vendor` to this function and log is as part of the context, like `"\(keyChainManagement) \(vendor)"` -- reason is context and code are easily searchable.

ErnestFan: Added

PR #1164, opened 2020-08-10
SIMPLY-2962 Fix Overdrive ePubs download failure

ErnestFan:
**What's this do?**
Fix Overdrive ePubs download failure
Remove duplicated acquisition
Bump version number

[SIMPLY-2962](https://jira.nypl.org/browse/SIMPLY-2962)

Open Overdrive ePubs and audiobook to make sure they both download and open without issues

N/A

N/A

@ErnestFan


Simplified/NYPLBookAcquisitionPath.m
```diff
@@ -42,7 +41,6 @@ - (instancetype _Nonnull)initWithRelation:(NYPLOPDSAcquisitionRelation const)rel
       ContentTypeOpenAccessAudiobook,
       ContentTypeOpenAccessPDF,
       ContentTypeFeedbooksAudiobook,
-      ContentTypeOverdriveCatalog,
```
ettore, line 45: can you explain why we should remove this? wouldn't this remove OD audiobooks from the visible catalog?

ErnestFan: Just found out the content of OverdriveCatalog is same as the opds catalog, so this does nothing.
Probably OD does not have its own catalog acquisition and I made a mistake at the very beginning.

ErnestFan:
Will create PR from another fork

PR #1165, opened 2020-08-10
SIMPLY-2962 Fix Overdrive ePubs download failure

ErnestFan:
**What's this do?**
Fix Overdrive ePubs download failure
Remove duplicated acquisition
Bump version number

[SIMPLY-2962](https://jira.nypl.org/browse/SIMPLY-2962)

Open Overdrive ePubs and audiobook to make sure they both download and open without issues

N/A

N/A

@ErnestFan


Simplified/NYPLBookAcquisitionPath.m
```diff
@@ -2,7 +2,6 @@
 
 #import "NYPLBookAcquisitionPath.h"
 
-NSString * const _Nonnull ContentTypeOverdriveCatalog = @"application/atom+xml;type=entry;profile=opds-catalog";
```
leonardr, line 5: I think this is being removed because it's redundant -- it's the same media type as `ContentTypeOPDSCatalog`. Is this right?

ErnestFan: Yeah, I added it when I first worked on Overdrive and didn't realize it's duplicated.

PR #1169, opened 2020-08-14
SIMPLY-2987 Add range check to avoid out of bound

ErnestFan:
**What's this do?**
Add range check to avoid out of bound

[SIMPLY-2987](https://jira.nypl.org/projects/SIMPLY/issues/SIMPLY-2987)

Unable to reproduce issue

N/A

N/A

@ErnestFan

### DIFF
--- a/Simplified/Utilities/Authentication/NYPLUserAccountFrontEndValidation.m
+++ b/Simplified/Utilities/Authentication/NYPLUserAccountFrontEndValidation.m
@@ -56,7 +56,7 @@ replacementString:(NSString *)string
     }
 
     // Usernames cannot be longer than 25 characters.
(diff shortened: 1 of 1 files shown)


ettore:
Perhaps some QA steps could be to verify that copy and pasting in the barcode and password fields is not broken.

ErnestFan:
> Perhaps some QA steps could be to verify that copy and pasting in the barcode and password fields is not broken.

Tried that, and does not go out of bound

ettore:
> > Perhaps some QA steps could be to verify that copy and pasting in the barcode and password fields is not broken.
> 
> Tried that, and does not go out of bound

sure, I just meant for the QA guys as a smoke check we didn't break anything

PR #1175, opened 2020-08-25
SIMPLY-2935 Announcement feature

ErnestFan:
**What's this do?**
Add announcement in auth document
Present announcement if patron is logged in and announcement has not been present
Keep track of presented announcements

[SIMPLY-2935](https://jira.nypl.org/browse/SIMPLY-2935?filter=-1)

Go to NYPL QA server and log in, an announcement should pop up.
Dismiss the announcement and it should not show up anymore.

N/A

N/A

@ErnestFan


Simplified/NYPLAnnouncementManager.swift
```diff
@@ -0,0 +1,65 @@
+import Foundation
+
+private let announcementKey: String = "NYPLPresentedAnnouncementKey"
+private let filename: String = "NYPLPresentedAnnouncementsList"
+
+class NYPLAnnouncementManager {
+  private static var presentedAnnouncement: Set<String> = Set<String>()
```
ettore, line 7: this static variable is not thread-safe, so it might be good to add some docs to NYPLAnnouncementManager to say this class is not thread-safe and let callers deal with it.  That being said, do the callers need to worry about that?

also rename `presentedAnnouncement` to `presentedAnnouncements` since it contains a set of announcements.

NYPLAnnouncementManager maybe it can be renamed to NYPLAnnouncementBusinessLogic? Seems like a more precise name. Also I usually avoid "Manager" in class names (even if Apple themselves abuse this a lot, unfortunately) because it usually a sign of code smell, check out this post: https://blog.codinghorror.com/i-shall-call-it-somethingmanager/

Simplified/NYPLAnnouncementManager.swift
```diff
@@ -0,0 +1,65 @@
...
+  }
+  
+  // MARK: - Read
+    
+  class func shouldPresentAnnouncement(id: String) -> Bool {
+    guard let filePathURL = FileManager.default.urls(for: .documentDirectory, in: .userDomainMask).first?.appendingPathComponent(filename),
+      let filePathData = try? Data(contentsOf: filePathURL),
+      let unarchived = try? NSKeyedUnarchiver.unarchiveTopLevelObjectWithData(filePathData),
+      let presented = unarchived as? Set<String> else {
+        return true
```
ettore, line 27: shouldn't this be false? or is it that if the file is not found it means it's a new announcement? but even in that case if the file is found and decoding fails, seems like we should return `false`... I could be wrong though. Some comments would be really useful here to explain the intended logic.

Simplified/NYPLAnnouncementManager.swift
```diff
@@ -0,0 +1,65 @@
...
+    }
+    
+    do {
+      let codedData = NSKeyedArchiver.archivedData(withRootObject: presentedAnnouncement)
+      try codedData.write(to: filePathURL)
+    } catch {
+      NYPLErrorLogger.logError(error,
+                               summary: "Fail to write Presented Announcements file to local storage",
+                               message: nil,
+                               metadata: nil)
```
ettore, line 62: I'd pass the filePathURL and the `presentedAnnouncement` in the metadata

Simplified/NYPLSettingsAccountDetailViewController.m
```diff
@@ -207,7 +207,7 @@ - (void)viewDidLoad
     [self.view addSubview:activityIndicator];
     [activityIndicator startAnimating];
     self.loading = true;
-    [self.selectedAccount loadAuthenticationDocumentWithCompletion:^(BOOL success) {
+    [self.selectedAccount loadAuthenticationDocumentWithUserAccountProvider:self completion:^(BOOL success) {
```
ettore, line 210: here you could also pass the businessLogic and make that the implementor of `NYPLUserAccountProvider` instead.

Simplified/NYPLAnnouncementManager.swift
```diff
@@ -0,0 +1,65 @@
...
+  
+  // MARK: - Write
+    
+  class func addPresentedAnnouncement(id: String) {
+    presentedAnnouncement.insert(id)
+    
+    storePresentedAnnouncementToFile()
+  }
+
+  // For testing use
```
ettore, line 42: if it's for testing only I would move this into an extension. A swift sort-of pattern is to do:
```
extension YourClass {
  func testing_deletePresentedAnnouncement() {
  ...
  }
}
```

Simplified/NYPLAnnouncementManager.swift
```diff
@@ -0,0 +1,65 @@
...
+          let vc = NYPLAnnouncementViewController(announcement: announcement)
+          NYPLRootTabBarController.shared()?.safelyPresentViewController(vc, animated: true, completion: nil)
+        }
+      }
+    }
+  }
+  
+  // MARK: - Read
+    
+  class func shouldPresentAnnouncement(id: String) -> Bool {
```
ettore, line 22: can this be fileprivate? You could add a `testing_shouldPresent...` wrapper in an extension if you need it for testing.

Simplified/NYPLAnnouncementManager.swift
ErnestFan: Since we are making `NYPLAnnouncementBusinessLogic` a singleton, I have moved this part of code into the init.
So we only read the file once the singleton being created, and we will use the property `presentedAnnouncements` to keep track of the announcements for the life cycle.
So if the file is not found, cannot be decoded or corrupted, we will overwrite the file with the new presented announcements set.

Simplified/AccountsManager.swift
```diff
@@ -10,7 +10,7 @@ private let prodUrlHash = prodUrl.absoluteString.md5().base64EncodedStringUrlSaf
 
 /// Manage the library accounts for the app.
 /// Initialized with JSON.
-@objcMembers final class AccountsManager: NSObject
+@objcMembers final class AccountsManager: NSObject, NYPLSignedInStateProvider
```
ettore, line 13: I think this and the implementation at line 298 now can be removed, right?

Simplified/NYPLAnnouncementBusinessLogic.swift
```diff
@@ -0,0 +1,85 @@
...
+          let vc = NYPLAnnouncementViewController(announcement: announcement)
+          NYPLRootTabBarController.shared()?.safelyPresentViewController(vc, animated: true, completion: nil)
+        }
+      }
+    }
+  }
+  
+  // MARK: - Read
+    
+  fileprivate func restorePresentedAnnouncements() {
```
ettore, line 27: I think this and `shouldPresentAnnouncement` can be private now... I made a mistake in suggesting `fileprivate` earlier, sorry.... pretty much the only time you need `fileprivate` is if you need accessing something outside of a class in the same file, otherwise private is fine even from an `extension`.

ettore: same for all other fileprivate stuff in this class

Simplified/Account.swift
```diff
@@ -321,10 +325,12 @@ private let accountSyncEnabledKey        = "NYPLAccountSyncEnabledKey"
 
 
   /// Load authentication documents from the network or cache.
```
ettore, line 327: I would add here that this might also trigger the presentation of an announcement if one is present.

Simplified/NYPLAnnouncementBusinessLogic.swift
```diff
@@ -0,0 +1,85 @@
...
+  }
+    
+  fileprivate func shouldPresentAnnouncement(id: String) -> Bool {
+    return !presentedAnnouncements.contains(id)
+  }
+  
+  // MARK: - Write
+    
+  /// Add presented announcement to storage
+  /// No guarantees are being made about whether this is called on the main thread or not
```
ettore, line 44: hmm, not sure this belongs here... was this referred to a completion block maybe?

I'd just say "this class is not thread safe" as documentation for the class rather than each method

Simplified/NYPLSettingsAccountDetailViewController.m
```diff
@@ -402,6 +402,12 @@ - (void)scanLibraryCard
   }];
 }
 
+#pragma mark - NYPLUserAccountProvider Protocol
+
+- (BOOL)isSignedIn {
```
ettore, line 407: this and the NYPLSignedInStateProvider compliance at the top can also be removed I think

PR #1177, opened 2020-08-31
SIMPLY-3014 Convert JWA to PEM private key

vladimirfedorov:
**What's this do?**
Converts JWK data to RSA private key

We need to extract the private key for Feedbook DRM

Run NYPLJWKConversionTest

No dependencies, not releasing to production. We need this code for https://jira.nypl.org/browse/SIMPLY-2944

New classes are documented

@vladimirfedorov


SimplifiedTests/jwk_private
```diff
@@ -0,0 +1,25 @@
+MIIEpAIBAAKCAQEA7GcBZtUVccXneZdDu5KjhxUOo6cxqM718Uaiw6kZns+AGz/2
```
ettore, line 1: is this and the json file above the actual key from Feedbooks or just a random test key? sorry if it's an obvious question, just want to make sure.

vladimirfedorov: it actually was, I was testing to check if it matches the key in the secrets file. replaced with a random one.

Simplified/Utilities/JWK/JWK.swift
```diff
@@ -0,0 +1,50 @@
...
+ distributed under the License is distributed on an "AS IS" BASIS,
+ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ See the License for the specific language governing permissions and
+ limitations under the License.
+ */
+
+
+import Foundation
+
+public struct JWK: Codable {
```
ettore, line 17: can you add a comment about what kind of key(s) this can represent? I think it's for a private key but not sure if there are additional details regarding which allowed algorithms/formats etc

SimplifiedTests/jwk_private
ettore: thanks -- can you also make sure to squash the 2 commits so at least it's a bit more hidden?

Simplified/Utilities/JWK/JWK.swift
vladimirfedorov: JWK is a set of RSA variables, both private and public keys can be constructed with that variables. Also JWK is not limited to one specific cryptographic algorithm, so I don't  even know what to add here... we are interested in RSA data of the private key

vladimirfedorov: May be adding such a comment to our extension will be better as it is more specific? JWK structure is too general

ettore: sure, that's fine -- I'm sort of a newbie when it comes to JWK in general, sorry if my request was unclear. I sort of remember that JWK can be used for public and private keys, and while I do see some private related fields, there were some additional fields I didn't recognize, hence my question.

SimplifiedTests/jwk_private
vladimirfedorov: sure, squashed

Simplified/Utilities/JWK/JWK.swift
vladimirfedorov: I have a link to the documentation in ASN1.swift - 
 Documentation ITU-T X.690: https://www.itu.int/rec/T-REC-X.690-201508-I/en
Added a link to RSA RFC to JWK+RSAKeys.sqift, it describes the key structure

ettore: thanks!

Simplified.xcodeproj/project.pbxproj
```diff
@@ -2830,7 +2892,7 @@
 					"$(inherited)",
 				);
 				INFOPLIST_FILE = "Simplified/Simplified-Info.plist";
-				IPHONEOS_DEPLOYMENT_TARGET = 10.0;
+				IPHONEOS_DEPLOYMENT_TARGET = 9.3;
```
ErnestFan, line 2895: The reason we bump the version to 10.0 is because we are using `SecKeyCreateWithData` to create SecKey when we extract the PEM data from Keychain, in NYPLAudiobookToolkit. 
Unless we figure out how to create the Seckey without using this API, otherwise we need to keep the version at 10.0

vladimirfedorov: I see, changed back

PR #1178, opened 2020-09-01
SIMPLY-3016 Fix UIAlertController not init in main thread

ErnestFan:
**What's this do?**
Move UI related code into main thread to avoid crash

https://jira.nypl.org/browse/SIMPLY-3016

N/A

N/A

No

@ErnestFan

### DIFF
--- a/Simplified/NYPLReadiumViewSyncManager.m
+++ b/Simplified/NYPLReadiumViewSyncManager.m
@@ -148,69 +148,77 @@ const double RequestTimeInterval = 120;
       return;
     }
       
-    [[NSOperationQueue mainQueue] addOperationWithBlock:^{
-      NSDictionary *spineItemDetails = weakSelf.bookMapDictionary[responseJSON[@"idref"]];
-      NSString *elementTitle = spineItemDetails[@"tocElementTitle"];
-      if (!elementTitle) {
-        elementTitle = @"";
-      }
-                  
-      NSString *message = NSLocalizedString(@"Do you want to move to the page on which you left off?", nil);
-      NSAttributedString *atrString;
-      if (![elementTitle isEqualToString:@"Current Chapter"]) {
-        message = [message stringByAppendingString:[NSString stringWithFormat:@"<br><br>Chapter:\n&ldquo;%@&ldquo;", elementTitle]];
-        atrString = [[NSAttributedString alloc] initWithData:[message dataUsingEncoding:NSUTF8StringEncoding]
-                                                                         options:@{NSDocumentTypeDocumentAttribute: NSHTMLTextDocumentType,
-                                                                                   NSCharacterEncodingDocumentAttribute: @(NSUTF8StringEncoding)}
-                                                              documentAttributes:nil error:nil];
-      }
+    [self presentSyncReadingPositionAlertForResponse:responseJSON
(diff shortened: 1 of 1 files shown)


Simplified/NYPLReadiumViewSyncManager.m
```diff
@@ -205,11 +145,72 @@ - (void)syncReadingPositionForBook:(NSString *)bookID
...
       weakSelf.shouldPostLastRead = YES;
-    } else {
-      [[NSOperationQueue mainQueue] addOperationWithBlock:^{
-        [[NYPLRootTabBarController sharedController] safelyPresentViewController:alertController animated:YES completion:nil];
-      }];
+      return;
     }
+      
+    [[NSOperationQueue mainQueue] addOperationWithBlock:^{
+      NSDictionary *spineItemDetails = weakSelf.bookMapDictionary[responseJSON[@"idref"]];
```
ettore, line 152: can we move all the code inside the `addOperationWithBlock` closure to a new method? not sure if that will involve a gazillion parameters but it would be nice to break up this long deeply-nested method.

Simplified/NYPLReadiumViewSyncManager.m
```diff
@@ -137,66 +137,6 @@ - (void)syncReadingPositionForBook:(NSString *)bookID
     NYPLLOG_F(@"serverLocationString %@",serverLocationString);
     NYPLLOG_F(@"currentLocationString %@",currentLocationString);
 
-    NSDictionary *spineItemDetails = weakSelf.bookMapDictionary[responseJSON[@"idref"]];
```
ettore, line 140: this code was cut and pasted down below? no changes at all?

ErnestFan: No changes, except the `return` to get rid of the `else` bracket

Simplified/NYPLReadiumViewSyncManager.m
```diff
@@ -137,22 +137,43 @@ - (void)syncReadingPositionForBook:(NSString *)bookID
...
+    // 2 - The server and the client have the same page marked
+    // 3 - There is no recent page saved on the server
+    if ((currentLocationString && [deviceIDString isEqualToString:[NYPLUserAccount sharedAccount].deviceID]) ||
+        [currentLocationString isEqualToString:serverLocationString] ||
+        !serverLocationString) {
+      weakSelf.shouldPostLastRead = YES;
+      return;
     }
+      
+    [self presentSyncReadingPositionAlertForResponse:responseJSON
```
ettore, line 151: here I would use `weakSelf` for peace of mind... I'm not sure it's 100% needed, but the code before that is using it, so let's do the same.

Simplified/NYPLReadiumViewSyncManager.m
```diff
@@ -137,22 +137,43 @@ - (void)syncReadingPositionForBook:(NSString *)bookID
...
+                                      serverLocation:serverLocationString
+                                         withPackage:dictionary];
+  }];
+}
 
+- (void)presentSyncReadingPositionAlertForResponse:(NSDictionary *)response
+                                    serverLocation:(NSString *)location
```
ettore, line 158: can you rename this `serverLocation` for clarity?

PR #1181, opened 2020-09-08
SIMPLY-3017 Add nil check before accessing cache

ErnestFan:
**What's this do?**
Do a nil check before accessing cache to avoid crash

[SIMPLY-3017](https://jira.nypl.org/browse/SIMPLY-3017?filter=-1)

Unable to reproduce crash

N/A

N/A

@ErnestFan

### DIFF
--- a/Simplified/NYPLBookCoverRegistry.m
+++ b/Simplified/NYPLBookCoverRegistry.m
@@ -252,7 +252,7 @@ static NSUInteger const memoryCacheInMegabytes = 2;
 
 - (UIImage *)cachedThumbnailImageForBook:(NYPLBook *const)book
 {
-  if(!book.imageThumbnailURL) {
+  if(!book.imageThumbnailURL || !self.session.configuration.URLCache) {
     return nil;
   }
   


Simplified/NYPLBookCoverRegistry.m
```diff
@@ -252,7 +252,7 @@ - (void)thumbnailImagesForBooks:(NSSet *)books
 
 - (UIImage *)cachedThumbnailImageForBook:(NYPLBook *const)book
 {
-  if(!book.imageThumbnailURL) {
+  if(!book.imageThumbnailURL || !self.session.configuration.URLCache) {
```
ettore, line 255: so I don't think you need this additional check actually, because even if the cache is nil in objc the call to `cachedResponseForRequest` would not lead to a crash (because `nil.methodCall` results in nil essentially in objc).

it seems like my observation in the ticket was wrong. The nil check I was referring to was related to the argument of the `cachedResponseForRequest` call (`book.imageThumbnailURL`), but that's already done.

So I'm not sure where the crash originates, but since it's EXC_BAD_ACCESS KERN_INVALID_ADDRESS that refers to a dangling pointer on `book.imageThumbnailURL`, which likely is a deallocated object whose pointer is still being accessed for some reason. Tracking these errors is hard though: perhaps look at how that `imageThumbnailURL` property is assigned and what is its life path in the code. If you don't see any smoking gun (time box it) feel free to assign it to me for the future and move on to  something else since it's a rare crash.

ErnestFan: Sure, I will close this PR

PR #1182, opened 2020-09-08
SIMPLY-2904 Display barcode number

ErnestFan:
**What's this do?**
Display barcode number on account detail

[SIMPLY-2904](https://jira.nypl.org/browse/SIMPLY-2904?filter=-1)

Log into one of the libraries that support barcode display(eg. NYPL), and click "Show barcode"
A barcode image and number should show up

N/A

N/A

@ErnestFan

### DIFF
--- a/Simplified/NYPLSettingsAccountDetailViewController.m
+++ b/Simplified/NYPLSettingsAccountDetailViewController.m
@@ -1485,12 +1485,11 @@ didSelectRowAtIndexPath:(NSIndexPath *const)indexPath
                                           superviewWidth:self.tableView.bounds.size.width
                                                     type:NYPLBarcodeTypeCodabar];
 
-        NSString *barcodeText = self.selectedUserAccount.authorizationIdentifier;
         if (barcodeImage) {
           self.barcodeImageView = [[UIImageView alloc] initWithImage:barcodeImage];
           self.barcodeImageLabel = [[UILabel alloc] init];
           self.barcodeTextLabel = [[UILabel alloc] init];
-          self.barcodeTextLabel.text = barcodeText;
+          self.barcodeTextLabel.text = self.selectedUserAccount.barcode;
           self.barcodeTextLabel.font = [UIFont customFontForTextStyle:UIFontTextStyleBody];
           self.barcodeTextLabel.textAlignment = NSTextAlignmentCenter;
           self.barcodeImageLabel.text = NSLocalizedString(@"Show Barcode", nil);
--- a/Simplified/NYPLSettingsAccountDetailViewController.m
+++ b/Simplified/NYPLSettingsAccountDetailViewController.m
@@ -90,6 +90,10 @@ static const CGFloat sVerticalMarginPadding = 2.0;
 static const NSInteger sSection0AccountInfo = 0;
 static const NSInteger sSection1Sync = 1;
 
+// Constraint constants
+static const CGFloat sConstantZero = 0.0;
+static const CGFloat sConstantSpacing = 12.0;
+
 @implementation NYPLSettingsAccountDetailViewController
 
 /*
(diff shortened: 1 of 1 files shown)


Simplified/NYPLSettingsAccountDetailViewController.m
```diff
@@ -1275,11 +1278,15 @@ - (void)tableView:(__attribute__((unused)) UITableView *)tableView
...
         self.barcodeHeightConstraint.constant = 0.0;
+        self.barcodeTextHeightConstraint.constant = 0.0;
+        self.barcodeTextLabelSpaceConstraint.constant = 0.0;
         self.barcodeLabelSpaceConstraint.constant = 0.0;
         self.barcodeImageLabel.text = NSLocalizedString(@"Show Barcode", nil);
         [[UIScreen mainScreen] setBrightness:self.userBrightnessSetting];
       } else {
         self.barcodeHeightConstraint.constant = 100.0;
+        self.barcodeTextHeightConstraint.constant = 30.0;
+        self.barcodeTextLabelSpaceConstraint.constant = -12.0;
```
ettore, line 1289: just wondering if we can put some of these numbers in constants instead of repeating them... like this one seems related to a spacer distance.

ettore (approved):

Simplified/NYPLSettingsAccountDetailViewController.m
```diff
@@ -1481,18 +1488,27 @@ - (UITableViewCell *)tableView:(__attribute__((unused)) UITableView *)tableView
         if (barcodeImage) {
           self.barcodeImageView = [[UIImageView alloc] initWithImage:barcodeImage];
           self.barcodeImageLabel = [[UILabel alloc] init];
+          self.barcodeTextLabel = [[UILabel alloc] init];
+          self.barcodeTextLabel.text = self.selectedUserAccount.barcode;
```
ettore, line 1492: hmm one problem here is that the `barcode` could actually be a username in some situations. I wonder if we should use the `authorizationIdentifier` instead? there should be some comments about this in the code base, but I don't recall exactly where.

ettore: check out the documentation for `NYPLUserAccount::authorizationIdentifier` and `NYPLUserAccount::barcode`  -- perhaps we can write a method in NYPLSignInBusinessLogic that tries to return the barcode for NYPL at least? I suppose it depends how much we care, and this is a question for Risa ( @killerpuppytails ). But if we just use `barcode` this can essentially display an actual barcode or a username, and the app wouldn't know which one it is.

ErnestFan: You are right, the `barcode` could be a username.

But there is a note on `autherizationIdentifier` saying that is for internal only and do not expose to public.

ettore: I just double checked -- no, that is referred to the ILS ID, but that's something different. The main point there is that only for NYPL we can say for certain that the authorizationIdenfier is a barcode.

ErnestFan: Then should we keep using `barcode`? and only display it if the `barcode` is number only and 14 digits? Although it can still be an username

killerpuppytails: So while we would *like* to show the NYPL patron's barcode even if they only enter a username*, it's not required and you can hide the "Show barcode" feature if it's just a username.

* as they might not remember their barcode and it is used exclusively on some databases

ettore (commented):
make sure to verify there's no new Auto Layout warnings being generated in the console log upon hiding/showing the barcode, other than that this looks good to me.

PR #1183, opened 2020-09-09
SIMPLY-3056 Show OE login choice in Accounts Settings

ettore:
**What's this do?**
- Don't add account selection button (top-left nav item in Catalogs, MyBooks, Holds) for OE.
- Refactor migrations for SE and OE. Add OE migration for AdobeDRM user/device IDs.
- Add preprocessor macros to identify SE / OE.
- Add strings for OE Tutorial screens.
- Add OE URLs in configuration file.
- Refactor Settings classes for OE in Swift. SE still uses old implementation.
- Rename OE icon to match SE's equivalent and ease launch process set-up.
- Add generic VC presentation utility.

Much of this code was taken from old PR for merging OE in SE: https://github.com/NYPL-Simplified/Simplified-iOS/pull/1070

All of the new files (except obvious exceptions) are only added to the OE target.

https://jira.nypl.org/browse/SIMPLY-3056

Not much to test beside making sure SE still works fine. These changes should leave SE unaltered though.

Merging to develop but this is not breaking (or changing, really) the SE target.

yes

@ettore


Simplified.xcodeproj/project.pbxproj
```diff
@@ -511,7 +525,6 @@
 		73FCA2E825005BA4001B0C5D /* main.m in Sources */ = {isa = PBXBuildFile; fileRef = A823D81C192BABA400B55DE2 /* main.m */; };
 		73FCA2E925005BA4001B0C5D /* NYPLCaching.swift in Sources */ = {isa = PBXBuildFile; fileRef = 733875662423E540000FEB67 /* NYPLCaching.swift */; };
 		73FCA2EA25005BA4001B0C5D /* UIView+NYPLViewAdditions.m in Sources */ = {isa = PBXBuildFile; fileRef = 1107835D19816E3D0071AB1E /* UIView+NYPLViewAdditions.m */; };
-		73FCA2EB25005BA4001B0C5D /* NYPLSettingsPrimaryTableViewController.m in Sources */ = {isa = PBXBuildFile; fileRef = 111E75821A815CFB00718AD7 /* NYPLSettingsPrimaryTableViewController.m */; };
```
ErnestFan, line 514: Is there a reason why we remove the reference but not removing the file? or this is a clean up?

ettore: that might be related to removing it from the OE target, since that uses the swift version. But yeah the reference is still there for SimplyE.

ErnestFan (approved):

PR #1188, opened 2020-09-15
Merge DPLA key update code for signature verification

vladimirfedorov:
**What's this do?**
- requests DRM public key from available vendors when the app starts;
- stores expiration date based on Cache-Control time for each vendor;
- checks the date when an audiobook opens and updates the key if necessary.

To support rotation of keys, https://jira.nypl.org/browse/SIMPLY-2944

Clean up simulator data, run the app and open an audiobook from "DPLA Exchange Audiobook Test" library.

No

Added the new code documentation

@vladimirfedorov


ettore (changes requested):
some merge issues, scary git didn't catch these.

couple more things:
- can you bump the version to 3.6.1?
- can you make sure the new Open eBooks target also builds? Most likely all these new files should be added to that target too.

Simplified/NYPLErrorLogger.swift
```diff
@@ -132,6 +136,66 @@ fileprivate let nullString = "null"
     }
   }
 
+  /// Broad areas providing some kind of operating context for error reporting.
+  /// These are meant to be related to the code base more than functionality,
+  /// (e.g. an error related to audiobooks may happen in different classes)
+  /// although the two things may obviously overlap.
+  enum Context: String {
```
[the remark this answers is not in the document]
vladimirfedorov: git is fine, it reported conflicts, I added the files because I saw it as files are missing and it's easier to remove from Xcode than edit Xcode project file manually. made it draft because fo this to test locally first

Simplified/NYPLKeychainManager.swift
```diff
@@ -172,38 +174,52 @@ import Foundation
...
+        errMsg = "errSecUserCanceled"
+      case errSecBadReq:
+        errMsg = "errSecBadReq"
+      default:
+        errMsg = "Unknown OSStatus: \(status)"
       }
     }
+    
+    NYPLErrorLogger.logError(withCode: .keychainItemAddFail,
+                             context: "\(NYPLErrorLogger.Context.keychainManagement.rawValue) \(vendor)",
```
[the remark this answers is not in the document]
vladimirfedorov: right, now it reports the error! changed to the newer declaration

Simplified/NYPLErrorLogger.swift
```diff
@@ -132,6 +136,66 @@ fileprivate let nullString = "null"
...
+  }
+
+  // MARK:- Generic helpers
+
+  /**
+   Helper method for other logging functions that adds relevant account info
+   to our crash reporting system.
+   - parameter metadata: report metadata dictionary
+   */
+  private class func addAccountInfoToMetadata(_ metadata: inout [String: Any]) {
```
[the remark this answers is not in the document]
vladimirfedorov: there were two declarations of the function, removed the upper one

vladimirfedorov:
This pull request is ready for review. Version 3.6.1, build 2. Added files and built Open eBooks target.

Simplified/NYPLErrorLogger.swift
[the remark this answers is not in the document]
vladimirfedorov: That part is fixed now, I took the code from `develop` and changed one call that was using this Context enum

Simplified/NYPLKeychainManager.swift
```diff
@@ -172,38 +174,52 @@ import Foundation
...
+        errMsg = "errSecUserCanceled"
+      case errSecBadReq:
+        errMsg = "errSecBadReq"
+      default:
+        errMsg = "Unknown OSStatus: \(status)"
       }
     }
+    
+    NYPLErrorLogger.logError(withCode: .keychainItemAddFail,
+                             summary: "Keychan \(errMsg) error for vendor \(vendor)",
```
ettore, line 221: can you remove the `errMsg` from the summary?  The reason is that the summary is used as the title in Crashlytics, so we don't want to have a huge proliferation of error types there. 

It's a difficult compromise, but after browsing those reports a lot I am trying to be more precise in how we log events, they easily become hard to browse.

vladimirfedorov: sure, done

Simplified/NYPLKeychainManager.swift
```diff
@@ -35,6 +36,7 @@ import Foundation
     #endif
     updateKeychainForBackgroundFetch()
     manageFeedbooksData()
+    manageFeedbookDrmPrivateKey()
```
ettore, line 39: just a note for the future: this code is run during `application:didFinishLaunchingWithOptions:` which has to finish as quickly as possible, so we should start thinking at different ways to do this kind of initialization instead of tagging it inside that method. This is manipulating the keychain, so I am just wondering if there's something to worry about here.

vladimirfedorov: I understand, inside the code checks vendor's key date (date is stored in UserDefaults) and if no date or date < today, creates URLSession.shared.dataTask, which are running in a separate queue, so this functions quits quickly. No keychain manipulation at that point.

Simplified/NYPLBookCellDelegate.m
```diff
@@ -175,70 +175,79 @@ - (void)openAudiobook:(NYPLBook *)book {
...
+    [NSOperationQueue.mainQueue addOperationWithBlock:^{
+      id<Audiobook> const audiobook = [AudiobookFactory audiobook: dict ?: json];
+
+      if (!audiobook) {
+        if (error) {
+          [self presentDRMKeyError:error];
+        } else {
+          [self presentUnsupportedItemError];
+        }
+        return;
```
ettore, line 188: do we need to log an error event to Crashlytics for both cases here? Not sure how hard it is to diagnose these errors (I know you're logging with `Log` but those are buried inside individual events whenever they are reported)

ettore (approved):

ErnestFan (approved):

Simplified/NYPLBookCellDelegate.m
vladimirfedorov: `presentUnsupportedItemError` doesn't log to Crashalytics. Do you think we need to log all unsupported items to Crashalytics? Maybe we can create a separate ticket for that kind of issues, it is called for unsupported pdfs and any unknown book object.

PR #1189, opened 2020-09-16
SIMPLY-3065 Fix Auto Layout warning when displaying barcode

ErnestFan:
**What's this do?**
Lower the priority of one of the height constraints, so the system generated constraint does not break.
The behaviour and size of the components remain unchanged after lowering the priority.
[Reference](https://stackoverflow.com/questions/23308400/auto-layout-what-creates-constraints-named-uiview-encapsulated-layout-width-h/43070239)

[SIMPLY-3065](https://jira.nypl.org/browse/SIMPLY-3065?filter=-1)

Test in Xcode, there should be no warnings when showing and hiding the barcode image

N/A

N/A

@ErnestFan


Simplified/NYPLSettingsAccountDetailViewController.m
```diff
@@ -1520,7 +1520,9 @@ - (UITableViewCell *)tableView:(__attribute__((unused)) UITableView *)tableView
           }];
           [self.barcodeImageView autoPinEdgeToSuperviewEdge:ALEdgeTop withInset:sConstantSpacing];
           [self.barcodeImageLabel autoAlignAxisToSuperviewAxis:ALAxisVertical];
-          [self.barcodeImageLabel autoPinEdgeToSuperviewEdge:ALEdgeBottom withInset:10.0];
+          [NSLayoutConstraint autoSetPriority:UILayoutPriorityDefaultHigh forConstraints:^{
```
ettore, line 1523: did you verify this on a physical device too? If I remember correctly, on a physical device there was a glitchy line being displayed after hiding the barcode.

ErnestFan: It works fine on a physical iPhone xs (iOS 13.7)

ettore (approved):

PR #1194, opened 2020-09-22
Merge incomplete Readium2 feature (disabled)

ettore:
**What's this do?**
Merges the R2 branch into develop. The feature is currently disabled, so there should be no changes beside the ones related to the position of the reader settings menu, which has already been approved by Product.

Maintaining the feature/readium2 branch as become costly because of the constant merge conflicts. Since the feature can be turned off in one point, merging this branch down should be harmless.

This code has already been reviewed, so perhaps things to watch out in this code review are things like missing files, merge oddities, wrong configurations, and the integration points.

There's not much to test since the feature is disabled.

n/a

yes (from the branch directly)

@ettore


vladimirfedorov (approved):

vladimirfedorov:
Do I understand correctly to enable Readium2 back we need to change Simplified/NYPLBookCellDelegate.m only? All R2 frameworks are in the project, just checking as I need R2 for LCP and upcoming Content Protection

ettore:
> Do I understand correctly to enable Readium2 back we need to change Simplified/NYPLBookCellDelegate.m only? All R2 frameworks are in the project, just checking as I need R2 for LCP and upcoming Content Protection

yes, that's correct. The frameworks are all in, but they will be "disabled" unless that code in NYPLBookCellDelegate.m is flipped. I also filed https://jira.nypl.org/browse/SIMPLY-3102 for enabling that via a Developer menu toggle.

ettore:
on second thought, I think we should wait on merging this. I forgot that there are changes in the way reader settings are being handled under the hood that affect R1 as well, and that's completely untested.  Given other upcoming deadlines (Open eBooks), I don't think it's smart to risk this merge.

ErnestFan (approved):

PR #1202, opened 2020-10-01
SIMPLY-3148 Refactor links on OE Settings tab so that they are read from Auth doc

ettore:
**What's this do?**
- Reads EULA, privacy policy and acknowledgements links from current Account info (i.e. from authentication document) instead of using hard coded URLs. Hard coded URLs are still in place as a fallback. This is done for OE.
- In the class accessing the account's URLs (NYPLLibraryAccountURLsProvider), instead of adding an implicit dependency to the AccountsManager, I injected the dependency via a protocol. This is just a small step toward breaking from the pattern of having every class tightly coupled with the AccountsManager, which is a huge singleton.
- incremental project reorg moving some files (no changes) into more semantics-based directories.

https://jira.nypl.org/browse/SIMPLY-3148

test that the items in Settings tab do not lead to a 404, in both signed in and signed out cases.

not breaking for SE.

yes

@ettore


Simplified/NYPLRootTabBarController.m
```diff
@@ -50,7 +50,9 @@ - (instancetype)init
   self.catalogNavigationController = [[NYPLCatalogNavigationController alloc] init];
   self.myBooksNavigationController = [[NYPLMyBooksNavigationController alloc] init];
   self.holdsNavigationController = [[NYPLHoldsNavigationController alloc] init];
-  self.settingsSplitViewController = [[NYPLSettingsSplitViewController alloc] init];
+  self.settingsSplitViewController = [[NYPLSettingsSplitViewController alloc]
+                                      initWithCurrentLibraryAccountProvider:
+                                      AccountsManager.shared];
```
ettore, line 55: @ErnestFan we talked about this some time ago. As you can see here, the dependency on `AccountsManager` that the `NYPLSettingsSplitViewController` would have (in order to get the current Account's URLs) is now explicit, i.e. expressed outside of the class. However the NYPLSettingsSplitViewController interface only sees a protocol (NYPLCurrentLibraryAccountProvider). 

If instead we hard-coded the usage of `AccountsManager` inside the `NYPLSettingsSplitViewController` implementation (as often do everywhere else in the code base!), the 2 classes would be very tightly coupled: you could not even compile the SplitVC without the AccountsManager. This dependency would be implicit, in the sense that by looking at only the interface of `NYPLSettingsSplitViewController` you could not know that its implementation depends on `AccountsManager`.

By breaking this coupling via a protocol the SplitVC doesn't know about the `AccountsManager`. It only knows about the `NYPLCurrentLibraryAccountProvider` protocol (implemented by `AccountsManager`). This is the key point. This would allow us to write unit tests for SplitVC without needing to set up a huge class like `AccountsManager`.

This case here is not the best example (unit testing SplitVC might be good, sure, but there are other areas that need testing more urgently), but it's an example of how I would like us to move forward in the future, especially when dealing with singletons.

ErnestFan (approved):

PR #1204, opened 2020-10-02
SIMPLY-3018 Upgrade to Firebase Crashlytics (from Fabric CL)

ettore:
**What's this do?**
1. Upgrades to Firebase Crashlytics from Fabric Crashlytics, which will stop reporting events starting November 15 2020. See commit message for the process followed.
2. Reorganize supporting files since there are now more differences between SimplyE and Open eBooks, and these different files were scattered around the file system. Now they are easily identifiable in the `SimplyE` and `OpenEbooks` directories at the root level. This includes image assets, plists, crashlytics stuff. 
3. I also collected all scripts into one `scripts` directory for similar reasons. Adjusted scripts and README accordingly.

https://jira.nypl.org/browse/SIMPLY-3018

No QA is needed: I verified that Fabric's frameworks are no longer in the build, and that the version we use is the same that we'd be using if we used the officially supported Cocoapods instead of Carthage (whose support is still experimental). I also verified that events are being reported To CL and that all targets of the project build.

none, this takes care of SimplyE and OE.

yes

@ettore


scripts/build-3rd-parties-dependencies.sh
```diff
@@ -2,22 +2,24 @@
...
+#     ./scripts/build-3rd-parties-dependencies.sh
 #
 # Description: This scripts wipes your Carthage folder, checks out and rebuilds
 #              all Carthage dependencies. It also rebuilds OpenSSL and cURL
 #              from scratch.
 
+echo "Building 3rd party dependencies..."
+
 # update dependencies from Certificates repo
 ./update-certificates.sh
```
ErnestFan, line 14: Shouldn't this line has the `/scripts/` as well?

scripts/build-3rd-parties-dependencies.sh
```diff
@@ -2,22 +2,24 @@
...
 
 # rebuild all Carthage dependencies from scratch
-./build-carthage.sh
+./scripts/build-carthage.sh
 
 # this is required for the Adobe SDK
 ./build-openssl-curl.sh
```
ErnestFan, line 20: same as above

scripts/build-3rd-parties-dependencies.sh
ettore: whoops! you're right, not sure how I missed this.

scripts/build-3rd-parties-dependencies.sh
ettore: nice catch!

ErnestFan (approved):

scripts/build-3rd-parties-dependencies.sh
ettore: fixed both

PR #1205, opened 2020-10-02
SIMPLY-3153 Added support for books with content-type of application/octet-stream

vladimirfedorov:
**What's this do?**
Adds support for books with `Content-type` of `application/octet-stream`.

Columbia University Libraries - Demo books fail to download because of that unsupported content type.
https://jira.nypl.org/browse/SIMPLY-3153

Open Columbia University Libraries - Demo library and download "Mathematical Theory of Advanced Computing"

No

No

@vladimirfedorov

### DIFF
--- a/Simplified/NYPLBookAcquisitionPath.h
+++ b/Simplified/NYPLBookAcquisitionPath.h
@@ -12,6 +12,7 @@ static NSString * const _Nonnull ContentTypeFindaway = @"application/vnd.library
 static NSString * const _Nonnull ContentTypeOpenAccessAudiobook = @"application/audiobook+json";
 static NSString * const _Nonnull ContentTypeOpenAccessPDF = @"application/pdf";
 static NSString * const _Nonnull ContentTypeFeedbooksAudiobook = @"application/audiobook+json;profile=\"http://www.feedbooks.com/audiobooks/access-restriction\"";
+static NSString * const _Nonnull ContentTypeOctetStream = @"application/octet-stream";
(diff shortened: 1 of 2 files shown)


ettore:
I don't see any problems with these changes, but did you verify with Leonard if this content-type is officially supported?

vladimirfedorov:
Yes, we have a thread in # mobile-development "Thread about Columbia University Libraries - Demo Content-type for books", Android and server scripts support that content-type value

Simplified/NYPLBookAcquisitionPath.h
```diff
@@ -12,6 +12,7 @@ static NSString * const _Nonnull ContentTypeFindaway = @"application/vnd.library
 static NSString * const _Nonnull ContentTypeOpenAccessAudiobook = @"application/audiobook+json";
 static NSString * const _Nonnull ContentTypeOpenAccessPDF = @"application/pdf";
 static NSString * const _Nonnull ContentTypeFeedbooksAudiobook = @"application/audiobook+json;profile=\"http://www.feedbooks.com/audiobooks/access-restriction\"";
+static NSString * const _Nonnull ContentTypeOctetStream = @"application/octet-stream";
```
ettore, line 15: can you move all these definitions to the `.m`?  It's wrong to have static variables in a header file.

vladimirfedorov: sure, done

ettore (approved):

PR #1218, opened 2020-10-13
SIMPLY-2510 Refactor account update logic into swift, fix a retain cycle + bug locking up sign-in modal UI

ettore:
**What's this do?**
3 things (see commits):
- refactors the duplicated logic for updating the user account after signing in. This was duplicated in Sign in VC modal and Sign In VC in Settings tab.
- Sign in VC modal had a retain cycle
- Both Sign In VCs have calls to completely block user interactions during sign in. This is dangerous and deprecated, especially because these calls are not balanced! So I removed most of them because they are more harmful than useful, except one were it's easy to see the balancing.

this refactor is needed for facilitating https://jira.nypl.org/browse/SIMPLY-3092
but the real ticket is https://jira.nypl.org/browse/SIMPLY-2510

There are no new features here, so a regression of all the sign-in flows from both settings tab and sign-in modal will address these changes. 

n/a

yes

@ettore

### DIFF
--- a/Simplified/Accounts/Library/AccountsManager.swift
+++ b/Simplified/Accounts/Library/AccountsManager.swift
@@ -6,9 +6,15 @@ let currentAccountIdentifierKey  = "NYPLCurrentAccountIdentifier"
   var currentAccount: Account? {get}
 }
 
+@objc protocol NYPLLibraryAccountsProvider: NYPLCurrentLibraryAccountProvider {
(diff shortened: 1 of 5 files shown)


ettore:
I forgot to add a unit test for the new `updateUserAccount...` method on NYPLSignInBusinessLogic but I will in the next PR. It may have been because there is still some decoupling to do (I will do that too in the next PR).

PR #1219, opened 2020-10-13
Full list of code changes going into SimplyE 3.6.1

ettore:
**What's this do?**
This is a list of ALL the changes that went into 3.6.1. It looks like a lot but the appearance is deceiving. Let's re-review them and see if we spot some problems we didn't see before.

To summarize the changes:
- upgraded Crashlytics to Firebase Crashlytics: this was required because the old Crashlytics will stop working on November 15 2020, and we need some leeway time before that so that we can get bug reports consistently. Minor code changes.
- upgraded AudioEngine from 6.1.15 to 6.5.2: this was addressing a crash and should provide better stability to audiobooks playback. Findaway addressed a ton of bugs. A list of changes on their end is here (requires authentication: https://developer.audioengine.io/downloads )
- fix crashes introduced in 3.6.0
- refactors for Open eBooks: for SimplyE is just moving code around. This is done with separate SimplyE  / Open eBooks extensions in separate files ('file+SE' / 'file+OE') compiled for one app but not for the other. So ignore any file ending in `+OE.swift` or starting as `OExyz.swift`.
- refactor switch Library code which was duplicated verbatim 3 times.
- **added signature verification for DPLA audiobooks**
- **[added support for eBooks of type `application/octet-stream`](https://github.com/NYPL-Simplified/Simplified-iOS/commit/9630836107bffebc0286d90fa764c0fb4524a5a6).**
- **[fix Overdrive audiobooks download issues introduced in 3.6.0](https://github.com/NYPL-Simplified/Simplified-iOS/commit/dd50e09ccf04e176ed982735aa52a2a3e4bd4045).** 

This is a follow up to the discussions about QA we had with Risa and others today.

Let's re-review all these changes. This is overwhelming to do in one sitting so let's do it in 1 hour-max chunks. Use the "Viewed" checkbox GitHub offers to keep track of what you already reviewed. If you don't understand something, just ask. 

Note: the last 3 items in the list above in bold are fixes that affect downloads directly. Since this is one of the primary functions of SimplyE we should pay attention to those commits in particular. Look not just at the changed code, but also at what it might potentially affect, such as related classes, etc.

Not ready to merge now since we have not shipped yet. 

yes

we all did :)


Simplified/WelcomeScreen/NYPLEULAViewController.swift
```diff
@@ -0,0 +1,171 @@
+import WebKit
+
+class NYPLEULAViewController : UIViewController {
```
ettore, line 3: this is currently used only by Open eBooks. You can ignore it.

Simplified/AppInfrastructure/NYPLLibraryNavigationController.m
```diff
@@ -11,70 +22,27 @@
...
 
-#import "SimplyE-Swift.h"
 
-#if defined(FEATURE_DRM_CONNECTOR)
-#import <ADEPT/ADEPT.h>
-#endif
+@interface NYPLLibraryNavigationController ()
 
-@implementation NYPLMyBooksNavigationController
+@end
```
ettore, line 28: it is unfortunate Github shows this as a rename. It is not. NYPLLibraryNavigationController is a new superclass for NYPLCatalogNavigationController, NYPLMyBooksNavigationController, NYPLHoldsNavigationController.

Simplified/AppInfrastructure/NYPLLibraryNavigationController.m
```diff
@@ -11,70 +22,27 @@
...
-    dispatch_async(dispatch_get_main_queue(), ^{
-      [self popToRootViewControllerAnimated:NO];
-    });
-  } else {
-    [self popToRootViewControllerAnimated:NO];
-  }
-}
-
-#ifdef SIMPLYE
-- (void) switchLibrary
```
ettore, line 75: this entire function was removed from NYPLCatalogNavigationController, NYPLMyBooksNavigationController, NYPLHoldsNavigationController and moved it into their new superclass, NYPLLibraryNavigationController.

Simplified/Accounts/Library/NYPLLibraryAccountURLsProvider.swift
```diff
@@ -0,0 +1,37 @@
...
+//  NYPLLibraryAccountURLsProvider.swift
+//  Open eBooks
+//
+//  Created by Ettore Pasquini on 9/30/20.
+//  Copyright © 2020 NYPL Labs. All rights reserved.
+//
+
+import Foundation
+
+class NYPLLibraryAccountURLsProvider {
```
ettore, line 11: currently only used by Open eBooks

ettore:
@vladimirfedorov @ErnestFan I just removed the flag since just released. I had not found any issues in this code after re-reviewing it.

ErnestFan (approved):

PR #1221, opened 2020-10-15
SIMPLY-3092 Fixes for Clever authentication in Open eBooks

ettore:
**What's this do?**
- Add new entitlement only for OE because we are in the process of shipping a new release and we should not change how SE is set up right now.
- Add NYPLIsSigningIn notification to address a bug in OE where the login choice picker would show up after redirection from Clever.
- Avoid tight coupling in NYPLSignInBusinessLogic via NYPLUniversalLinksSettings protocol.
- Add sign-in VC convenience initializer based on OE login choice, using refactor done recently in PR#1217.

https://jira.nypl.org/browse/SIMPLY-3092

I used the sandbox school district  from here: https://github.com/NYPL-Simplified/operations/wiki/TestCardsOpenEbooks#clever
You can then sign in by typing in one of those student-IDs for both user id and password.
If you have a real school district to try this on it would be even better, of course.

not affecting SimplyE.

yes

@ettore

### DIFF
--- a/OpenEbooks/OpenEbooks.entitlements
+++ b/OpenEbooks/OpenEbooks.entitlements
@@ -1,0 +1,14 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
+<plist version="1.0">
+<dict>
+	<key>com.apple.developer.associated-domains</key>
+	<array>
+		<string>applinks:dev-librarysimplified.pantheonsite.io</string>
+	</array>
+	<key>keychain-access-groups</key>
+	<array>
+		<string>7262U6ST2R.org.nypl.labs.SharedKeychainGroup</string>
+	</array>
+</dict>
+</plist>
--- a/Simplified/NYPLSettingsAccountDetailViewController.m
+++ b/Simplified/NYPLSettingsAccountDetailViewController.m
@@ -159,6 +159,7 @@ Authenticating with any of those barcodes should work.
   self.businessLogic = [[NYPLSignInBusinessLogic alloc]
                         initWithLibraryAccountID:libraryUUID
                         libraryAccountsProvider:AccountsManager.shared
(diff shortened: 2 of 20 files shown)


OpenEbooks/OpenEbooks.entitlements
```diff
@@ -0,0 +1,14 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
+<plist version="1.0">
+<dict>
+	<key>com.apple.developer.associated-domains</key>
+	<array>
+		<string>applinks:dev-librarysimplified.pantheonsite.io</string>
```
ettore, line 7: note: this is a temporary site. It will change once I get the production URL, which is still not ready at the moment.

ErnestFan (approved):

vladimirfedorov (approved):

Simplified/NYPLSettingsAccountDetailViewController.m
```diff
@@ -632,16 +637,21 @@ - (void)performLogOut
...
-                        currentBarcode.md5String];
-       [NYPLErrorLogger logNetworkError:error
-                                   code:NYPLErrorCodeApiCall
-                                summary:@"signOut"
-                                request:request
-                               response:response
-                                message:msg
-                               metadata:nil];
-       [self showLogoutAlertWithError:error responseCode:statusCode];
+       if (!self.businessLogic.selectedAuthentication.isOauth) {
```
ettore, line 640: note: I realize now that this `if` is probably wrong. I added it to avoid the error message to show up for OAuth sign-outs thinking it would just apply to barcode sign-outs. But I think this whole sign-out logic is actually wrong: the sign-out request is identical to the sign-in request. So before making any more changes I filed SIMPLY-3208 to fix the sign-out call.

ettore: for the mean time this `if` doesn't change much, so I'll leave it in.

PR #1223, opened 2020-10-16
SIMPLY-2820 Fix R2 bookmark create bug

ErnestFan:
**What's this do?**
Fixing a bug of bookmark create failed in R2 reader
Also fix the double "Ok" on the fail alert
Note:
There is a bug navigating to the bookmark in R2 reader.
The readium team has located this issue and has a PR opened for it. [PR link](https://github.com/readium/r2-navigator-swift/pull/142)
We need to upgrade our r2-navigator once they have this issue fixed.

[SIMPLY-2820](https://jira.nypl.org/browse/SIMPLY-2820)

Open a book that never been open in R1, create a bookmark and check if it exist in the bookmark list

N/A

N/A

@ErnestFan

### DIFF
--- a/Simplified/Reader2/UI/NYPLBaseReaderViewController.swift
+++ b/Simplified/Reader2/UI/NYPLBaseReaderViewController.swift
@@ -252,10 +252,6 @@ class NYPLBaseReaderViewController: UIViewController, NYPLBackgroundWorkOwner, L
     guard let bookmark = bookmarksBusinessLogic.addBookmark(location) else {
       let alert = NYPLAlertUtils.alert(title: "Bookmarking Error",
                                        message: "A bookmark could not be created on the current page.")
-      let action = UIAlertAction(title: NSLocalizedString("OK", comment: ""),
-                                 style: .default,
-                                 handler: nil)
-      alert.addAction(action)
       NYPLAlertUtils.presentFromViewControllerOrNil(alertController: alert,
                                                     viewController: self,
                                                     animated: true,
--- a/Simplified/NYPLReadiumBookmark.swift
+++ b/Simplified/NYPLReadiumBookmark.swift
@@ -6,9 +6,9 @@
   var chapter:String?
   var page:String?
 
-  var location:String
+  var location:String?
   var idref:String
-  var contentCFI:String
+  var contentCFI:String?
   var progressWithinChapter:Float = 0.0
   var progressWithinBook:Float = 0.0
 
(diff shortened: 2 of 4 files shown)


Simplified/Reader2/UI/NYPLBaseReaderViewController.swift
```diff
@@ -252,10 +252,6 @@ class NYPLBaseReaderViewController: UIViewController, NYPLBackgroundWorkOwner, L
     guard let bookmark = bookmarksBusinessLogic.addBookmark(location) else {
       let alert = NYPLAlertUtils.alert(title: "Bookmarking Error",
                                        message: "A bookmark could not be created on the current page.")
-      let action = UIAlertAction(title: NSLocalizedString("OK", comment: ""),
```
ettore, line 255: oh is this just unneeded since it's the default?

ettore (approved):
nice catch!

ettore:
one more note, it seems like R2 team is merging those navigator fixes to their Alpha branch, does that mean that we should track that branch now instead of `develop`, @ErnestFan?

Simplified/Reader2/UI/NYPLBaseReaderViewController.swift
ErnestFan: Yeah, `NYPLAlertUtils` has a default `Ok` button

vladimirfedorov (approved):

ErnestFan:
@ettore switching r2-navigator to `alpha` branch requires r2-streamer and r2-shared to be on `alpha`, and we will need to adopt the new Streamer API in our app.
The other way would be cherry-pick the fix in a fork.
I will work on the rest of the Bookmark feature, and we can decide which works the best for us by the time the Bookmark feature is nearly finish

ettore:
> @ettore switching r2-navigator to `alpha` branch requires r2-streamer and r2-shared to be on `alpha`, and we will need to adopt the new Streamer API in our app.
> The other way would be cherry-pick the fix in a fork.
> I will work on the rest of the Bookmark feature, and we can decide which works the best for us by the time the Bookmark feature is nearly finish

We already have forks for r2-streamer and r2-shared, so I'd rather avoid a new fork unless absolutely necessary. I think what we should do here is updating our r2-streamer / r2-shared forks with the new alpha changes. Eventually when they ship the new version we'll have to do that anyway. @vladimirfedorov do you agree?

PR #1229, opened 2020-10-21
Enable R2 by default

ettore:
**What's this do?**
Enables R2 by default on feature branch.

it makes no sense it is disabled on the feature branch.

n/a

n/a

n/a

@ettore


vladimirfedorov (approved):
I like it

PR #1232, opened 2020-10-23
SIMPLY-2804 R2 Bookmark syncing

ErnestFan:
**What's this do?**
Add, delete and sync bookmarks from server for R2
Probably easier to review by single commit

[SIMPLY-2804](https://jira.nypl.org/projects/SIMPLY/issues/SIMPLY-2804)

Make sure `bookmark syncing` is turned on in account detail
Open any epub book and create a bookmark
Go to bookmark list, the bookmark should be shown
Pull down to refresh and look for `server bookmark count` in debug area, which should be the same as the number of bookmarks in the list
*Since the bookmarks in the list could be local cache, this is the way to test the syncing without clearing the cache*
Delete a bookmark and repeat the step above to make sure bookmark on server is removed
For QA, I will have to talk to Joe to figure out what's their method on testing server bookmark

N/A

N/A

@ErnestFan

### DIFF
--- a/Simplified/NYPLBookRegistry.h
+++ b/Simplified/NYPLBookRegistry.h
@@ -13,7 +13,25 @@ static NSString *const _Nonnull NYPLBookRegistryDidChangeNotification =
 static NSString *const _Nonnull NYPLBookProcessingDidChangeNotification =
   @"NYPLBookProcessingDidChangeNotification";
 
-@interface NYPLBookRegistry : NSObject
+@protocol NYPLBookRegistryProvider <NSObject>
+
+- (nonnull NSArray<NYPLReadiumBookmark *> *)readiumBookmarksForIdentifier:(nonnull NSString *)identifier;
+
+- (nullable NYPLBookLocation *)locationForIdentifier:(nonnull NSString *)identifier;
+
+- (void)addReadiumBookmark:(nonnull NYPLReadiumBookmark *)bookmark
+             forIdentifier:(nonnull NSString *)identifier;
+  
+- (void)deleteReadiumBookmark:(nonnull NYPLReadiumBookmark *)bookmark
+                forIdentifier:(nonnull NSString *)identifier;
+
+- (void)replaceBookmark:(nonnull NYPLReadiumBookmark *)oldBookmark
+                   with:(nonnull NYPLReadiumBookmark *)newBookmark
+          forIdentifier:(nonnull NSString *)identifier;
+
+@end
+
+@interface NYPLBookRegistry : NSObject <NYPLBookRegistryProvider>
 
 // Returns all registered books.
 @property (atomic, readonly, nonnull) NSArray *allBooks;
--- a/Simplified/NYPLReadiumBookmark.swift
+++ b/Simplified/NYPLReadiumBookmark.swift
@@ -47,6 +47,9 @@
     self.idref = idref
     self.chapter = chapter ?? ""
     self.page = page ?? ""
+    // This location structure originally comes from R1 Reader's Javascript
+    // and its not available in R2, we are mimicking the structure
+    // in order to pass the needed information to the server
     self.location = location ?? "{\"idref\":\"\(idref)\",\"contentCFI\":\"\(contentCFI ?? "")\"}"
     self.progressWithinChapter = progressWithinChapter
     self.progressWithinBook = progressWithinBook
@@ -108,11 +111,13 @@
         return self.idref == other.idref
             && self.contentCFI == other.contentCFI
             && self.location == other.location
+            && self.chapter == other.chapter
     } else {
       // R2
       return self.idref == other.idref
         && self.progressWithinBook == other.progressWithinBook
         && self.progressWithinChapter == other.progressWithinChapter
+        && self.chapter == other.chapter
     }
   }
 }
--- a/Simplified/NYPLReadiumBookmark.swift
+++ b/Simplified/NYPLReadiumBookmark.swift
@@ -8,7 +8,7 @@
 
   // location and contentCFI are location information generated from R1 reader
   // which are not available in R2, therefore they are now optionals
-  var location:String?
+  var location:String
   var idref:String
   var contentCFI:String?
   var progressWithinChapter:Float = 0.0
@@ -47,7 +47,7 @@
     self.idref = idref
     self.chapter = chapter ?? ""
     self.page = page ?? ""
-    self.location = location
+    self.location = location ?? "{\"idref\":\"\(idref)\",\"contentCFI\":\"\(contentCFI ?? "")\"}"
     self.progressWithinChapter = progressWithinChapter
     self.progressWithinBook = progressWithinBook
     self.time = time ?? NSDate().rfc3339String()
@@ -90,7 +90,7 @@
             "idref":self.idref,
             "chapter":self.chapter ?? "",
             "page":self.page ?? "",
-            "location":self.location ?? "",
+            "location":self.location,
             "time":self.time,
             "device":self.device ?? "",
             "progressWithinChapter":self.progressWithinChapter,
@@ -101,14 +101,19 @@
   override func isEqual(_ object: Any?) -> Bool {
     let other = object as! NYPLReadiumBookmark
 
-    if (self.contentCFI == other.contentCFI &&
-      self.idref == other.idref &&
-      self.chapter == other.chapter &&
-      self.location == other.location)
-    {
-      return true
+    if let contentCFI = self.contentCFI,
+      let otherContentCFI = other.contentCFI,
+      contentCFI.count > 0 && otherContentCFI.count > 0 {
+        // R1
+        return self.idref == other.idref
+            && self.contentCFI == other.contentCFI
+            && self.location == other.location
+    } else {
+      // R2
+      return self.idref == other.idref
+        && self.progressWithinBook == other.progressWithinBook
+        && self.progressWithinChapter == other.progressWithinChapter
     }
-    return false
   }
 }
 
--- a/Simplified/Reader2/NYPLReaderBookmarksBusinessLogic.swift
+++ b/Simplified/Reader2/NYPLReaderBookmarksBusinessLogic.swift
@@ -198,9 +198,20 @@ class NYPLReaderBookmarksBusinessLogic: NSObject, NYPLReadiumViewSyncManagerDele
   private func didDeleteBookmark(_ bookmark: NYPLReadiumBookmark) {
     bookRegistry.delete(bookmark, forIdentifier: book.identifier)
 
-    // TODO: SIMPLY-2804 (syncing)
-    // see NYPLReaderReadiumView::deleteBookmark
-    // delete bookmark on server
+    guard let currentAccount = AccountsManager.shared.currentAccount,
+        let details = currentAccount.details,
+        let annotationId = bookmark.annotationId else {
+      Log.debug(#file, "Delete on Server skipped: Sync is not enabled or Annotation ID did not exist for bookmark.")
+      return
+    }
+    
+    if details.syncPermissionGranted && annotationId.count > 0 {
+      NYPLAnnotations.deleteBookmark(annotationId: annotationId) { (success) in
+        Log.debug(#file, success ?
+          "Bookmark successfully deleted" :
+          "Failed to delete bookmark from server. Will attempt again on next Sync")
+      }
+    }
   }
 
   var noBookmarksText: String {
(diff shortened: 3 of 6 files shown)


Simplified/Reader2/NYPLReaderBookmarksBusinessLogic.swift
```diff
@@ -200,25 +227,119 @@ class NYPLReaderBookmarksBusinessLogic: NSObject, NYPLReadiumViewSyncManagerDele
...
+        if let indexToRemove = serverBookmarksToKeep.index(of: serverBookmark) {
+          serverBookmarksToKeep.remove(at: indexToRemove)
+        }
+      }
+    }
+    
+    for localBookmark in localBookmarks {
+      if !localBookmarksToKeep.contains(localBookmark) {
+        bookRegistry.delete(localBookmark, forIdentifier: self.book.identifier)
+        localBookmarksToDelete.append(localBookmark)
```
ettore, line 315: do we actually use `localBookmarksToDelete` anywhere? Just wondering if we can remove it

Simplified/Reader2/NYPLReaderBookmarksBusinessLogic.swift
```diff
@@ -200,25 +227,119 @@ class NYPLReaderBookmarksBusinessLogic: NSObject, NYPLReadiumViewSyncManagerDele
...
+        }
+      }
+    }
+        
+    for bookmark in bookmarksToAdd {
+      bookRegistry.add(bookmark, forIdentifier: self.book.identifier)
+    }
+    
+    if serverBookmarksToDelete.count > 0 {
+      NYPLAnnotations.deleteBookmarks(serverBookmarksToDelete)
```
ettore, line 335: should this call be moved out of this function, since this function is about updating the local bookmarks?

vladimirfedorov (approved):

Simplified/Reader2/NYPLReaderBookmarksBusinessLogic.swift
```diff
@@ -200,25 +227,119 @@ class NYPLReaderBookmarksBusinessLogic: NSObject, NYPLReadiumViewSyncManagerDele
...
+    for serverBookmark in serverBookmarksToKeep {
+      for localBookmark in localBookmarksToKeep {
+        if serverBookmark.isEqual(localBookmark), let index = bookmarksToAdd.index(of: localBookmark) {
+          bookmarksToAdd.remove(at: index)
+        }
+      }
+    }
+        
+    for bookmark in bookmarksToAdd {
+      bookRegistry.add(bookmark, forIdentifier: self.book.identifier)
```
ettore, line 331: is this and the `delete` at line 314 the actual calls to update the local bookmarks?

Simplified/Reader2/NYPLReaderBookmarksBusinessLogic.swift
```diff
@@ -200,25 +227,119 @@ class NYPLReaderBookmarksBusinessLogic: NSObject, NYPLReadiumViewSyncManagerDele
...
+            return
+          }
+            
+          Log.debug(#file, serverBookmarks.count == 0 ? "No server bookmarks" : "Server bookmarks count: \(serverBookmarks.count)")
+          
+          self.updateLocalBookmarks(serverBookmarks: serverBookmarks,
+                                     localBookmarks: localBookmarks,
+                                     bookmarksFailedToUpload: bookmarksFailedToUpload)
+          {
+            self.bookmarks = self.bookRegistry.readiumBookmarks(forIdentifier: self.book.identifier)
```
ettore, line 269: I think there's a retain cycle here: self -> updateLocalBookmarks -> completion -> self

ettore (changes requested):
I think overall looks good, nice job.

The main change I'd ask is, is it possible to add some unit tests for some of the logic?

Simplified/NYPLReadiumBookmark.swift
```diff
@@ -47,7 +47,7 @@
     self.idref = idref
     self.chapter = chapter ?? ""
     self.page = page ?? ""
-    self.location = location
+    self.location = location ?? "{\"idref\":\"\(idref)\",\"contentCFI\":\"\(contentCFI ?? "")\"}"
```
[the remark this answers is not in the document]
ErnestFan: For R1, we pass the bookmark's `location` to the server which contains the `idref` and `contentCFI`.
For R2, we don't have the `location` nor the `contentCFI`, we do have `idref` but it is not being passed to the server directly and we need to include it in the bookmark's `location`.
`idref` is responsible for comparing bookmarks.

Simplified/Reader2/NYPLReaderBookmarksBusinessLogic.swift
```diff
@@ -200,25 +227,119 @@ class NYPLReaderBookmarksBusinessLogic: NSObject, NYPLReadiumViewSyncManagerDele
...
-//        vc.bookmarksRefreshControl.endRefreshing()
-//        if !success {
-//          let alert = NYPLAlertUtils.alert(title: "Error Syncing Bookmarks",
-//                                           message: "There was an error syncing bookmarks to the server. Ensure your device is connected to the internet or try again later.")
-//          vc.present(alert, animated: true)
-//        }
-//      }
-//    }
+    
+  func syncBookmarks(completion: @escaping (Bool, [NYPLReadiumBookmark]) -> ()) {
```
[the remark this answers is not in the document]
ErnestFan: I believe with the check, we don't need to run all the code below if the upload fails because of reachability

Simplified/Reader2/NYPLReaderBookmarksBusinessLogic.swift
ErnestFan: Yes

Simplified/Reader2/NYPLReaderBookmarksBusinessLogic.swift
ErnestFan: Right, I will remove it

Simplified/Reader2/NYPLReaderBookmarksBusinessLogic.swift
```diff
@@ -180,11 +196,22 @@ class NYPLReaderBookmarksBusinessLogic: NSObject, NYPLReadiumViewSyncManagerDele
...
+      Log.debug(#file, "Delete on Server skipped: Sync is not enabled or Annotation ID did not exist for bookmark.")
+      return
+    }
+    
+    if details.syncPermissionGranted && annotationId.count > 0 {
+      NYPLAnnotations.deleteBookmark(annotationId: annotationId) { (success) in
+        Log.debug(#file, success ?
+          "Bookmark successfully deleted" :
+          "Failed to delete bookmark from server. Will attempt again on next Sync")
+      }
```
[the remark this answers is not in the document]
ErnestFan: `NYPLAnnotations.deleteBookmark(annotationId:completionHandler)` already logs to Crashlytics, this is just for us

Simplified/Reader2/NYPLReaderBookmarksBusinessLogic.swift
ErnestFan: The only place I can think of is the completion block in `syncBookmarks`, which would make it more complicated, what do you think? I removed the if-statement though since I don't think it's necessary

Simplified/NYPLBookRegistry.h
```diff
@@ -13,7 +13,25 @@ static NSString *const _Nonnull NYPLBookRegistryDidChangeNotification =
 static NSString *const _Nonnull NYPLBookProcessingDidChangeNotification =
   @"NYPLBookProcessingDidChangeNotification";
 
-@interface NYPLBookRegistry : NSObject
+@protocol NYPLBookRegistryProvider <NSObject>
+
+- (nonnull NSArray<NYPLReadiumBookmark *> *)readiumBookmarksForIdentifier:(nonnull NSString *)identifier;
```
ettore, line 18: Thanks for adding this protocol. One step forward toward less tight coupling 👍 

I think the return value for this and `locationForIdentifier:` should be nullable though... what If I pass in an invalid identifier?

Secondly, are all these apis implemented in Objc? if some of them are implemented in swift, nonnulls can lead to a crash if the argument is actually nil.

Simplified/NYPLReadiumBookmark.swift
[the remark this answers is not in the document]
ettore: can you add a comment in the code about where does this structure come from?

Simplified/NYPLReadiumBookmark.swift
```diff
@@ -101,14 +101,19 @@
...
-      self.location == other.location)
-    {
-      return true
+    if let contentCFI = self.contentCFI,
+      let otherContentCFI = other.contentCFI,
+      contentCFI.count > 0 && otherContentCFI.count > 0 {
+        // R1
+        return self.idref == other.idref
+            && self.contentCFI == other.contentCFI
+            && self.location == other.location
```
ettore, line 110: for R1 this test should also include `self.chapter == other.chaper`? (according to previous code)

actually maybe for R2 too? (not sure)

Simplified/Reader2/NYPLReaderBookmarksBusinessLogic.swift
```diff
@@ -143,15 +149,28 @@ class NYPLReaderBookmarksBusinessLogic: NSObject, NYPLReadiumViewSyncManagerDele
...
+      let accountDetails = currentAccount.details,
+      accountDetails.syncPermissionGranted else {
+        self.bookRegistryProvider.add(bookmark, forIdentifier: book.identifier)
+        return
+    }
+    
+    NYPLAnnotations.postBookmark(forBook: book.identifier,
+                                 toURL: nil,
+                                 bookmark: bookmark) { (serverAnnotationID) in
+      Log.debug(#function, serverAnnotationID != nil ? "Bookmark upload succeed" : "Bookmark failed to upload")
```
ettore, line 169: in case of failure, should we log an event with NYPLErrorLogger instead ? I don't think `postBookmark(forBook...` logs it. Or we could log it there, since I think you said it does do that for deletions.

Simplified/Reader2/NYPLReaderBookmarksBusinessLogic.swift
```diff
@@ -200,25 +230,113 @@ class NYPLReaderBookmarksBusinessLogic: NSObject, NYPLReadiumViewSyncManagerDele
...
+          
+          self.updateLocalBookmarks(serverBookmarks: serverBookmarks,
+                                     localBookmarks: localBookmarks,
+                                     bookmarksFailedToUpload: bookmarksFailedToUpload)
+          { [weak self] in
+            guard let self = self else {
+              completion(false, localBookmarks)
+              return
+            }
+            self.bookmarks = self.bookRegistryProvider.readiumBookmarks(forIdentifier: self.book.identifier)
```
ettore, line 276: since we always use the book registry to get the current bookmarks, do we actually need to have a `bookmarks` property here?

ettore (approved):
I'm going to approve this, but please let me know if you disagree with my comments.

Simplified/Reader2/NYPLReaderBookmarksBusinessLogic.swift
ErnestFan: The `bookmarks` only being updated when sync/add/delete happens, and the `BookRegistry` does not get updated until the network request is completed, so I think a local property is necessary

Simplified/Reader2/NYPLReaderBookmarksBusinessLogic.swift
ErnestFan: `postBookmark(forBook...` does log the event with `Log.error()` like the deletions. The actual network request happens in `postAnnotation(forBook...` and that's where we have access to the network error. Maybe another ticket/PR to update the whole `NYPLAnnotations` class to log with NYPLErrorLogger when possible?

Simplified/NYPLBookRegistry.h
ErnestFan: `readiumBookmarksForIdentifier:` returns an empty array if invalid identifier being passed in.

`locationForIdentifier:` is already nullable.

Simplified/Reader2/NYPLReaderBookmarksBusinessLogic.swift
ettore: sounds good, make sure to file the ticket. Keep in mind that logging with `Log.error` alone does pretty much nothing -- in the sense that that log will be buried in the next event that it is reported.

PR #1236, opened 2020-10-29
SIMPLY-2510 Refactor sign-in logic handling DRM authorization

ettore:
**What's this do?**
Note that the 2 objc implementations of `drmAuthorizeWithUsername:password:loggingContext:` where slightly different in the 2 VCs, with the modal implementation performing an additional check to warn the user if the DRM authorization takes too long. Since this seems like a good idea, this is now added to both (the VC dismissal will result in nothing for a non-modal VC in the stack).

The implementations of `drmAuthorizeUserData:loggingContext` were instead identical in the 2 VCs.

Both are now rewritten in swift and added to NYPLSignInBusinessLogic as an extension.

https://jira.nypl.org/browse/SIMPLY-2510

As other PRs for this ticket, no changes in functionality should be observed from what we currently have.

n/a

Yes, added docs for all the new methods.

@ettore


vladimirfedorov (approved):

ErnestFan (approved):

ettore:
Just did a rebase because there were conflicts.

PR #1239, opened 2020-10-30
SIMPLY-2510 Refactor business logic delegate methods

ettore:
**What's this do?**
Rewrite the delegate methods from the point of view of the business logic. Move some logging logic that was in the deleegate implementation to the business logic. Rewrite the URLSession challenge handler in swift to facilitate integration.

This way it was possible to clear out some more state from the VCs and simplify the api with less back'n'forth between business logic and VC.

NOTE: this PR looks kind of big, but a lot of changes are just indentation.

https://jira.nypl.org/browse/SIMPLY-2510

we should test out all different authentication methods from both the sign-in modal and the settings tab, for SimplyE and Open eBooks.

n/a

yes

@ettore


Simplified/NYPLSettingsAccountDetailViewController.m
```diff
@@ -701,39 +699,14 @@ - (void)alertUserOfValidationError:(NSError * const)error
...
-    problemDocument = [NYPLProblemDocument fromData:data
-                                              error:&problemDocumentParseError];
-    if (problemDocumentParseError == nil && problemDocument != nil) {
-      alert = [NYPLAlertUtils alertWithTitle:problemDocument.title
-                                     message:problemDocument.detail];
-    }
-  }
-
-  // error logging
-  if (problemDocumentParseError != nil) {
```
ettore, line 718: this logging logic has been moved to NYPLSignInBusinessLogic

ErnestFan (approved):

PR #1253, opened 2020-11-19
SIMPLY-3292 Re-authenticate if authentication expired

ettore:
**What's this do?**
Centralizes the logic to determine if an api call indicates that the authentication credentials have somehow expired for a currently signed-in user. Typically this can happen for Clever (and I think SAML) which is common in OE. In that case, the app will prompt the user to re-authenticate. I have not observed this happening ever for basic (barcode) authentication flows, although this should be able to handle that too.

The logic in question is `URLResponse+NYPLAuthentication.swift`. @leonardr I'd really appreciate it if you can take a look at that file and let me know if it makes sense to you, or if it's too generic. The idea is that we should be able to use it for any api call in SimplyE/OE (except card creator stuff I suppose).

https://jira.nypl.org/browse/SIMPLY-3292

1. sign in via Clever, and make sure to say that "this computer is shared" on the Clever pages in Safari.
2. leave the app alone for some time -- I think 1 day, although I am unclear on exactly how long it takes for the token to expire.
3. reopen the app. Usually it should automatically direct you to the Catalog (for example if you had force-quit the app, or if the system killed it in case it was strapped for resources)
4. at this point if the token was expired you should see a modal sign-in pane popup in Catalog, with Clever automatically selected. If you now tap on `Login` you should be directed to Safari to get a new token, and then come back to OE. This "going back" step sometimes fails with a 404, but that's outside of my control: if you see that, let me know how persistent it is. Otherwise, it should proceed as normal and display the Catalog.

n/a

yes

@ettore

### DIFF
--- a/Simplified.xcodeproj/project.pbxproj
+++ b/Simplified.xcodeproj/project.pbxproj
@@ -485,6 +485,12 @@
 		7347F11D245A4DE200558D7F /* GoogleService-Info.plist in Resources */ = {isa = PBXBuildFile; fileRef = 738EF2CB2405E38800F388FB /* GoogleService-Info.plist */; };
 		7347F133245A508400558D7F /* NYPLCardCreator.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = 7347F132245A508400558D7F /* NYPLCardCreator.framework */; };
 		7347F135245A50CA00558D7F /* NYPLCardCreator.framework in CopyFiles */ = {isa = PBXBuildFile; fileRef = 7347F132245A508400558D7F /* NYPLCardCreator.framework */; settings = {ATTRIBUTES = (CodeSignOnCopy, RemoveHeadersOnCopy, ); }; };
+		734B789125659611006FB8AD /* URLResponse+NYPLAuthentication.swift in Sources */ = {isa = PBXBuildFile; fileRef = 734B789025659611006FB8AD /* URLResponse+NYPLAuthentication.swift */; };
+		734B789225659611006FB8AD /* URLResponse+NYPLAuthentication.swift in Sources */ = {isa = PBXBuildFile; fileRef = 734B789025659611006FB8AD /* URLResponse+NYPLAuthentication.swift */; };
+		734B789325659611006FB8AD /* URLResponse+NYPLAuthentication.swift in Sources */ = {isa = PBXBuildFile; fileRef = 734B789025659611006FB8AD /* URLResponse+NYPLAuthentication.swift */; };
+		734B78992565F7DE006FB8AD /* NYPLReauthenticator.swift in Sources */ = {isa = PBXBuildFile; fileRef = 734B78982565F7DE006FB8AD /* NYPLReauthenticator.swift */; };
+		734B789A2565F7DE006FB8AD /* NYPLReauthenticator.swift in Sources */ = {isa = PBXBuildFile; fileRef = 734B78982565F7DE006FB8AD /* NYPLReauthenticator.swift */; };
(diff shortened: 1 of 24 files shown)


ErnestFan (approved):

vladimirfedorov (approved):

leonardr:
* The default lifetime for an OAuth token issued by the CM is 42 days, and Open eBooks hasn't overridden this. I could change it to 1 day on the QA server; that should make it easier to test this. Let me know if that will help.

Here's what I think is the case you're testing (from OAuthAuthenticationProvider.authenticated_patron):

```
        # This token wasn't in our database, or was expired. The
        # patron will have to log in through the OAuth provider again
        # to get a new token.
        return None
```

This in turn is handled by this code in BaseCirculationManagerController.authenticated_patron_from_request:

```
        if patron is None:
            # Credentials were provided but they turned out not
            # to identify anyone in particular.
            return self.authenticate()
```

And authenticate() sends a 401 response with the authentication document:

```
    def authenticate(self):
        """Sends a 401 response that demands authentication."""
        headers = self.manager.auth.create_authentication_headers()
        data = self.manager.authentication_for_opds_document
        return Response(data, 401, headers)
```

So in the situation you describe (credentials are provided but the token has expired) I would expect to get 401 and the authentication document, which I believe is handled by your code.

As far as I can tell, INVALID_CREDENTIALS is mainly sent in unlikely situations, like when the ILS (or equivalent) accepts your credentials but some third party (such as Overdrive) rejects those same credentials. It's good that you're also handling this case but I would expect it to be much less common.

Which are you seeing in the case of the expired OAuth token? I'm assuming it's 401 + authentication document.

ettore:
> The default lifetime for an OAuth token issued by the CM is 42 days, and Open eBooks hasn't overridden this. 

I definitely had to wait less than 42 days. I think if in Clever you say "This computer is shared" when you first authenticate, I'm thinking the expiration is triggered earlier?

> Which are you seeing in the case of the expired OAuth token? I'm assuming it's 401 + authentication document.

I got 401 + problem document. 

The only way I currently reproduced this is from the Catalog page, and I am trying to be generic enough where if instead of fetching the Catalog I fetched, say, My books, or maybe I did a download, I'll still catch the expired token situation with my code. I assume 401 + problem doc is returned for all api calls we use, but I am not 100% sure (also because I don't have a full map of all api calls).

leonardr:
That's interesting, I'm not sure what's going on.

When you authenticate with Clever, the circulation manager creates a brand new bearer token and sends that to the client. From that point onward the client is supposed to send that bearer token along with circulation manager requests.

The CM doesn't check in with Clever after the login, which is why we have our own expiration time of 42 days.

Looking at the source code, I think the server might be sending the Clever token over to the mobile client instead of creating a new one, but it shouldn't matter -- we don't know when Clever thinks the token will expire, and the expiration we put down for the token is 42 days in the future.

So I don't know why your token would expire after one day. If we _were_ periodically checking back with Clever in a way I don't currently see, then I can see the behavior you describe happening: Open eBooks sends a token that the circulation manager thinks is valid, so we authenticate you. But before we try to do something like check out a book, we check with Clever (this is the part I don't see happening). Clever says the token has expired, so we end up sending INVALID_CREDENTIALS instead of going forward with the checkout.

Regardless of when or how a token expires, I think this code is correct; it handles both cases.

ettore:
> So I don't know why your token would expire after one day.

That was just my guess, it might be longer. But I've been using Clever authentication often (less keyboard taps :)) in the last couple weeks. I did not observe this behavior when I chose "this computer is not shared" on Clever. I'll try to leave my phone alone and check back next week and see if it reproduced again.

Thank you for the code review and for checking, it's very helpful.

PR #1254, opened 2020-11-19
SIMPLY-2609 Restore reading progress in R2 reader

ErnestFan:
**What's this do?**
Store/restore reading progress from local storage/server
Present alert when reading progress on server does not match with local

[SIMPLY-2609](https://jira.nypl.org/browse/SIMPLY-2609?filter=-1)

Open any ePub book and make sure syncing is on
Navigate to any pages, then close the reader
Reopen reader and progress should be restored
P.S: This cannot be tested due to bug in r2-navigator, but devs can still validate by printing the locator value in debugging area

N/A

N/A

@ErnestFan


Simplified/Reader2/UI/NYPLBaseReaderViewController.swift
```diff
@@ -261,7 +263,63 @@ class NYPLBaseReaderViewController: UIViewController, Loggable {
...
+      return
+    }
+    let bookLocation = NYPLBookLocation(locator: locator, publication: publication, renderer: NYPLBookLocation.r2Renderer)
+    NYPLBookRegistry.shared().setLocation(bookLocation, forIdentifier: book.identifier)
+    bookmarksBusinessLogic.postReadPosition(locationString: bookLocation.locationString)
+  }
 
+  /// Restore locally stored location and navigate to it
+  /// Then sync location from server and present alert for navigation
+  private func restoreReadPosition() {
```
ettore, line 282: most of this is also business logic, but we could refactor this in a future ticket if that's more practical.

ettore (changes requested):

vladimirfedorov:
It may be a good idea to add a ticket to address the issue with r2-navigator in the future, later versions of r2-navigator probably had that problem fixed.

vladimirfedorov (approved):

Simplified/Reader2/UI/NYPLBaseReaderViewController.swift
ErnestFan: Refactored to bookmark business logic

Simplified/Reader2/UI/NYPLBaseReaderViewController.swift
```diff
@@ -261,7 +263,63 @@ class NYPLBaseReaderViewController: UIViewController, Loggable {
...
+    }
+    
+    bookmarksBusinessLogic.syncReadPosition(bookID: book.identifier, currentLocation: currentLocation, url: book.annotationsURL) { [weak self] (serverLocationString) in
+      guard let locationString = serverLocationString,
+        let locator = NYPLBookLocation(locationString: locationString, renderer: NYPLBookLocation.r2Renderer).convertToLocator() else {
+        Log.debug(#file, "Unable to create locator from server stored location string - \(serverLocationString ?? "")")
+        return
+      }
+      
+      self?.persentReadPositionNavigationAlert(locator: locator)
```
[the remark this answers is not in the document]
ErnestFan: fixed

Simplified/Reader2/UI/NYPLBaseReaderViewController.swift
```diff
@@ -261,7 +263,63 @@ class NYPLBaseReaderViewController: UIViewController, Loggable {
       updateBookmarkButton(withState: false)
     }
   }
+  
+  //----------------------------------------------------------------------------
+  // MARK: - Read Position
+  
+  private func storeReadPosition(locator: Locator) {
```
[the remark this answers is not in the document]
ErnestFan: Refactored to bookmark business logic

Simplified/Reader2/NYPLReaderBookmarksBusinessLogic.swift
```diff
@@ -337,7 +345,76 @@ class NYPLReaderBookmarksBusinessLogic: NSObject, NYPLReadiumViewSyncManagerDele
     completion()
   }
 
-  // TODO: Sync reading position
+  // MARK: - Read Position
+  
+  /// Post the read position to server if we have synced the read position from server
+  /// There is a 120 seconds interval to avoid high frequency of requests
+  /// - Parameters:
+  /// - locationString: A json string that contains required information to create a Locator object
```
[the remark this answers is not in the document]
ErnestFan: We are storing the location info as a string in NYPLBookLocation, so any format would work as long as we know how to decode it. I am now using dictionary instead of string, and generating a json string from the dictionary.

Simplified/Reader2/NYPLReaderBookmarksBusinessLogic.swift
```diff
@@ -337,7 +345,76 @@ class NYPLReaderBookmarksBusinessLogic: NSObject, NYPLReadiumViewSyncManagerDele
...
+      // Pass through without presenting the Alert Controller if:
+      // 1 - There is no recent page saved on the server
+      // 2 - The most recent page on the server comes from the same device
+      // 3 - The server and the client have the same page marked
+      
+      if (serverLocationString == nil ||
+        deviceID == self?.drmDeviceID ||
+        currentLocation?.locationString == serverLocationString) {
+        self?.shouldPostLastReadPosition = true
+        completion(serverLocationString)
```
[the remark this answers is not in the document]
ErnestFan: Fixed

Simplified/Reader2/NYPLReaderBookmarksBusinessLogic.swift
```diff
@@ -337,7 +345,76 @@ class NYPLReaderBookmarksBusinessLogic: NSObject, NYPLReadiumViewSyncManagerDele
...
+  /// Post the read position to server if we have synced the read position from server
+  /// There is a 120 seconds interval to avoid high frequency of requests
+  /// - Parameters:
+  /// - locationString: A json string that contains required information to create a Locator object
+  func postReadPosition(locationString: String) {
+    if !shouldPostLastReadPosition {
+      return
+    }
+    
+    DispatchQueue.global(qos: .background).async {
```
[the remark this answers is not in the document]
ErnestFan: Added a serial queue to NYPLBookmarkBusinessLogic

Simplified/Reader2/NYPLReaderBookmarksBusinessLogic.swift
```diff
@@ -337,7 +345,76 @@ class NYPLReaderBookmarksBusinessLogic: NSObject, NYPLReadiumViewSyncManagerDele
...
+      switch(self.readPositionSyncStatus) {
+      case .idle:
+        self.readPositionSyncStatus = .busy
+        self.queuedReadPosition = locationString
+        self.postQueuedReadPosition()
+        DispatchQueue.main.asyncAfter(deadline: .now() + 120) {
+          self.readPositionSyncStatus = .idle
+          if self.queuedReadPosition != "" {
+            self.postReadPosition(locationString: self.queuedReadPosition)
```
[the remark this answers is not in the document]
ErnestFan: Added a serial queue to NYPLBookmarkBusinessLogic

Simplified/Reader2/NYPLReaderBookmarksBusinessLogic.swift
```diff
@@ -337,7 +344,99 @@ class NYPLReaderBookmarksBusinessLogic: NSObject, NYPLReadiumViewSyncManagerDele
...
+    serialQueue.async {
+      self.queuedReadPosition = locationString
+      
+      if Date() > self.lastReadPositionUploadDate.addingTimeInterval(120) {
+        self.postQueuedReadPosition()
+      }
+    }
+  }
+  
+  @objc private func postQueuedReadPosition() {
```
ettore, line 400: this function seems to be called from 2 contexts:
- from inside the `serialQueue` (i.e. from `postReadPosition()`)
- from outside the `serialQueue` (i.e. as a NotificationCenter observer callback)

in the latter case we are accessing shared instance data outside the `serialQueue`, so we have race conditions. 

One way to fix this would be to add a new function that calls `postQueuedReadPosition()` from inside the `serialQueue`, and then update the NotificationCenter observation to call that new function instead of `postQueuedReadPosition`.

ettore (changes requested):
minor change to fix race conditions, good otherwise

Simplified/Reader2/NYPLReaderBookmarksBusinessLogic.swift
```diff
@@ -20,6 +20,9 @@ class NYPLReaderBookmarksBusinessLogic: NSObject, NYPLReadiumViewSyncManagerDele
   private let drmDeviceID: String?
   private let bookRegistryProvider: NYPLBookRegistryProvider
   private let currentLibraryAccountProvider: NYPLCurrentLibraryAccountProvider
+  private var lastReadPositionUploadDate: Date
+  private var queuedReadPosition: String = ""
+  private let serialQueue = DispatchQueue(label: "\(Bundle.main.bundleIdentifier!).bookmarkBusinessLogic")
```
ettore, line 25: a performance improvement here would be to specify the `target` param to be one of the global queues. This optimizes thread handling on the OS, i.e. preventing iOS to needlessly generate new threads.

Simplified/Reader2/UI/NYPLBaseReaderViewController.swift
```diff
@@ -261,7 +263,40 @@ class NYPLBaseReaderViewController: UIViewController, Loggable {
...
+  // MARK: - Read Position
 
+  private func restoreReadPosition() {
+    bookmarksBusinessLogic.restoreReadPosition(currentLocator: navigator.currentLocation, localFetchCompletion: { (localLocator) in
+      if let localLocator = localLocator {
+        self.navigator.go(to: localLocator, animated: true) {}
+      }
+    }) { [weak self] (serverLocator) in
+      if let serverLocator = serverLocator {
+        self?.presentReadPositionNavigationAlert(locator: serverLocator)
```
ettore, line 277: make sure to call this from main thread (you can use NYPLMainThreadRun helper) -- same thing for the `navigator.go(to:)` call

Simplified/Reader2/NYPLBookLocation+Locator.swift
```diff
@@ -0,0 +1,68 @@
...
+
+import Foundation
+import R2Shared
+
+extension NYPLBookLocation {
+  static let r2Renderer = "readium2"
+  
+  convenience init?(locator: Locator, publication: Publication, renderer: String) {
+    guard let idref = publication.idref(forHref: locator.href) else {
+      Log.warn(#file, "Unable to find Publication's idref from locator's idref - \(locator.href)")
```
ErnestFan, line 17: @ettore Since you are removing `Log.warn`, should I change this to `debug`? And the one on line 32 too

ettore: actually, sorry, I am not removing it. I spoke too soon!

Simplified/Reader2/NYPLReaderBookmarksBusinessLogic.swift
```diff
@@ -337,7 +344,105 @@ class NYPLReaderBookmarksBusinessLogic: NSObject, NYPLReadiumViewSyncManagerDele
...
+    }
+  }
+  
+  /// Post the read position to server if we have synced the read position from server
+  /// There is a 120 seconds interval to avoid high frequency of requests
+  /// - Parameters:
+  /// - locationString: A json string that contains required information to create a Locator object
+  private func postReadPosition(locationString: String) {
+    serialQueue.async {
+      self.queuedReadPosition = locationString
```
ettore, line 392: sorry I missed this earlier. There's a retain cycle between self > serialQueue > async closure > self

Also below in `postQueuedReadPositionInSerialQueue`

ettore (approved):
there's a retain cycle to fix (just needs [weak self]) other than that looks good to me

Simplified/Reader2/NYPLReaderBookmarksBusinessLogic.swift
ErnestFan: Fixed, thank you for catching it

PR #1258, opened 2020-12-01
Use Unified Logging System

ettore:
**What's this do?**
Updates our logging utilities to rely on os_log / Unified Logging System. This will allow easier debugging using the Console.app and `log` command line tools,  and their related filtering capabilities.

It also removes some old cruft that we no longer need.

faster / more performant logging
more easily searchable logs

- Open Console.app and select the device you're running the app on on the left sidebar. 
- You can customize the columns in the log entries by right-clicking on the toolbar (e.g. next to "Message")
- Launch the app
- You can also apply filtering by right clicking on a log entry you don't care for, and select Hide Library XYZ 
- From the Action menu you can show/hide Info and Debug log entries (disabled by default)
- Similarly you can filter on the even type (Errors have a yellow dot, default is light gray, debug is dark gray)

console-log

n/a

y

@ettore


ErnestFan (approved):
Thanks @ettore . This is very useful for debugging!

vladimirfedorov (approved):

ettore:
I expanded the above in this doc: https://github.com/NYPL-Simplified/Simplified/wiki/How-to-inspect-logs-and-app-behavior-on-SimplyE---Open-eBooks-for-iOS

PR #1264, opened 2020-12-11
SIMPLY-2961 "View Issues" section on Book Detail pages doesn't appear

ErnestFan:
**What's this do?**
"View Issues" section doesn't appear because there is no problem document when download failed.
The only fix for now is to remove the "Scroll down to view issue" message.

[SIMPLY-2961](https://jira.nypl.org/browse/SIMPLY-2961)

Force a download fail
The error message should be one line "The download could not be completed"

N/A

N/A

@ErnestFan

### DIFF
--- a/Simplified/it.lproj/Localizable.strings
+++ b/Simplified/it.lproj/Localizable.strings
@@ -31,6 +31,7 @@
 "DownloadCouldNotBeCompleted" = "Il download è fallito.";
 "DownloadCouldNotBeCompletedFormat" = "Il download di %@ non può essere completato.";
 "The download could not be completed.\nScroll down to 'View Issues' to see details." = "Il download non è stato completato.\nScorri in fondo alla pagina per i dettagli dell'errore.";
+"The download could not be completed." = "Il download non è stato completato.";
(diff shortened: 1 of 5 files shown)


vladimirfedorov (approved):

Simplified/it.lproj/Localizable.strings
```diff
@@ -31,6 +31,7 @@
 "DownloadCouldNotBeCompleted" = "Il download è fallito.";
 "DownloadCouldNotBeCompletedFormat" = "Il download di %@ non può essere completato.";
 "The download could not be completed.\nScroll down to 'View Issues' to see details." = "Il download non è stato completato.\nScorri in fondo alla pagina per i dettagli dell'errore.";
+"The download could not be completed." = "Il download non è stato completato.";
```
ettore, line 34: perfect translation!

ettore (approved):
approving pending the minor fix

Simplified/it.lproj/Localizable.strings
ErnestFan: I copied and pasted from the line above haha

PR #1267, opened 2020-12-14
[SIMPLY-3375] Fix NYPLProblemDocumentCacheManager

ErnestFan:
**What's this do?**
Complete the implementation of NYPLProblemDocumentCacheManager (Clear cache, get cached document etc)

[SIMPLY-3375](https://jira.nypl.org/browse/SIMPLY-3375)

Force a download fail on two different books, they should both show a problem document in the "View Issue" section
A download success on a book should remove the "View Issue" section

N/A

N/A

@ErnestFan

### DIFF
--- a/Simplified/ErrorHandling/NYPLProblemDocumentCacheManager.swift
+++ b/Simplified/ErrorHandling/NYPLProblemDocumentCacheManager.swift
@@ -27,38 +27,42 @@ extension Notification.Name {
   }
   
   // Member values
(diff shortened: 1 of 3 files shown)


ettore (approved):
this is a big improvement over what we had previously, so it's fine to merge as-is IMO. 

There are still thread-safety issues in the implementation but they are independent from your changes, so can you please file a ticket for that and put it in the staging sprint? it's lower priority compared to other things we need to do.

vladimirfedorov (approved):

PR #1268, opened 2020-12-15
SIMPLY-3361 CI workflow for running unit tests on PR

ettore:
**What's this do?**
Adds a GitHub Action to run unit tests whenever a PR is published.

It uses a Personal Access Token of the `simplified-ci` user to authenticate for fetching the private repos. The token value is saved in the IOS_DEV_CI_PAT repo secret. @leonardr please let me know if this ok security wise: per GitHub docs PATs are not the most secure authentication method, but it's the only way I was able to make this work in our set-up. Android folks are doing the same in my understanding.

There are a few not-strictly-needed verification checks left in there for the time being since this is still in flux.

Publishing this PR triggered the action and ran the build and unit tests. The [first time it failed](https://github.com/NYPL-Simplified/Simplified-iOS/runs/1554290357) (also see red cross next to commit 756bb05) because it detected a build error for OE. The last commit (7f9f49a) should fix the error, and the check should pass (hopefully).

n/a

yes

@ettore


ettore:
ok, I'm noticing that the action is already being triggered, which makes sense because it's using the source branch too, not just the destination. This first check will fail however, because there's an actual compile error in Open eBooks. I will fix that in a separate PR.

leonardr (approved):
Using a PAT is fine IMO. I'm happy that this branch runs tests for both SimplyE and open eBooks.

ErnestFan (approved):

vladimirfedorov (approved):

PR #1272, opened 2020-12-16
SIMPLY-3206 LCP audiobooks support

vladimirfedorov:
**What's this do?**
Adds LCP audiobook support to SimplyE.

https://jira.nypl.org/browse/SIMPLY-3206 

Add "LCP" library in the app, open "Audiotest" book

Please make sure to update NYPLAudiobookToolkit first

Yes

@vladimirfedorov


Simplified/Audiobooks/LCP/LCPAudiobooks.swift
```diff
@@ -0,0 +1,112 @@
...
+    if self.lcpLicense != nil {
+      completion(nil)
+      return
+    }
+    // Load publication to get license
+    let drm = DRM(brand: .lcp)
+    lcpService.loadPublication(at: audiobookUrl, drm: drm) { result in
+      switch result {
+      case .success(let updatedDrm):
+        self.lcpLicense = updatedDrm?.license
```
ettore, line 103: there's a possible retain cycle here depending on lcpService manages the closure. Can you fix (if needed) and also make sure this is documented on the LCPLibraryService class?

Simplified/Audiobooks/LCP/LCPAudiobooks.swift
```diff
@@ -0,0 +1,112 @@
...
+
+  /// Decrypt protected file
+  /// - Parameters:
+  ///   - url: encrypted file URL.
+  ///   - resultUrl: URL to save decrypted file at.
+  ///   - completion: decryptor callback with optional `Error`.
+  func decrypt(url: URL, to resultUrl: URL, completion: @escaping (Error?) -> Void) {
+    loadLicense { error in
+      if let error = error {
+        completion(error)
```
ettore, line 71: is this error reported to NYPLErrorLogger on the calling side? if not, we should do that here (assuming here it's better than there, I don't know about that). After that, it would be good to document what logging this method is doing too.

Simplified/Audiobooks/LCP/LCPAudiobooks.swift
ettore: also there are some multithreading issues I believe on the `lcpLicense` property, since `loadLicense` can be invoked asynchronously by `decrypt`. Can we just pass the `lcpLicense` as a completion param instead?

Simplified/Audiobooks/LCP/LCPAudiobooks.swift
```diff
@@ -0,0 +1,112 @@
...
+        return
+      }
+    }
+  }
+  
+  /// Load `DRMLicense` license for audiobook once
+  /// - Parameter completion: `LCPError`, if any
+  private func loadLicense(completion: @escaping (_ error: Error?) -> Void) {
+    // If the license has already been loaded, just call back
+    if self.lcpLicense != nil {
```
ettore, line 94: why don't we just do this at the call site and save ourselves one level of indirection? It would also simplify the semantics of this method, since in this `if` we're not really loading anything. If we do have a license we could just short-circuit to the `do-catch` block at line 78 (after refactoring that in a function).

Simplified/MyBooks/NYPLMyBooksDownloadCenter.m
```diff
@@ -548,16 +548,30 @@ - (void)deleteLocalContentForBookIdentifier:(NSString *const)identifier account:
...
       
       [[AudiobookFactory audiobook:dict ?: json] deleteLocalContent];
+      
+      if ([book.distributor isEqualToString:LCPAudiobooks.distributorKey]) {
+        if ([[NSFileManager defaultManager] fileExistsAtPath:bookURL.path]) {
+          NSError *error = nil;
+          [[NSFileManager defaultManager] removeItemAtURL:bookURL error:&error];
+          if (error) {
+            NSString *message = [NSString stringWithFormat:@"%@ %@", book.identifier, book.title];
+            [NYPLErrorLogger logError:error summary:@"Failed to delete audiobook local content" message:message metadata:NULL];
```
ettore, line 565: the message param is going to be deprecated soon. It's preferred to use the `metadata` field instead, because it's more descriptive and it is much easier to inspect on Crashlytics, since you can define your own keys-values.

Can you also change the summary "Failed to delete LCP audiobook local content"?

Simplified/MyBooks/NYPLMyBooksDownloadCenter.m
```diff
@@ -681,12 +714,10 @@ - (NSURL *)fileURLForBookIndentifier:(NSString *const)identifier
 - (NSURL *)fileURLForBookIndentifier:(NSString *const)identifier account:(NSString * const)account
 {
   if(!identifier) return nil;
-
```
ettore, line 684: can you leave this comment and move it to your new `pathExtensionForBook:`  method, unless it's wrong? it seems quite useful. It could be on its documentation for example.

Simplified/NYPLBookCellDelegate.m
```diff
@@ -183,10 +183,19 @@ - (void)openAudiobook:(NYPLBook *)book {
     dict[@"id"] = book.identifier;
   }
   
+  id<DRMDecryptor> audiobookDrmDecryptor = nil;
+
+  if ([book.distributor isEqualToString:LCPAudiobooks.distributorKey]) {
```
ettore, line 188: this check is repeated many times (3 in MyBooksDownloadCenter, 1 here) and it seems like a key part of the business logic to identify a LCP audiobook. It would be worth it to have this as a static function in LCPAudiobooks, like `func verifyBookIsLCP(_ book: NYPLBook)`. This would also be more future-proof in case we need to expand its semantics in the future.

ettore (changes requested):

Simplified/Audiobooks/LCP/LCPAudiobooks.swift
vladimirfedorov: Yes, it is in `NYPLAudiobookToolkit` https://github.com/NYPL-Simplified/NYPLAudiobookToolkit/blob/master/NYPLAudiobookToolkit/Player/LCPPlayer.swift line 33:

```
            delegate.decrypt(url: task.url, to: decryptedUrl) { error in
                if let error = error {
                    ATLog(.error, "Error decrypting file", error: error)
                    return
                }
```

Simplified/NYPLBookCellDelegate.m
vladimirfedorov: a good idea, added `canOpenBook` to `LCPAudiobooks`

Simplified/MyBooks/NYPLMyBooksDownloadCenter.m
vladimirfedorov: moved it to `pathExtensionForBook`, it's not wrong but it seems other libraries work fine with `.epub` file name extensions

Simplified/MyBooks/NYPLMyBooksDownloadCenter.m
vladimirfedorov: changed and added book.loggableShortString to metadata, I noticed another function logging it

ettore: oh nice, that's perfect

Simplified/Audiobooks/LCP/LCPAudiobooks.swift
vladimirfedorov: Multithreading can be an issue here... I don't think keeping caching the license worth the trouble, changed the code to read the license for each decrypt call (LCP saves the license to the book when we open it)

Simplified/Audiobooks/LCP/LCPAudiobooks.swift
vladimirfedorov: the code has changed; `completion` returns `license` and `error`

Simplified/Audiobooks/LCP/LCPAudiobooks.swift
```diff
@@ -0,0 +1,116 @@
...
+  /// Content dictionary for `AudiobookFactory`
+  @objc func contentDictionary() -> NSDictionary? {
+    let manifestPath = "manifest.json"
+    do {
+      // Relative path inside the audiobook
+      let data = try container.data(relativePath: manifestPath)
+      let publicationObject = try JSONSerialization.jsonObject(with: data, options: .allowFragments) as? NSDictionary
+      return publicationObject
+    } catch {
+      NYPLErrorLogger.logError(error, summary: "Error reading audiobook \(manifestPath)")
```
ettore, line 52: Instead of `manifestPath` (which is a constant and doesn't really add value) can you add something that specifies this is an LCP error? This also in the `init` at line 38

Simplified/MyBooks/NYPLMyBooksDownloadCenter.m
```diff
@@ -558,14 +558,34 @@ - (void)deleteLocalContentForBookIdentifier:(NSString *const)identifier account:
...
+          [[NSFileManager defaultManager] removeItemAtURL:bookURL error:&error];
+          if (error) {
+            [NYPLErrorLogger logError:error
+                              summary:@"Failed to delete LCP audiobook local content"
+                              message:NULL
+                             metadata:@{ @"book": [book loggableShortString] }];
+          }
+        }
+      }
+#endif
```
ettore, line 583: do you think it's possible to move all the code between 544-583 into a separate method? Just trying to avoid expanding this gigantic switch case and limit the pain  :)

Simplified/Audiobooks/LCP/LCPAudiobooks.swift
ettore: I see, that's pretty low level and it's just a log line. I would file an actual report with `NYPLErrorLogger.logError` so these surface on crashlytics. Make sure to mention LCP in the summary.

ettore (approved):
Minor comments/suggestions, look good otherwise.

ErnestFan (approved):

Simplified/Audiobooks/LCP/LCPAudiobooks.swift
vladimirfedorov: sure, added "LCP" in both cases

Simplified/MyBooks/NYPLMyBooksDownloadCenter.m
vladimirfedorov: looks more like a necessity than a possibility, moved into a separate method.

Simplified/Audiobooks/LCP/LCPAudiobooks.swift
vladimirfedorov: added logging with "LCP" text to `decrypt` function as well

PR #1274, opened 2020-12-16
Merge develop into feature/readium2

ettore:
**What's this do?**
This updates mainly the project to fix the non-drm build of the apps. There were a few compile time preprocessor macros that needed to be added to the R2dev target.

part of git flow

These changes were already code-reviewed. I had to fix the project file and a few of the build scripts for conflicts. I updated the README with new build instructions, which haven't really changed except I removed the skipR2 flag, because it was complicating the script too much, and the effort didn't seem worth it, since this is temporary now anyway.

n/a

yes

@ettore


ettore:
note that the check fails because I changed the name of the token. Even if I fixed that (I will do so on develop though) there are also probably more intricacies related to the separate R2 repos. Since that's a temporary thing, I'm postponing that work.

ErnestFan (approved):

vladimirfedorov (approved):

PR #1275, opened 2020-12-18
SIMPLY-3356 NYU Book status wording

vladimirfedorov:
**What's this do?**
- change book details "This open-access book is available to keep" to "This book is available to borrow." for cases when the book is available and has no specified number of total or available to borrow items.
- removes `NYPLBookButtonsStateCanKeep` case from `NYPLBookButtonsState` 

https://jira.nypl.org/browse/SIMPLY-3356
Currently NYU book details display the misleading status "This open-access book is available to keep". This status appears when OPDS entries for these books don't have any specified amount available to borrow and isn't connected with "open-access" link relation.

In "New York University Libraries" tap any book, e.g., "A Doll's House". Book status should read "This book is available to borrow.".

No

N/A

@vladimirfedorov


leonardr (commented):
This branch removes the "yours to keep" language altogether, rather than fixing it to only apply to books that are actually open-access. This is my preferred solution but I don't want to merge this code until we've actually agreed to go in that direction.

We also need to investigate whether this issue applies on Android, and make the messaging consistent across platforms. Removing the language here means also removing it from Android, even though I'm guessing Android displays the message only for actual open-access books.

vbessonov:
@vladimirfedorov, could you keep `This open-access book is available to keep` for open-access acquisition links to not change the current behaviour?

ettore (approved):
after reading the ticket, these changes make sense to me code-wise, but my approval is conditional to @leonardr's for timing about when we should merge this.

vladimirfedorov:
> @vladimirfedorov, could you keep `This open-access book is available to keep` for open-access acquisition links to not change the current behaviour?

Currently the app doesn't have that feature, we can't keep this line for open-access acquisitions but instead we can add this feature the way it is done on Android. If Android displays this text correctly, we can add the same logics to iOS version of the app.

leonardr (approved):
Risa has approved removing this message altogether until we come up with a better solution.

PR #1277, opened 2020-12-22
SIMPLY-3361 Add CI to upload build to iOS-binaries for every commit on develop

ettore:
**What's this do?**
Adds a GitHub action to archive SimplyE, export .ipa for adhoc distribution and upload it to [iOS-binaries](https://github.com/NYPL-Simplified/ios-binaries). You can see a first build uploaded there.
 
There's a lot of moving parts here so I plan to write a wiki entry about this. But briefly:
- I created a temporary Apple ID and added it to our iOS team in the Apple dev portal. This is used by CI to authenticate with Apple and download provisioning profiles and code sign the app. This user has the minimum set of permissions required to do so. I can share credentials with anybody. This user's email will at some point be replaced with a proper @nypl.org email.
- I use [Fastlane](https://docs.fastlane.tools/best-practices/continuous-integration/#continuous-integration) to download provisioning profiles. It's the only tool that was able to successfully install them so that Xcode can use them.
- Xcode also needs a certificate identity to codesign the build. This is imported via a GitHub secret, also protected by a password (also a GitHub secret, so maybe doesn't add much). I could maybe move this to fastlane if we think it's more secure.
- All credentials we need are in this repo's secrets. The documentation in `decode-install-secrets.sh` explains briefly how they are used.

I'd like @leonardr to take a look at the security aspects of this work. This is as secure as GitHub Secrets are secure, but these are accessed by a public workflow after all. @MalcolmMcFly correct me if I'm wrong but I think the android project is set up similarly.

yes

@ettore


ErnestFan (approved):

ettore:
I confirm that trying to print one of the secrets to a file and then `cat`ting the file does not print out the secrets in the console. This even if I run it as myself i.e. a user with push rights on this repo.

vladimirfedorov (approved):

fastlane/Fastfile
```diff
@@ -0,0 +1,20 @@
...
+  end
+
+  lane :fetch_install_provisioning do
+    sigh(
+      app_identifier: ENV['APP_BUNDLE_ID']
+    )
+    sigh(
+      app_identifier: ENV['APP_BUNDLE_ID'],
+      adhoc: "true"
+    )
```
leonardr, line 18: Based on looking stuff up this looks like we're getting an App Store provisioning profile and an ad hoc provisioning profile. Why are we doing both? Because we're building several different types of binaries?

scripts/ios-binaries-upload.sh
```diff
@@ -46,14 +46,14 @@ cp "$ADHOC_EXPORT_PATH/$APP_NAME.ipa" "$IOS_BINARIES_DIR_PATH/$IPA_NAME"
...
 cd "$IOS_BINARIES_DIR_PATH"
 git add "$IPA_NAME"
+git status
 
 if [ "$BUILD_CONTEXT" == "ci" ]; then
-  git config --global user.email "ci@librarysimplified.org" ||
-    fatal "could not configure git"
-  git config --global user.name "Library Simplified CI" ||
-    fatal "could not configure git"
+  git config --global user.email "ci@librarysimplified.org"
```
leonardr, line 52: The email we set up yesterday was librarysimplifiedci@nypl.org -- should we be using that one as the email of record?

fastlane/Fastfile
ettore: I added the App Store one too because I figure at some point in the future we'll need that too if we want to push to TestFlight. But I can comment that out for now.

leonardr: I don't mind, was just curious.

scripts/ios-binaries-upload.sh
ettore: I took this email from @io7m scripts some time ago and forgot about it, but yes it makes sense to swap it for the new one.

fastlane/Fastfile
ettore: I take that back, we do need it because the Xcode project is configured to use that provisioning profile during archiving by default, and we swap it only it during exporting.

leonardr:
Since this workflow only runs when something is merged into `develop`, you can't run code with access to our certificates just by putting up a PR -- it has to pass code review first. So, from a security perspective I think this is all right.

leonardr:
Except, stupid question, if someone does put up a random PR that includes changes to workflows that run _before_ merge, do those workflows run as soon as the PR comes up?

ettore:
> Except, stupid question, if someone does put up a random PR that includes changes to workflows that run _before_ merge, do those workflows run as soon as the PR comes up?

Right, those workflow _can_ run as soon as the PR comes up (you just have to specify a `on: pull_request:` clause in your workflow), but if the PR comes from a fork the secrets are not passed to the runner:
https://docs.github.com/en/free-pro-team@latest/actions/reference/encrypted-secrets#using-encrypted-secrets-in-a-workflow

If you have the rights to push to a branch in this repo then you'll be able to use secrets, but I think that's ok/expected, since you must be inside NYPL-Simplified's org or have been granted those rights in some way.

leonardr (approved):

PR #1279, opened 2020-12-29
SIMPLY-3361 Check build number before uploading

ettore:
**What's this do?**
The last commit triggers the workflow that archives and uploads a build to iOS-binaries only when the build number on develop is not already present on iOS-binaries. It will be up to the developer to bump the build number if we want a new build to be uploaded.

This PR also makes sure that a script command failure inside a pipe fails the build.

Once this is merged, any PR merged to develop will trigger the check-build-number workflow, which if it succeeds will trigger the actual archive-and-upload workflow.

n/a

https://github.com/NYPL-Simplified/Simplified/wiki/iOS-CI-CD

@ettore


ettore:
The failures are caused by use of secrets to checkout the private repos, but secrets are not shared if the PR comes from an external repo. This is a problem separate from this PR, but it's very annoying since we want anybody to push PR to us and be able to run the checks. I'll see if I can find a workaround.

ettore:
sorry for the noise here, lots of trial and error. The gist here is that we'll need to post PRs from branches off of the main repo if we want to be able to run the DRM build and unit tests instead than our own forks.

ErnestFan (approved):

PR #1280, opened 2020-12-31
SIMPLY-3361 Upload to testflight for release branches

ettore:
**What's this do?**
Adds a step in the main build workflow to also upload to TestFlight if the branch name starts with `release/`. The actual upload to TF is done by `fastlane deliver`.

Our release process is started with a new branch called e.g. `release/3.4.5`. Once we do that, this will trigger a build that is uploaded to both iOS-binaries (as ad-hoc) and to TestFlight (as app-store). 

n/a

partially: I need to update the main CI doc in the wiki.

**Did you bump the SimplyE build number?**
yes, to verify if the trigger worked on a release/ branch.

@ettore


ErnestFan (approved):

ettore:
the last force-push is because I forgot to disable a couple more checks in the testflight_upload script... sorry!

PR #1288, opened 2021-01-14
SIMPLY-705 Improvement to Reader Page + Location in footer

ErnestFan:
**What's this do?**
Include chapter title in reader footer
Fix build script for R2 dev

[SIMPLY-705](https://jira.nypl.org/browse/SIMPLY-705)

Make sure R2 is on in developer setting menu
Open any book except audiobook
The footer in the reader should show the current location in the format of `Page x of x (Chapter Title)`
The chapter title is only shown when it's available

N/A

N/A

N/A

@ErnestFan

### DIFF
--- a/Simplified.xcodeproj/project.pbxproj
+++ b/Simplified.xcodeproj/project.pbxproj
@@ -6044,10 +6044,7 @@
 					"$(inherited)",
 					"@executable_path/Frameworks",
 				);
-				LIBRARY_SEARCH_PATHS = (
-					"$(inherited)",
-					"$(PROJECT_DIR)/Carthage/Build/iOS/ZIPFoundation.framework",
-				);
+				LIBRARY_SEARCH_PATHS = "$(inherited)";
 				MARKETING_VERSION = 3.7.0;
 				PRODUCT_BUNDLE_IDENTIFIER = org.nypl.labs.SimplyE;
 				PRODUCT_MODULE_NAME = SimplyE;
@@ -6108,10 +6105,7 @@
 					"$(inherited)",
 					"@executable_path/Frameworks",
 				);
-				LIBRARY_SEARCH_PATHS = (
-					"$(inherited)",
-					"$(PROJECT_DIR)/Carthage/Build/iOS/ZIPFoundation.framework",
-				);
+				LIBRARY_SEARCH_PATHS = "$(inherited)";
 				MARKETING_VERSION = 3.7.0;
 				PRODUCT_BUNDLE_IDENTIFIER = org.nypl.labs.SimplyE;
 				PRODUCT_MODULE_NAME = SimplyE;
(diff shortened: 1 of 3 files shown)


ErnestFan:
Currently in our R1 reader, we show the location in a format of `current page within chapter / total pages in chapter (chapter title)`. But I don't think this is possible as we don't have information about the total pages in current chapter, correct me if I'm wrong.

Simplified.xcodeproj/project.pbxproj
```diff
@@ -6030,6 +6044,10 @@
 					"$(inherited)",
 					"@executable_path/Frameworks",
 				);
+				LIBRARY_SEARCH_PATHS = (
+					"$(inherited)",
+					"$(PROJECT_DIR)/Carthage/Build/iOS/ZIPFoundation.framework",
```
ettore, line 6049: this should not be necessary, because the all other frameworks are in the same directory. Same for the one below.

Simplified.xcodeproj/project.pbxproj
```diff
@@ -2384,6 +2388,7 @@
 				739E6120244A0D8600D00301 /* FirebaseAnalytics.framework in Frameworks */,
 				731B2B25244A2B6E00A7A649 /* R2Shared.framework in Frameworks */,
 				739E6121244A0D8600D00301 /* nanopb.framework in Frameworks */,
+				170E5FBE25AFDD65007D6DE6 /* ZIPFoundation.framework in Frameworks */,
```
ettore, line 2391: was this missing only on the R2dev target or also other targets? In the latter case, that's weird, because the build was passing anyway.

ettore (commented):
> Currently in our R1 reader, we show the location in a format of `current page within chapter / total pages in chapter (chapter title)`. But I don't think this is possible as we don't have information about the total pages in current chapter, correct me if I'm wrong.

Is there something that can be done using the progressWithinChapter variable (the name might be slightly different, going off memory)? 

2nd question: regarding the project changes, did you just change the R2dev target or the others too? Just wondering.

Simplified.xcodeproj/project.pbxproj
ErnestFan: Yeah, it builds fine but crash on launch, at least on the simulator.
It's missing on all the targets, I added it on the targets that use r2-lcp.

Simplified.xcodeproj/project.pbxproj
ErnestFan: Removed, not sure why Xcode added that

ErnestFan:
> > Currently in our R1 reader, we show the location in a format of `current page within chapter / total pages in chapter (chapter title)`. But I don't think this is possible as we don't have information about the total pages in current chapter, correct me if I'm wrong.
> 
> Is there something that can be done using the progressWithinChapter variable (the name might be slightly different, going off memory)?
> 
> 2nd question: regarding the project changes, did you just change the R2dev target or the others too? Just wondering.

1) The progressWithinChapter is a float number, but without the total number of pages in chapter that's useless
2) Added on all targets that use r2-lcp

ettore (approved):

vladimirfedorov (approved):

PR #1289, opened 2021-01-15
SIMPLY-3342 Refresh credentials if expired

ettore:
**What's this do?**
- Avoid displaying a Sign-In modal with a Sign Out button
- Attempt no-UI refresh when possible, otherwise present modal and let the user sign-in. Prefill modal with current credentials (that's why I added `forceEditability` flag.) I am still trying to understand why there is logic in the modal to show a Sign Out button at all. That will be handled in a separate ticket.
- Fix up/simplify some error-handling logic in MyBooksDownloadCenter
- Avoid infinite async cycles when refreshing credentials in RemoteVC
- Refactor refresh credentials logic from VC to BusinessLogic class

The last point is the core of this change. You'll notice that the `refreshAuthIfNeeded(usingExistingCredentials:completion:)` function is not functionally equivalent to what what was present in NYPLAccountSignInViewController. This is because the previous logic I think had bugs (for example we should never execute immediately an OAuth authentication because that will redirect to Safari, which would be jarring).

https://jira.nypl.org/browse/SIMPLY-3342

All sign in and sign out methods should be tested. Especially via the modal.

n/a

yes

yes

@ettore


Simplified/MyBooks/NYPLMyBooksDownloadCenter.m
```diff
@@ -209,11 +209,6 @@ - (void)URLSession:(__attribute__((unused)) NSURLSession *)session
        summary:[NSString stringWithFormat:@"Error parsing problem doc downloading %@ book", book.distributor]
        metadata:@{ @"book": [book loggableShortString] }];
     }
-    [self logBookDownloadFailure:book
```
ettore, line 212: this was a dupe. The actual logging is at the bottom of the (very long) function.

Simplified/MyBooks/NYPLMyBooksDownloadCenter.m
```diff
@@ -319,74 +313,39 @@ - (void)URLSession:(__attribute__((unused)) NSURLSession *)session
...
-          [NYPLAlertUtils presentFromViewControllerOrNilWithAlertController:alert viewController:nil animated:YES completion:nil];
-        }
-      } else if (needsAuth) {
-        NYPLLOG(@"Present sign in VC");
-        [NYPLAccountSignInViewController
-         requestCredentialsUsingExisting:NO
-         completionHandler:^{
-          [self startDownloadForBook:book];
-        }];
+      BOOL hasCredentials = [NYPLUserAccount.sharedAccount hasCredentials];
```
ettore, line 338: this logic is not equivalent, but all we wanted to do here is 2 things: handle the expired creds error, and alert user in all other cases. The original logic was leaving some error paths unhandled.

Simplified/Network/NYPLRemoteViewController.m
```diff
@@ -128,54 +130,38 @@ - (void)load
...
-                                          ifNeededForResponse:response
-                                                 responseData:data
-                                                responseError:error
-                                      authenticationPreflight:^{ self.reloadView.hidden = NO; }
-                                     authenticationCompletion:^{
-        // make sure to retain the reauthenticator until end of auth flow and
-        // then break any possible retain cycle
-        reauthenticator = nil;
-        NYPLLOG(@"Re-loading from RemoteVC because authentication expired");
-        [self load];
```
ettore, line 150: this was an async infinite loop. Now we just fail, show the retry button, and record the need to reauthenticate in case we failed for that reason. Note that `needsReauthentication` is only accessed from main thread.

vladimirfedorov (approved):

Simplified/NYPLBookCellDelegate.m
```diff
@@ -81,20 +81,31 @@ - (void)didSelectDownloadForBook:(NYPLBook *)book
...
+                                            withDevice:[user deviceID]]) {
+    // NOTE: This was cut and pasted while refactoring preexisting work:
+    // "This handles a bug that seems to occur when the user updates,
+    // where the barcode and pin are entered but according to ADEPT the device
+    // is not authorized. To be used, the account must have a barcode and pin."
+    NYPLReauthenticator *reauthenticator = [[NYPLReauthenticator alloc] init];
+    [reauthenticator authenticateIfNeeded:user
+                 usingExistingCredentials:YES
+                 authenticationCompletion:^{
+      dispatch_async(dispatch_get_main_queue(), ^{
```
ErnestFan, line 99: Not important: Do we want to use `[NYPLMainThreadRun asyncIfNeeded:]` here just to keep things clean/consistent?

ettore: in general I have done that, but in this case since this logic is so touchy I left it as it was before. To test this I had do manually force the code to go here because I wasn't able to reproduce the conditions naturally.

ErnestFan (approved):

PR #1293, opened 2021-01-21
Merge R2 into develop

ettore:
**What's this do?**
Merges R2 into develop. It will appear as disabled by default.

https://jira.nypl.org/browse/SIMPLY-3376

See ticket.

The R2 branch refactored some code used also by R1, so we'll need to QA those changes as indicated by the ticket. Also, a bug exists in Readium2 that causes errors when jumping to a bookmark location. This will be solved after updating R2 to most recent `develop`.

[Bump the SimplyE build number to generate a new build on NYPL-Simplified/iOS-binaries]
yes

yes

@ettore


ettore:
ugh, conflicts -- will close and reopen from branch where I'll fix the conflicts.

PR #1296, opened 2021-01-22
Fix Readium 2 build for SimplyE with Content Protection

vladimirfedorov:
**What's this do?**
This is a buildable version of SimplyE with R2 Content Protection without DRM support created as a part of migration to R2 with Content Protection. NYPL R2 frameworks are removed and replaced with Readium versions. 

https://jira.nypl.org/browse/SIMPLY-3428

Open any book without DRM protection.

Please make sure to run `./scripts/bootstrap-drm.sh`, the workspace for R2 build doesn't contain R2 projects anymore and use recent R2 frameworks.

No

No

@vladimirfedorov


ettore:
@vladimirfedorov since we merged Readium2 yesterday, can you rebase your branch on top of `develop`? and then since this is not supporting DRM, I would target a new feature branch (also starting from `develop`) instead of `develop`.

ettore:
one more thing: now that we have CI, there's an assumption in the CI scripts that branches named `feature/XYZ` are interpreted as destination branches. The reasoning is that a feature branch will cause a build to be pushed to iOS-binaries and firebase. So I would also rename your `feature/content-protection` source branch to something else, for example using the ticket number as we usually do.

PR #1297, opened 2021-01-22
[SIMPLY-2906] Hide/show card creation button based on Juvenile eligibility

ErnestFan:
**What's this do?**
Check Juvenile eligibility when displaying AccountSettingDetailView
Hide/show card creation button based on eligibility

[SIMPLY-2906](https://jira.nypl.org/browse/SIMPLY-2906)

Login with Juvenile and parent account on NYPL library
Go to account setting detail
Card creation button should be hidden/shown based on account type

Using latest commit `e802572` on CardCreator-iOS framework

Yes, bumped build number

N/A

@ErnestFan


Simplified/Accounts/Library/Account.swift
```diff
@@ -234,18 +234,17 @@ class OPDS2SamlIDP: NSObject, Codable {
       if trimmedUrlStr.lowercased().hasPrefix("nypl.card-creator:") {
         let cartCreatorUrlStr = String(trimmedUrlStr.dropFirst("nypl.card-creator:".count))
         signUpUrl = URL(string: cartCreatorUrlStr)
-        supportsCardCreator = (signUpUrl != nil)
```
ettore, line 237: I think the intention here was to use the `signUpUrl` as a way to tell whether the library (not the user)  can support card creation. And I am not 100% sure but I _think_ `supportsCardCreator` actually means "supports native card creator"... something to clarify with Leonard maybe.

Simplified/SignInLogic/NYPLSignInBusinessLogic.swift
```diff
@@ -635,4 +635,18 @@ extension NYPLSignInBusinessLogic {
...
+    guard let parentBarcode = self.resolveUserBarcode() else {
+      completion()
+      return
+    }
+    
+    let coordinator = juvenileCardCreationCoordinator ?? makeJuvenileCardCreationCoordinator(using: parentBarcode)
+    juvenileCardCreationCoordinator = coordinator
+
+    coordinator.checkJuvenileCreationEligibility(parentBarcode: parentBarcode) { [weak self] error in
+      self?.libraryAccount?.details?.setCardCreationEligibilityError(error: error)
```
ettore, line 648: I don't know if it's correct to do this. Here you're modifying a library-related state (supportsCardCreator) based on the user's state (can the user create dependents?) but as far as I understand `supportsCardCreator` just means whether the library supports user creation via native card creator. I admit this was not documented very well, but it's what I can understand from the business logic class.

it might be cleaner to set the eligibility state as a new property on the business logic and then modify `juvenileCardsManagementIsPossible` to use it in addition to the other conditions.

ettore (changes requested):

Simplified/SignInLogic/NYPLSignInBusinessLogic.swift
ErnestFan: Right, thought the `supportsCardCreator` is an user's state for some reason, fixed

Simplified/Accounts/Library/Account.swift
ErnestFan: fixed

Simplified/Accounts/Library/Account.swift
```diff
@@ -334,6 +333,15 @@ class OPDS2SamlIDP: NSObject, Codable {
     }
   }
   
+  func setCardCreationEligibilityError(error: Error?) {
```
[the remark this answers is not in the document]
ErnestFan: Added flag in signInBusinessLogic instead

Simplified/NYPLSettingsAccountDetailViewController.m
```diff
@@ -213,26 +213,37 @@ - (void)viewDidLoad
   self.view.backgroundColor = [NYPLConfiguration backgroundColor];
   self.tableView.keyboardDismissMode = UIScrollViewKeyboardDismissModeInteractive;
 
+  UIActivityIndicatorView *activityIndicator = [[UIActivityIndicatorView alloc] initWithActivityIndicatorStyle: UIActivityIndicatorViewStyleGray];
```
ettore, line 216: by moving the set up and start of the activity indicator here there's a possibility for it to start spinning even in the case at line 233 i.e. when we just have to set up the views. If there is a condition where we just have to call `setupViews` and we're done, we could simply `return` early and then do the activity indicator setup.

Simplified/NYPLSettingsAccountDetailViewController.m
```diff
@@ -213,26 +213,37 @@ - (void)viewDidLoad
...
+      [self setupViews];
+      self.hiddenPIN = YES;
+      [self accountDidChange];
+      [self updateShowHidePINState];
+    });
+  };
+  
   if (self.businessLogic.libraryAccount.details != nil) {
-    [self setupViews];
+    if ([self.businessLogic juvenileCardsManagementIsPossible]) {
```
ettore, line 232: wondering if we should remove this `if` and always call `checkCardCreationEligibilityWithCompletion` here.  Technically if somehow the business logic is initialized in a way that `juvenileCardsManagementIsPossible` returns true, here we won't check for eligibility even if we should. It seems like we're relying on something external to this class.

It seems more correct to me to always check on eligibility? If so my previous comment about the activity indicator is irrelevant.

Simplified/NYPLSettingsAccountDetailViewController.m
```diff
@@ -213,26 +213,37 @@ - (void)viewDidLoad
...
-      dispatch_async(dispatch_get_main_queue(), ^{
-        [activityIndicator removeFromSuperview];
-        if (success) {
-          [self setupViews];
-          self.hiddenPIN = YES;
-          [self accountDidChange];
-          [self updateShowHidePINState];
-        } else {
+      if (success) {
+        completion();
```
ettore, line 240: now that we loaded the authentication document, shouldn't we check for eligibility?

ettore:
Looks good to me pending verification for OE and modal sign-in.

ettore (approved):

vladimirfedorov (approved):

PR #1298, opened 2021-01-25
SIMPLY-3436 Remove sign-out logic from SignIn modal

ettore:
**What's this do?**
Removes any remnants of Sign Out logic from the sign-in modal.

https://jira.nypl.org/browse/SIMPLY-3436

the only affected logic is related to SAML, so I would verify that sign-in and borrowing books when signed-out works correctly. It would also be good to check the behavior of the app is correct when the token expires: it should re-open the sign-in modal in that case.

no

yes

yes

@ettore


ErnestFan (approved):

vladimirfedorov (approved):

ettore:
forced-push to fix conflicts

PR #1300, opened 2021-01-27
Remove deprecated message param from NYPLErrorLogger

ettore:
**What's this do?**
Removes the deprecated `message` parameter from NYPLErrorLogger apis. 

The message parameter was misleading because it is not searchable and yet required when calling from objc. Also, it's superfluous since we have the open `metadata` param.

not user facing, but a regression will cover these changes.

n/a

yes

yes

@ettore 

crashlytics-groups


ErnestFan (approved):

vladimirfedorov (approved):
Looks good.

PR #1302, opened 2021-02-03
SIMPLY-2401 Process sign-in/out problem docs before HTTP status codes

ettore:
**What's this do?**
- consolidates handling of authentication challenges, possibly fixing a couple bugs (see related commit for more details) 
- refactors the sign in and sign out logic to use a custom NYPLNetworkExecutor object instead of URLSession so that we can take advantage of its centralized response processing. This required a few refactors:
    1. sign-in/out requests should not be cached, so I parametrized caching behavior on the executor
    2. sign-in/out requires to provide credentials for authentication challenges without using the UserAccount singleton, because the user is not fully signed in yet. So I factored that into a protocol (NYPLBasicAuthCredentialsProvider) that both NYPLUserAccount and the sign-in VCs conform to.
    3. I refactored the auth challenge functions into an object for better reusability for different credentials providers (see (2)).

https://jira.nypl.org/browse/SIMPLY-2401

We should test all authentication methods since this is a pretty wide refactor.

n/a

yes

yes

@ettore


Simplified/Network/NYPLCaching.swift
```diff
@@ -183,12 +183,36 @@ extension HTTPURLResponse {
...
+/// `ephemeral` caching corresponds to `NYPLSessionConfiguration::ephmeral`.
+/// `default` caching corresponds to the "stock" behavior of URLSession in
+/// processing caching headers, caching responses to memory and disk.
+/// `fallback` caching is like `default`, but the executor will attempt to cache
+/// responses even when these would not be cached by URLSession despite having a
+/// sufficient set of caching headers per https://tools.ietf.org/html/rfc7234 --
+/// see `HTTPURLResponse::modifyingCacheHeaders()`.
+enum NYPLCachingStrategy: NSInteger {
+  case ephemeral
+  case `default`
```
ErnestFan, line 198: Is there a reason why `default` is wrapped in apostrophes?

ErnestFan (approved):

Simplified/Network/NYPLCaching.swift
ettore: it's because it's a swift keyword, so it's its way to allow that value

vladimirfedorov (approved):

ettore:
I forgot to add some unit tests but I will do so in a subsequent PR.

PR #1307, opened 2021-02-10
SIMPLY-3430 LCP Passphrase retrieval

ErnestFan:
**What's this do?**
This retrieve the LCP Passphrase from the `hint link` within the license
Removed some unnecessary files that we added for LCP passphrase before migrating to CP

[SIMPLY-3430](https://jira.nypl.org/browse/SIMPLY-3430)

Go to setting, turn on `test libraries` and `r2`
Go to library `LCP`
Select an ePub, download and read
The passphrase retrieval request is performed in the background and the book should be opened when it's completed

N/A

N/A

N/A

@ErnestFan


Simplified.xcodeproj/project.pbxproj
```diff
@@ -5620,10 +5599,9 @@
 				A4BA1D131B43046B006F83DF /* NYPLCatalogGroupedFeed.m in Sources */,
 				11F3773319E0876F00487769 /* NYPLCatalogFacet.m in Sources */,
 				73C3CF5425C8EB6900CA8166 /* NYPLUserAccount.swift in Sources */,
-				21F238C32499155E004DC0B1 /* AdobeDRMContainer.mm in Sources */,
+        21F238C32499155E004DC0B1 /* AdobeDRMContainer.mm in Sources */,
```
ErnestFan, line 5602: Not sure why Xcode removed this line, I added it back manually but the spacing is a bit off

Simplified.xcodeproj/project.pbxproj
```diff
@@ -6468,7 +6446,6 @@
 				CLANG_WARN_NON_LITERAL_NULL_CONVERSION = YES_ERROR;
 				CODE_SIGN_ENTITLEMENTS = Simplified/SimplyE.entitlements;
 				CODE_SIGN_IDENTITY = "iPhone Developer";
-				CURRENT_PROJECT_VERSION = 7;
```
ettore, line 6471: weird, I think this was a merge failure from my PR. `17` is actually incorrect because it's related to version 3.6.3, but on this branch we use 3.7.0. Do you mind making sure SimplyE and SimplyE-R2dev targets both have 3.7.0 (7) ?

Simplified/Reader2/LCP/LCPPassphraseAuthenticationService.swift
```diff
@@ -0,0 +1,46 @@
...
+      let hintURL = URL(string: hintLink.href) else {
+      Log.error(#file, "LCP Authenticated License does not contain valid hint link")
+      completion(nil)
+      return
+    }
+    
+    NYPLNetworkExecutor.shared.GET(hintURL) { (result) in
+      switch result {
+      case .success(let data, _):
+        if let json = try? JSONSerialization.jsonObject(with: data, options: []) as! [String:Any],
```
ettore, line 26: since you are in an `if-let` can you use `as?` instead of `as!` for more safety?

Simplified/Reader2/LCP/LCPPassphraseAuthenticationService.swift
```diff
@@ -0,0 +1,46 @@
...
+          let passphrase = json["passphrase"] as? String {
```
ettore, line 27: do we have docs for the format of the response? For example that you're looking for that specific `passphrase` key.  If it's documented somewhere it would be good to link it somewhere in this class.

Simplified/Reader2/LCP/LCPPassphraseAuthenticationService.swift
```diff
@@ -0,0 +1,46 @@
...
+          completion(passphrase)
+        } else {
+          Log.error(#file, "Error parsing JSON or finding passphrase key/value.")
```
ettore, line 30: should we log an event instead, since it seems like a rather serious error? That way we could log the whole json maybe (unless it's really huge and has sensitive info).

Simplified/Reader2/LCP/LCPPassphraseAuthenticationService.swift
```diff
@@ -0,0 +1,46 @@
...
+          completion(nil)
+        }
+      case .failure(let error, _):
+        NYPLErrorLogger.logError(
+          withCode: .lcpPassphraseAuthorizationFail,
+          summary: "Unable to retrieve LCP passphrase from \(hintURL)",
```
ettore, line 36: I would move the `hintURL` to the metadata because that can vary.

ettore (commented):
looks good, just minor stuff. One question though: what happens on the originating side when we call `completion(nil)`? is the error handled there gracefully to the user?

Simplified.xcodeproj/project.pbxproj
ErnestFan: You are right, SimplyE has 3.7.0(17) for some reason, SimplyE-R2dev is 3.7.0(7) though.
Fixed

Simplified/Reader2/LCP/LCPPassphraseAuthenticationService.swift
ErnestFan: It's not on Readium's requirement, but I will check with the backend developer.
https://readium.org/lcp-specs/releases/lcp/latest#41-introduction

Simplified/Reader2/LCP/LCPPassphraseAuthenticationService.swift
ErnestFan: The `json` won't be available since it's outside of the `if-let`, should I log the plain `data`? But it won't be readable on Crashlytics...

Simplified/Reader2/LCP/LCPPassphraseAuthenticationService.swift
ettore: sounds good -- that link is a good thing to add though.

Simplified/Reader2/LCP/LCPPassphraseAuthenticationService.swift
ettore: doh you're right... you could log the data converted to string though.

ErnestFan:
> completion(nil)

`r2-lcp-swift` will cancel the `openBook` action if passphrase is nil.
Which ends up in `NYPLReaderExtensions.swift` - line 62, case `cancelled` in `presentBook(book:)`.
And shows an alert saying `This book is corrupt and cannot be read`

ettore (approved):
looks good to me. Before merging though make sure to add some docs for the retrieve passphrase contract, if available.

ErnestFan:
> looks good to me. Before merging though make sure to add some docs for the retrieve passphrase contract, if available.

There is no documentation on the passphrase API endpoint, I will just merge it as-is.

PR #1309, opened 2021-02-11
SIMPLY-3425 Migrate R2 Dependencies for Adobe DRM (ePubs)

vladimirfedorov:
**What's this do?**
Adds Adobe DRM protection support to R2 with Content Protection

https://jira.nypl.org/browse/SIMPLY-3425

Open NYPL, open any Adobe DRM-protected book

No

No

Yes

@vladimirfedorov


Simplified/Reader2/AdobeDRM/AdobeDRMContentProtection.swift
```diff
@@ -0,0 +1,45 @@
...
+
+#if FEATURE_DRM_CONNECTOR
+
+class AdobeDRMContentProtection: ContentProtection {
+  func open(asset: PublicationAsset, fetcher: Fetcher, credentials: String?, allowUserInteraction: Bool, sender: Any?, completion: @escaping (CancellableResult<ProtectedAsset?, Publication.OpeningError>) -> Void) {
+    
+    // Publication asset must be a FileAsset, as we are opening a file
+    // If not, we can't open and use Adobe DRM
+    guard let fileAsset = asset as? FileAsset else {
+      completion(.failure(.unavailable(nil)))
```
ettore, line 20: should we add a `Log.error` line to this?

Simplified/Reader2/AdobeDRM/AdobeDRMContentProtection.swift
```diff
@@ -0,0 +1,45 @@
...
+      return
+    }
+    
+    do {
+      // META-INF is a part of .epub structure
+      let encryptionPath = "META-INF/encryption.xml"
```
ettore, line 26: do you have docs/specs that describe this encryption.xml path? It would be good to reference them here.

Simplified/Reader2/AdobeDRM/AdobeDRMFetcher.swift
```diff
@@ -0,0 +1,62 @@
...
+    self.fetcher = fetcher
+    self.links = fetcher.links
+  }
+  
+  /// Known resources available in the medium, such as file paths on the file system.
+  var links: [Link]
+  
+  /// Get resource such as content file by its link.
+  ///
+  /// `AdobeDRMFetcher` `get` function open .epub resouorces using the `fetcher` passed to `init`.
```
ettore, line 40: typos (resouorces and AAdobe)

Simplified/Reader2/AdobeDRM/AdobeDRMContentProtection.swift
```diff
@@ -0,0 +1,45 @@
...
+      let protectedAsset = ProtectedAsset(
+        asset: asset,
+        fetcher: adobeFetcher,
+        onCreatePublication: nil
+      )
+      completion(.success(protectedAsset))
+    } catch {
+      // .success(nil) means it is not an asset protected with Adobe DRM
+      // Streamer continues to iterate over available ContentProtections
+      completion(.success(nil))
```
ettore, line 39: so this means that no matter the reason why `resource.read().get()` throws, here we are saying "there's no valid Adobe DRM, but there IS an encrypted asset that someone else should decrypt" ?  And basically "Failure" means "there's no encrypted asset" ?

ettore (commented):
I have a couple questions but this looks good to me.

Simplified/Reader2/AdobeDRM/AdobeDRMContentProtection.swift
vladimirfedorov: It's a good question if it can be anything other than file... FileAsset is the only class implementing PublicationAsset protocol at the moment.

Simplified/Reader2/AdobeDRM/AdobeDRMContentProtection.swift
vladimirfedorov: it's in ADEPT code, there's no good documentation about this file, except it must be present in .epub for DRM to work. File data (it describes encryption algorithms used to encrypt each file) is passed to Adobe DRM decryption functions.

Simplified/Reader2/AdobeDRM/AdobeDRMFetcher.swift
vladimirfedorov: my keyboard is killing me. fixed

Simplified/Reader2/AdobeDRM/AdobeDRMContentProtection.swift
ettore: in that case it might be good to add one for future proofing

Simplified/Reader2/AdobeDRM/AdobeDRMContentProtection.swift
vladimirfedorov: not exactly. this function is called once, when we open a book. at this moment Streamer doesn't know what Content Protection can work with it, and iterates over Content Protections it has (the app initialise Streamer passing an array of Content Protection classes). If the code in do..catch fails, it means it couldn't initialise AdobeDRMFetcher, but there may be other types of Content Protection that work with it. If we return `.failure` here, Streamer stops iterating and shows the error we pass. It means if the book is not protected, Streamer will show the error and won't try to open it without protection. When Streamer receives `.success` with `nil` value for `ProtectedAsset`, it goes to the next Content Protection.

In short, `.failure` means "this is an adobe DRM protected item, but we can't decrypt it for some reason". `catch` block is called when we have no DRM at all.

Simplified/Reader2/AdobeDRM/AdobeDRMContentProtection.swift
vladimirfedorov: added

Simplified/Reader2/AdobeDRM/AdobeDRMContentProtection.swift
ErnestFan: Would `resource.read().get()` comes back with valid data if the resource is LCP ePub?
If that's the case, we will not be calling `completion(nil)`

vladimirfedorov: > Would resource.read().get() comes back with valid data if the resource is LCP ePub?

No, LCP `.epub`s don't contain `META-INF/encryption.xml`. LCP checks for `.lcpl` in `.epub` (which is embedded into the book during fulfillment process).

ettore: that makes sense. could you add that description to the code? it’s makes things much clearer.

ettore (approved):

ErnestFan (approved):

PR #1318, opened 2021-02-18
Merge develop into CP branch

ettore:
**What's this do?**
merges recent changes (including critical crash fixes) from develop

to send out a new build with the latest CP changes to QA

all this code has been already reviewed

n/a

yes

yes

@ettore


Simplified.xcodeproj/project.pbxproj
```diff
@@ -6162,7 +6157,7 @@
 					"$(inherited)",
 					"@executable_path/Frameworks",
 				);
-				MARKETING_VERSION = 3.7.0;
+				MARKETING_VERSION = 3.6.4;
```
ErnestFan, line 6160: Should we keep 3.7.0 for CP branch?

ettore: I only changed the SimplyE and SimplyE-R2dev targets since those are ones we really care about and to avoid doing this fixing over and over when we merge

ettore: so in other words 3.6.4 is only used for SimplyECardCreator and SimplyE-noDRM targets.

vladimirfedorov (approved):

ErnestFan (approved):

PR #1320, opened 2021-02-25
SIMPLY-3555 Button name is not updated on category screen after Remove action is performed

vladimirfedorov:
**What's this do?**
Fixes cell's state after a reserved book is returned. `Return` button changes back to `Reserve` after the book is returned.

https://jira.nypl.org/browse/SIMPLY-3555

Please follow the steps under Acceptance Criteria in the ticket. The issue is reproducible on iPhone devices.

NA

Yes

Yes

@vladimirfedorov


ErnestFan (approved):

Simplified/NYPLBookRegistry.m
```diff
@@ -469,10 +469,13 @@ - (void)updateAndRemoveBook:(NYPLBook *)book
...
       [self broadcastChange];
-      // Queue this up so it happens after the broadcast is done.
-      [[NSOperationQueue mainQueue] addOperationWithBlock:^{
-        [self performSynchronizedWithoutBroadcasting:^{
-          [self removeBookForIdentifier:book.identifier];
+      // Timer delays removing from identifiersToRecords to give interfaces enough time for update.
+      // NSOperationQueue's addOperationWithBlock doesn't provide enough time between broadcastChange and removeBookForIdentifier,
+      // as the result, cells don't update book information correctly.
+      __weak NYPLBookRegistry *weakSelf = self;
+      [NSTimer scheduledTimerWithTimeInterval:1 repeats:NO block:^(NSTimer * _Nonnull timer __unused) {
```
ettore, line 476: I really would prefer not to rely on timers with arbitrary time intervals. Waiting 1 second is pretty slow and weird from a user perspective, but more importantly is prone to break.

Did we identify the reason behind the delay? From the comment it seems a synchronization issue. Can we use a queue to synchronize all the interested events maybe? Or even a NSNotification would be better and not costly to implement. Just 2 random ideas without much thought.

That being said, if this (or other) bug exposes an architecture defect, let's discuss it. The BookRegistry class is pretty old, and so are the classes around it, and we are starting to think of ways to make the app more modular. We don't have to do this now, obviously, but these kind of issues are good documentation to have to justify a bigger reworking of the app in the future.

vladimirfedorov: > Waiting 1 second is pretty slow and weird from a user perspective, but more importantly is prone to break

Users don't see the delay, the change from [Return] to [Reserve] happens instantly, after `[self broadcastChange];` issues a notification at line 471. The added code delays removing the book from the registry.

There are two lines that directly affect views behaviour:

- Line 470 `self.identifiersToRecords[book.identifier] = [[record recordWithBook:book] recordWithState:NYPLBookStateUnregistered];` — sets state `NYPLBookStateUnregistered` to the book in the registry; views should be able to read the changed state;
- Line 475 (removed) `[self removeBookForIdentifier:book.identifier];` — effectively deletes `self.identifiersToRecords[book.identifier]`, after that the book won't be found in the registry. 

If any view receives and processes the notification after line 475 (removed) is executed, it won't find the record for the book with book.identifier and won't update its state accordingly.

Taking the complexity of interaction between all parts of the code into account (the user taps [Return], answers "Are you sure?" question, the app interacts with the library to cancel the reservation, deletes all local data, and only after that we get to this point) adding some time before we delete the book record form NYPLBookRegistry to allow notifications to do the job was the least invasive and adding no potential bugs, as we know that the code works.

Simplified/NYPLBookRegistry.m
```diff
@@ -469,10 +469,13 @@ - (void)updateAndRemoveBook:(NYPLBook *)book
     if(record) {
       self.identifiersToRecords[book.identifier] = [[record recordWithBook:book] recordWithState:NYPLBookStateUnregistered];
       [self broadcastChange];
-      // Queue this up so it happens after the broadcast is done.
-      [[NSOperationQueue mainQueue] addOperationWithBlock:^{
-        [self performSynchronizedWithoutBroadcasting:^{
-          [self removeBookForIdentifier:book.identifier];
+      // Timer delays removing from identifiersToRecords to give interfaces enough time for update.
```
ettore, line 472: I think I misunderstood what `interfaces` meant here... reading your other comment below you're referring to views, correct?

vladimirfedorov: yes, right, = "to give views enough time"

Simplified/NYPLBookRegistry.m
ettore: I see -- so if I understand correctly the desired order of execution is:
1. set `self.identifiersToRecords` for the `book`to "Unregistered"
2. `broadcastChange` 
3. views and all other interested parties update their state. These may need to be able to access the record for `book` in `identifiersToRecords` before it is removed.
4. finally remove the book from the BookRegistry entirely.

we have control over (1) and (2) but we can't know who is involved in (3). For all we know from here this can involve a really complex operation (or one that requires user input) that takes longer than 1 second. On the other hand, the BookRegistry really shouldn't need to care about what other parties need to do but it could/should give them more info on what's happening.

I have 2 questions since I'm not familiar with this code:
1. in step (3) above, do those views/3rd parties actually need to find the "Unregistered" record for the `book`? Or instead, if the book record is not found in `identifiersToRecords`, can they assume it's "Unregistered" and react the same way?
2. otherwise, could we pass the record we're about to remove with the notification? that way the 3rd parties listening for the notification have all the information they need and can update their state. Do you think this is something doable?

I don't know if these alternatives are possible solutions... I'm just trying to see if we can start decoupling these pieces and remove implicit assumptions.

vladimirfedorov: > For all we know from here this can involve a really complex operation (or one that requires user input) that takes longer than 1 second.

No, not at all. After "broadcastChange" book's cell change buttons instantly, as soon as it receives update notification. Not fast enough to queue a block of code that deletes the book from the registry right after we send the notification though.

> I don't know if these alternatives are possible solutions...

Currently all related views communicate state change through registry using book identifiers, and changing this seems to be quite a global task. For (1) "unregistered" state leads views to go through the book's defaultAquisition availability, which, while testing, was returning "Holding" status, meaning the button remained "Return"; for (2) we _could_ do this, but then the app will define buttons status through the book's defaultAquisition availability sa in (1). In this case we'll create some sort of an exception from how each book cell gets book's status, doesn't seem to be a good idea to me.

ettore: > > For all we know from here this can involve a really complex operation (or one that requires user input) that takes longer than 1 second.
> 
> No, not at all. 

right, sorry, yes it is true that the current code that responds to the notification is pretty fast, but I was speaking in general. BookRegistry shouldn't need to care about what is being done in the notification listeners, but in this case we are depending on them being done in 1 second.

> 
> > I don't know if these alternatives are possible solutions...
> 
> Currently all related views communicate state change through registry using book identifiers, and changing this seems to be quite a global task. For (1) "unregistered" state leads views to go through the book's defaultAquisition availability, which, while testing, was returning "Holding" status, meaning the button remained "Return"; 

what happens instead if there's no record at all in the registry?

>for (2) we _could_ do this, but then the app will define buttons status through the book's defaultAquisition availability sa in (1). In this case we'll create some sort of an exception from how each book cell gets book's status, doesn't seem to be a good idea to me.

let me rephrase: if I understand correctly, we need to leave a record in the registry with in the Unregistered state, wait for all notification listeners to be done because they need to see that "unregistered" record before it's removed from the registry: is this correct? If so, I was thinking, what if we remove the record immediately from the BookRegistry but just pass the deleted record? The receiving listener will then have (conceptually) the same state and is in the condition to do the same things it's doing now. Obviously this will require changes, but my question was more about how hard this is to implement.

vladimirfedorov: > what if we remove the record immediately from the BookRegistry but just pass the deleted record

This notification refreshes collection view controller with book cells, the desired receiver—the cell with the book—will be refreshed, so we'll need some other method to pass the change. So far looking into the class I think the question should probably be why do we need to delete book records from the registry at all, as the book remains visible and still exists in the feed. I'll take some more time tomorrow to dig deeper into the registry and book statuses then.

ettore: > why do we need to delete book records from the registry at all, as the book remains visible and still exists in the feed.

that's a very good point, and I agree. I don't know if there's any usage of the registry that relies on a book _not_ being in the registry, but it would be great if we can simplify that logic.

vladimirfedorov: After testing I removed `removeBookIdentifier` call from `updateAndRemoveBook` in `NYPLBookRegistry`. I moved `removePinnedThumbnailImageForBookIdentifier` to the function as this function actually does some action, removing image files from the app's directory. Other than that, I don't see the use for `removeBookIdentifier` when we return a reserved book: it doesn't remove book data (`NYPLMyBookDownloadCenter` does) and OPDS feed correctly updates book record when needed. The code correctly works on iPhones (where the problem was originally noticed) and iPads now.

ettore: Great! my only remaining question is whether there's any logic that relies on the book _not_ being in the registry, since now we do not remove it anymore.

vladimirfedorov: `NYPLBookDetailViewController` shows book information from local registry or library feed when the book is not available locally. This is a side effect of not deleting the book—it remains in the registry until updated by `syncResettingCache`. It doesn't affect book presentation or getting/borrowing though: registry book is updated from the feed when the app syncs the feed.

ettore: clarified in person that it should read: "shows book information from local registry, OR library feed when the book is not available locally" which makes sense.

ettore (approved):

PR #1322, opened 2021-02-27
SIMPLY-3494 Implement new Age Gate design

ErnestFan:
**What's this do?**
Implement new Age Gate design with dark mode support
Refactor usage of NYPLAgeCheck singleton into AccountsManager

[SIMPLY-3494](https://jira.nypl.org/browse/SIMPLY-3494?filter=-1)

Uninstall app and reinstall app
Select `Find a library later`
The `Age Verification` screen should show up
Test with different birth year

N/A

Will bump the build number after Vladimir's PR is merged

N/A

@ErnestFan


ErnestFan:
Forgot about the unit tests, will work on that on Monday and let you know when the PR is ready @ettore @vladimirfedorov

ettore:
> Forgot about the unit tests, will work on that on Monday and let you know when the PR is ready @ettore @vladimirfedorov

no worries, they can also be in a separate PR if you prefer.

question about testing: the age gate can also be triggered from the Settings tab, correct?

OpenEbooks/OEImages.xcassets/ArrowDown.imageset/Contents.json
```diff
@@ -0,0 +1,26 @@
+{
```
ettore, line 1: just wondering, do we actually need this for OE? because I thought an account on OE is hard-bound to a specific age group? I may be wrong though.

ErnestFan: I just add it to OE since the NYPLAgeCheck is included in OE build, I can remove them if you think it's unnecessary

And yes, the age gate can be trigger from the Settings tab on iOS 13+, I will leave a detailed testing steps on the ticket when it's ready for QA

Simplified/NYPLSettingsAccountDetailViewController.m
```diff
@@ -538,26 +538,24 @@ - (void)tableView:(__attribute__((unused)) UITableView *)tableView
   
   switch(cellKind) {
     case CellKindAgeCheck: {
+      [self.tableView deselectRowAtIndexPath:indexPath animated:YES];
       UITableViewCell *cell = [tableView cellForRowAtIndexPath:indexPath];
-      if (self.selectedAccount.details.userAboveAgeLimit == YES) {
-        [self confirmAgeChange:^(BOOL under13) {
-          if (under13) {
-            cell.accessoryView = [[UIImageView alloc] initWithImage:[UIImage imageNamed:@"CheckboxOff"]];
```
ettore, line 545: do we still use this `CheckboxOff` and `CheckboxOn` images or can we chuck them?

ettore (changes requested):

Simplified.xcodeproj/project.pbxproj
```diff
@@ -2782,6 +2788,15 @@
 			name = Holds;
 			sourceTree = "<group>";
 		};
+		17631AE725E4885E006079C4 /* NYPLAgeCheck */ = {
```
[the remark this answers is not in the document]
ErnestFan: Just for clarity, should I remove the NYPL prefix from all the age check classes and protocols? or just the directories?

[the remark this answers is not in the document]
ettore: just the directory and project group.

Simplified/Accounts/NYPLAgeCheck/NYPLAgeCheckViewController.swift
```diff
@@ -0,0 +1,194 @@
...
+    inputTextField.leadingAnchor.constraint(equalTo: view.leadingAnchor, constant: 50).isActive = true
+    inputTextField.trailingAnchor.constraint(equalTo: view.trailingAnchor, constant: -50).isActive = true
+    inputTextField.heightAnchor.constraint(equalToConstant: textFieldHeight).isActive = true
+    view.bringSubviewToFront(inputTextField)
+    
+    titleLabel.centerXAnchor.constraint(equalTo: view.centerXAnchor).isActive = true
+    titleLabel.bottomAnchor.constraint(equalTo: view.centerYAnchor, constant: -30).isActive = true
+    titleLabel.widthAnchor.constraint(equalTo: inputTextField.widthAnchor).isActive = true
+    titleLabel.heightAnchor.constraint(equalToConstant: textFieldHeight).isActive = true
+    view.bringSubviewToFront(titleLabel)
```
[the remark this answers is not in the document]
ErnestFan: There is one on the `UIToolbar` when the picker view shows up.
It looks like an iOS 13 issue from [this link](https://stackoverflow.com/questions/54284029/uitoolbar-with-uibarbuttonitem-layoutconstraint-issue).
I have implemented the fix mentioned in the link above and there is no auto layout warning now.

Simplified/Accounts/NYPLAgeCheck/NYPLAgeCheck.swift
```diff
@@ -0,0 +1,95 @@
...
+
+  // Members
+  let serialQueue = DispatchQueue(label: "\(Bundle.main.bundleIdentifier!).ageCheck")
+  var handlerList = [((Bool) -> ())]()
+  var isPresenting = false
+
+  func verifyCurrentAccountAgeRequirement(_ completion: ((Bool) -> ())?) -> Void
+  {
+    serialQueue.async {
+      let userAccount = NYPLUserAccount.sharedAccount()
```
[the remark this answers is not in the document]
ErnestFan: Switched to dependency injection

Simplified/Accounts/NYPLAgeCheck/NYPLAgeCheck.swift
```diff
@@ -0,0 +1,95 @@
...
+      guard let accountDetails = AccountsManager.shared.currentAccount?.details else {
```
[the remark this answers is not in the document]
ErnestFan: Switched to dependency injection

Simplified/Accounts/NYPLAgeCheck/NYPLAgeCheck.swift
```diff
@@ -0,0 +1,95 @@
...
+        completion?(false)
+        return
+      }
+      
+      if userAccount.needsAuth == true || accountDetails.userAboveAgeLimit {
+        completion?(true)
+        return
+      }
+      
+      if !accountDetails.userAboveAgeLimit && NYPLSettings.shared.userPresentedAgeCheck {
```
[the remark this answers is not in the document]
ErnestFan: Added a protocol named `NYPLAgeCheckChoiceStorage `.

I make it a parameter in the initializer and store it as a property in `NYPLAgeCheck` since we will be updating the `userPresentedAgeCheck ` property when AgeCheck completes/fails.

vladimirfedorov (approved):

OpenEbooks/OEImages.xcassets/ArrowDown.imageset/Contents.json
ettore: it there's no way in the OE code to trigger this code and resources, I think we should remove it. You could also avoid compilation of the code via the OPEN_EBOOKS / SIMPLYE preprocessor macros

ettore: on the other hand, if it complicates things too much, don't worry about it. Perhaps you can just avoid the resources inclusion.

Simplified/Accounts/AgeCheck/NYPLAgeCheck.swift
```diff
@@ -0,0 +1,115 @@
...
+  func ageCheckFailed()
+}
+
+@objc protocol NYPLAgeCheckVerification {
+  func verifyCurrentAccountAgeRequirement(userAccountProvider: NYPLUserAccountProvider,
+                                          currentLibraryAccountProvider: NYPLCurrentLibraryAccountProvider,
+                                          _ completion: ((Bool) -> ())?) -> Void
+}
+
+@objcMembers final class NYPLAgeCheck : NSObject, NYPLAgeCheckDelegate, NYPLAgeCheckVerification {
```
ettore, line 20: looking at this now the semantics of these 2 protocols are kind of unclear. 

`NYPLAgeCheckVerification` is the interface for what is actually determining the need for age verification. To be more swifty perhaps we should rename it to `NYPLAgeCheckVerifying`.

`NYPLAgeCheckDelegate` is the interface that defines the things required to perform age validation from the point of view of the component that has to enforce it (the VC in our case). So maybe we could rename it to `NYPLAgeCheckValidationDelegate`, since it's the VC who's delegating validation to `NYPLAgeCheck`?

Also curious to hear @singhraman4282 thoughts.

Simplified/Settings/NYPLSettings.swift
```diff
@@ -12,7 +12,11 @@ import Foundation
   var accountMainFeedURL: URL? { get set }
 }
 
-@objcMembers class NYPLSettings: NSObject, NYPLFeedURLProvider {
+@objc protocol NYPLAgeCheckChoiceStorage {
```
ettore, line 15: I would move this to the NYPLAgeCheck class, because that's where it's actually used. That way the AgeCheck api is all defined there.

ettore (commented):
Looks good, just a question about threading and some readability improvement suggestions

Simplified/NYPLSettingsAccountDetailViewController.m
```diff
@@ -538,26 +538,26 @@ - (void)tableView:(__attribute__((unused)) UITableView *)tableView
...
-            self.selectedAccount.details.userAboveAgeLimit = NO;
-            //Delete Books in My Books
-            [[NYPLMyBooksDownloadCenter sharedDownloadCenter] reset:self.selectedAccountId];
-            [[NYPLBookRegistry sharedRegistry] reset:self.selectedAccountId];
+      
+      if (!NYPLSettings.shared.userPresentedAgeCheck) {
+        __weak NYPLSettingsAccountDetailViewController *weakSelf = self;
+        [[[AccountsManager shared] ageCheck] verifyCurrentAccountAgeRequirementWithUserAccountProvider:self.businessLogic.userAccount
```
ErnestFan, line 546: The user could be signed into another library (eg. NYPL) and try to perform age check on SimplyE.
Using the account within `businessLogic` here handles such edge case.

Simplified/Accounts/AgeCheck/NYPLAgeCheck.swift
singhraman4282: I agree with @ettore here. I usually follow the swift style guide for protocol naming https://swift.org/documentation/api-design-guidelines/#protocols-describing-what-is-should-read-as-nouns. 

Protocols that describe a capability should be named using the suffixes able, ible, or ing. Therefore NYPLAgeCheckVerification should be changed to NYPLAgeCheckVerifying.

Also, I think NYPLAgeCheckDelegate can use the Should, Will, Did protocol pattern. So func ageCheckCompleted(_ birthYear: Int) can be changed to didCompleteAgeCheck(_ birthYear: Int) and ageCheckFailed() can be changed to didFailAgeCheck(_ error: Error?)

Simplified/Accounts/AgeCheck/NYPLAgeCheck.swift
```diff
@@ -0,0 +1,121 @@
...
+  var isPresenting = false
+  var ageCheckCompleted: Bool = false
+  let ageCheckChoiceStorage: NYPLAgeCheckChoiceStorage
+  
+  var minYear: Int = 1900
+  
+  var currentYear: Int = Calendar.current.component(.year, from: Date())
+  
+  var birthYearList: [Int] {
+    return Array(minYear...currentYear)
```
singhraman4282, line 39: With this, the user can use 2021 as year of birth. Is this intended?

Simplified/Accounts/AgeCheck/NYPLAgeCheck.swift
```diff
@@ -0,0 +1,121 @@
...
+  }
+
+  init(ageCheckChoiceStorage: NYPLAgeCheckChoiceStorage) {
+    self.ageCheckChoiceStorage = ageCheckChoiceStorage
+    
+    super.init()
+  }
```
singhraman4282, line 46: Is there a reason serialQueue, handlerList, isPresenting, & ageCheckChoiceStorage are not private?

Simplified/Accounts/AgeCheck/NYPLAgeCheck.swift
```diff
@@ -0,0 +1,121 @@
...
+  // Members
+  let serialQueue = DispatchQueue(label: "\(Bundle.main.bundleIdentifier!).ageCheck")
+  var handlerList = [((Bool) -> ())]()
+  var isPresenting = false
+  var ageCheckCompleted: Bool = false
+  let ageCheckChoiceStorage: NYPLAgeCheckChoiceStorage
+  
+  var minYear: Int = 1900
+  
+  var currentYear: Int = Calendar.current.component(.year, from: Date())
```
singhraman4282, line 36: since minYear, currentYear are get only and we're not using them as computed properties, and birthYear does not change at all with a change in state, does it make more sense to do this instead?
`// NYPLAgeCheckDelegate
    let minYear: Int
    let currentYear: Int
    let birthYearList: [Int]

    init(ageCheckChoiceStorage: NYPLAgeCheckChoiceStorage) {
        self.ageCheckChoiceStorage = ageCheckChoiceStorage
        minYear = 1900
        currentYear = Calendar.current.component(.year, from: Date())
        birthYearList = Array(minYear...currentYear)
        super.init()
    }`

Even though the compiler is smart enough to know that a variable is actually a constant, it kinda makes it easier on the eyes. Just a _weak_ suggestion.

Simplified/Accounts/AgeCheck/NYPLAgeCheckViewController.swift
```diff
@@ -0,0 +1,198 @@
...
+    
+    navigationItem.setRightBarButton(rightBarButtonItem, animated: true)
+    
+    inputTextField.translatesAutoresizingMaskIntoConstraints = false
+    titleLabel.translatesAutoresizingMaskIntoConstraints = false
+    
+    view.addSubview(inputTextField)
+    view.addSubview(titleLabel)
+    
+    inputTextField.centerXAnchor.constraint(equalTo: view.centerXAnchor).isActive = true
```
singhraman4282, line 80: we're already setting leading and trailing constraints at line 82 so we don't need a center constraint

Simplified/Accounts/AgeCheck/NYPLAgeCheck.swift
ettore: those should all be `lets` if they never change, that's definitely a good suggestion -- nice catch. 

Initializing them in the initializer vs inline is usually cosmetic, but in this case it allows for clearer behavior for `birthYearList` so I'm in favor of this.

Simplified/Accounts/AgeCheck/NYPLAgeCheck.swift
ErnestFan: There isn't a requirement on the upper bound of birth year, so I just use the current year.

Simplified/Accounts/AgeCheck/NYPLAgeCheck.swift
ErnestFan: Updated the naming of the functions

singhraman4282 (approved):

Simplified/Accounts/AgeCheck/NYPLAgeCheck.swift
```diff
@@ -0,0 +1,120 @@
...
+    currentYear = Calendar.current.component(.year, from: Date())
+    birthYearList = Array(minYear...currentYear)
+    
+    super.init()
+  }
+  
+  func verifyCurrentAccountAgeRequirement(userAccountProvider: NYPLUserAccountProvider,
+                                          currentLibraryAccountProvider: NYPLCurrentLibraryAccountProvider,
+                                          completion: ((Bool) -> ())?) {
+    serialQueue.async {
```
ettore, line 50: not sure how I missed this but there's a big retain cycle here:

self->serialQueue->closure -> self

ettore (approved):
approving pending the fix on the last retain cycle

PR #1327, opened 2021-03-10
SIMPLY-2609 Restore last read position before opening R2

ettore:
**What's this do?**
- It makes the opening of a book in R2 more reliable by first syncing the reading progress, then opening the R2 VC for presenting the book. Syncing is now delegated to the `ReaderModule`, which is a class that is book-format agnostic.
- I extracted the restore-reading-progress logic from the BookmarksBusinessLogic class and split it in 2 separate classes: one for syncing (`NYPLLastReadPositionSynchronizer`) and one for posting progress updates (`NYPLLastReadPositionPoster`). 
- I re-orged and documented the R2 files to identify the areas of concern for each class because I couldn't understand what anything was responsible for anymore. The new folders/groups hopefully make sense.
- I also streamlined various R2 classes apis, removing unneeded params, etc. 

The commit history also provides additional context.

https://jira.nypl.org/browse/SIMPLY-2609

see ticket.

This requires integration of the [patch submitted to the r2-navigator-swift](https://github.com/readium/r2-navigator-swift/pull/154) repo. I will do that in a separate PR because it will require some carthage wrangling.

not yet (see above)

yes

@ettore


singhraman4282 (approved):

Simplified.xcodeproj/project.pbxproj
```diff
@@ -6076,7 +6133,7 @@
 				CLANG_WARN_NON_LITERAL_NULL_CONVERSION = YES_ERROR;
 				CODE_SIGN_ENTITLEMENTS = Simplified/SimplyE.entitlements;
 				CODE_SIGN_IDENTITY = "iPhone Distribution";
-				CURRENT_PROJECT_VERSION = 9;
+				CURRENT_PROJECT_VERSION = 13;
```
ErnestFan, line 6136: Is there a reason this is 13 but the others are 10?

ettore: I think it's because that's the build number of the R2dev target, which is the one I was using. In my next PR where I integrate the changes for r2-navigator and create an actual build, I will straighten those up.

ErnestFan (approved):

vladimirfedorov (approved):

PR #1328, opened 2021-03-10
Added logging for better catching the condition where characterSet would be nil

singhraman4282:
**What's this do?**
Logs the book name which would cause the crash in https://jira.nypl.org/browse/SIMPLY-3578

https://jira.nypl.org/browse/SIMPLY-3578

Try to reproduce the crash on a device running iOS 14.5 beta.

n/a

Yes.

n/a

n/a


ettore:
the Open eBooks unit tests are failing, do they build correctly on your local machine?

In any case, since this is just a change for debugging an issue, I don't think we need a PR for it. I'd rather temporarily change the GitHub Actions workflow that builds and submits a binary to Firebase instead, and push those changes to a separate commit in your branch, which you can later nuke once QA is done. The workflow files in question are:

1. [`check-build-number.yml`](https://github.com/NYPL-Simplified/Simplified-iOS/blob/594889569caedcd986a5bb2dff4acde93504baf3/.github/workflows/check-build-number.yml#L7): add your branch name so this is triggered. This workflow checks if a build with the same version already exists on [iOS-binaries](https://github.com/NYPL-Simplified/iOS-binaries)
2. [`archive-and-upload.yml`](https://github.com/NYPL-Simplified/Simplified-iOS/blob/594889569caedcd986a5bb2dff4acde93504baf3/.github/workflows/archive-and-upload.yml#L6): remove this entirely or add a condition to match your branch here. This workflow actually archives the app and pushes to iOS-binaries.

ErnestFan (approved):

vladimirfedorov:
It may be a good idea to log actual character values passed to `indexOf:inString:` as well, as Crashalytics points to this functions and it seems we are treating `unichar`s (16-bit) from `characterAtIndex:` on line NYPLTenPrintCoverView.m:463 as `char`s.

vladimirfedorov (approved):

PR #1331, opened 2021-03-12
SIMPLY-3578: Fixed crashes on iOS 14.5 beta when generating cover from book with non ASCII characters

singhraman4282:
**What's this do?**
Fixes crashes on iOS 14.5 beta when generating cover from book with non ASCII characters

https://jira.nypl.org/browse/SIMPLY-3578

The original crash was found on iOS devices running iOS 14.5 beta while displaying books with no covers that have accented letters in their title (https://cdncms.fonts.net/images/de7722aa7a6beee0/D.Accents.jpg). To test the fix, try viewing such books in the list on a device running iOS 14.5 beta. The app should not crash while viewing those books.

n/a

Yes

n/a

Raman Singh tested on iOS 14.5 beta and iOS 13.7

### DIFF
--- a/Simplified.xcodeproj/project.pbxproj
+++ b/Simplified.xcodeproj/project.pbxproj
@@ -1631,6 +1631,13 @@
 			remoteGlobalIDString = A823D80C192BABA400B55DE2;
 			remoteInfo = SimplyE;
 		};
+		59D9CC1B25FFDE1B00C89B81 /* PBXContainerItemProxy */ = {
+			isa = PBXContainerItemProxy;
+			containerPortal = 1112A8191A322C53002B8CC1 /* TenPrintCover.xcodeproj */;
+			proxyType = 2;
+			remoteGlobalIDString = 59D9CBF525FFDBC500C89B81;
+			remoteInfo = TenPrintCoverTests;
+		};
 		5A569A2B1B8351C6003B5B61 /* PBXContainerItemProxy */ = {
 			isa = PBXContainerItemProxy;
 			containerPortal = 5A569A261B8351C6003B5B61 /* ADEPT.xcodeproj */;
@@ -2637,6 +2644,7 @@
 			isa = PBXGroup;
 			children = (
 				1112A81F1A322C53002B8CC1 /* libTenPrintCover.a */,
+				59D9CC1C25FFDE1B00C89B81 /* TenPrintCoverTests.xctest */,
 			);
 			name = Products;
 			sourceTree = "<group>";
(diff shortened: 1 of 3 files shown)


SimplifiedTests/SpecialChars.json
```diff
@@ -0,0 +1 @@
+†¤¶§†!"#$%&'()*+,-./0123456789:;<=>?@ABCDEFGHIJKLMNOPQRSTUVWXYZ[\]^_`abcdefghijklmnopqrstuvwxyz{|}~ÇüéâäàåçêëèïîìæÆôöòûùÿ¢£¥PƒáíóúñÑ,,¿¬½¼¡«»¦ßµ±°•·²€„…†‡ˆ‰Š‹Œ‘’“”–—˜™š›œŸ¨©®¯³´¸¹¾ÀÁÂÃÄÅÈÉÊËÌÍÎÏÐÒÓÔÕÖ×ØÙÚÛÜÝÞãðõ÷øüýþ
```
ettore, line 1: this file is not a json, can you just change the extension to (e.g.) `txt`?

ettore (commented):
thorough unit tests, thanks -- 

I think we are missing a PR for `tenprintcover` though right? also, currently SimplyE points to release `1.0.1` for `tenprintcover`, which will need to be updated after we merge that PR there.

SimplifiedTests/SpecialChars.json
singhraman4282: Great catch. Done.

ErnestFan (approved):

Simplified.xcodeproj/project.pbxproj
```diff
@@ -6551,6 +6566,7 @@
 				CODE_SIGN_ENTITLEMENTS = Simplified/SimplyE.entitlements;
 				CODE_SIGN_IDENTITY = "iPhone Developer";
 				CURRENT_PROJECT_VERSION = 7;
+				CURRENT_PROJECT_VERSION = 6;
```
ettore, line 6569: I think Xcode messed up here :) can you remove one of these?

ettore (approved):
GitHub is showing the updates from TenPrintCover in a weird way (usually it just shows the updated commit on the sub repo) but things do seem normal on your end, so let's monitor the build log and see if the sub-repo changes are being picked up correctly.

vladimirfedorov (approved):

ettore:
> let's monitor the build log and see if the sub-repo changes are being picked up correctly.

I do see your new file that you added being compiled here, so we should be good. Please also take a look though: https://github.com/NYPL-Simplified/Simplified-iOS/runs/2124108615?check_suite_focus=true

Simplified.xcodeproj/project.pbxproj
singhraman4282: Fixed

PR #1333, opened 2021-03-12
Merge feature Content Protection to develop

ettore:
**What's this do?**
Upgrades the way R2 handles DRM to using Content Protection. 

Since development on the CP branch is complete, we might as well merge it and make out lives a bit easier. R2 is still disabled by default.

All this code has already been reviewed but @ErnestFan @vladimirfedorov @singhraman4282 if you do a quick spot check on this to verify nothing weird is here that would be useful. The merge completed with no conflicts.
Official QA for this integration will be addressed via a regression.

n/a

yes

yes

@ettore did a quick smoke test.


Simplified/Reader2/BusinessLogic/NYPLLastReadPositionPoster.swift
```diff
@@ -0,0 +1,97 @@
...
+      self.queuedReadPosition = locationString
+
+      if Date() > self.lastReadPositionUploadDate.addingTimeInterval(NYPLLastReadPositionPoster.throttlingInterval) {
+        self.postQueuedReadPosition()
+      }
+    }
+  }
+
+  private func postQueuedReadPosition() {
+    if self.queuedReadPosition != "" {
```
singhraman4282, line 83: Minor (weak) suggestion

`if !self.queuedReadPosition.isEmpty`

Simplified/Reader2/ReaderPresentation/NYPLRootTabBarController+R2.swift
```diff
@@ -0,0 +1,44 @@
...
+        
+      case .failure(let error):
+        // .failure is retured for an error raised while trying to unlock publication
+        // error is supposed to be visible to users, it is defined by ContentProtection error property
+        NYPLErrorLogger.logError(error, summary: "Error accessing book resources", metadata: [
+          "book": book.loggableDictionary
+        ])
+        let alertController = NYPLAlertUtils.alert(title: "Content Protection Error", message: error.localizedDescription)
+        NYPLAlertUtils.presentFromViewControllerOrNil(alertController: alertController, viewController: self, animated: true, completion: nil)
+      }
```
singhraman4282, line 40: There's some repetition here which can be fixed

```
func presentBook(_ book: NYPLBook) {
    guard let libraryService = r2Owner?.libraryService, let readerModule = r2Owner?.readerModule else {
      return
    }

    libraryService.openBook(book, sender: self) { [weak self] result in
      guard let navVC = self?.selectedViewController as? UINavigationController else {
        preconditionFailure("No navigation controller, unable to present reader")
      }
        
      let alertTitle: String?
      let alertMessage: String?
        
      switch result {
      case .success(let publication):
        readerModule.presentPublication(publication, book: book, in: navVC)
      case .cancelled:
        // .cancelled is returned when publication has restricted access to its resources and can't be rendered
        alertTitle = "ReaderViewControllerCorruptTitle"
        alertMessage = "ReaderViewControllerCorruptMessage"
        
      case .failure(let error):
        // .failure is retured for an error raised while trying to unlock publication
        // error is supposed to be visible to users, it is defined by ContentProtection error property
        
        alertTitle = "Content Protection Error"
        alertMessage = error.localizedDescription
      }
        
      if let alertTitle = alertTitle, let alertMessage = alertMessage {
        NYPLErrorLogger.logError(nil,
                                 summary: "Error accessing book resources",
                                 metadata: ["book": book.loggableDictionary])
        let alertController = NYPLAlertUtils.alert(title: alertTitle,
                                                   message: alertMessage)
        
        NYPLAlertUtils.presentFromViewControllerOrNil(alertController: alertController,
                                                      viewController: self,
                                                      animated: true,
                                                      completion: nil)
            
      }
        
    }
```

ErnestFan (approved):

vladimirfedorov (approved):

Simplified/Reader2/ReaderPresentation/NYPLRootTabBarController+R2.swift
ettore: that's possible, although this is just and integration PR, so not really about code changes, but just to verify if there's any glaring merge oddities/artifacts.

singhraman4282 (approved):

PR #1334, opened 2021-03-13
SIMPLY-3596 Implement unit tests and mocks for AgeGate

ErnestFan:
**What's this do?**
Add unit tests and mocks for AgeGate

[SIMPLY-3596](https://jira.nypl.org/browse/SIMPLY-3596)

Run the unit tests

N/A

N/A

N/A

@ErnestFan


SimplifiedTests/Mocks/NYPLAgeCheckChoiceStorageMock.swift
```diff
@@ -0,0 +1,19 @@
+//
+//  NYPLAgeCheckChoiceStorage.swift
```
singhraman4282, line 2: NYPLAgeCheckChoiceStorageMock

SimplifiedTests/NYPLAgeCheckTests.swift
```diff
@@ -0,0 +1,129 @@
...
+  // This way we can set whatever value we need in each tests, and no need to worry about the default value on different machine
+  var defaultUserAboveAgeLimit: Bool!
+  
+  // Use expectation to test result within closure
+  var expectation: XCTestExpectation!
+  
+  override func setUpWithError() throws {
+    try super.setUpWithError()
+    
+    libraryAccountMock = NYPLLibraryAccountMock()
```
singhraman4282, line 34: We can delete libraryAccountMock since it's not used anywhere in the below tests

SimplifiedTests/Mocks/NYPLAgeCheckChoiceStorageMock.swift
ErnestFan: Nice catch!

SimplifiedTests/NYPLAgeCheckTests.swift
ErnestFan: Nice catch!

SimplifiedTests/NYPLAgeCheckTests.swift
```diff
@@ -0,0 +1,127 @@
...
+  func testAgeCheckFailed() throws {
+    // Use an inverted expectation to make sure the completion closure is not executed
+    self.expectation.isInverted = true
+    ageCheck.verifyCurrentAccountAgeRequirement(userAccountProvider: userAccountProviderMock,
+                                                currentLibraryAccountProvider: simplyeLibraryAccountProviderMock) { [weak self] (aboveAgeLimit) in
+      self?.expectation.fulfill()
+    }
+    
+    ageCheck.didFailAgeCheck()
+    waitForExpectations(timeout: 3, handler: nil)
```
ettore, line 123: nice use of inverted expectations! question, does this timeout then mean that this test will always take 3 seconds to run if it succeeds? If so just wondering if we can lower it a bit.

ettore (approved):

vladimirfedorov (approved):

SimplifiedTests/NYPLAgeCheckTests.swift
ErnestFan: It will succeed if the expectation is not fulfilled by the end of the timeout. Sure I can lower it, nothing here really takes time to process so it should be fine.

singhraman4282 (approved):

PR #1335, opened 2021-03-16
SIMPLY-3511: Use Network Executor for reporting errors with a book

singhraman4282:
**What's this do?**
Uses Network Executor for reporting errors with a book

https://jira.nypl.org/browse/SIMPLY-3511

Cannot be QA tested

n/a

No

n/a

Raman Singh verified it

### DIFF
--- a/Simplified/NYPLBookDetailViewController.m
+++ b/Simplified/NYPLBookDetailViewController.m
@@ -19,6 +19,7 @@
 
 @property (nonatomic) NYPLBook *book;
 @property (nonatomic) NYPLBookDetailView *bookDetailView;
+@property (nonatomic) NYPLNetworkExecutor *executor;
 
 -(void)didCacheProblemDocument;
 
@@ -36,6 +37,11 @@
   }
 
   self.book = book;
+    
+  self.executor = [[NYPLNetworkExecutor alloc]
+                   initWithCredentialsProvider:nil
+                   cachingStrategy:NYPLCachingStrategyEphemeral
+                   delegateQueue:nil];
 
   self.title = book.title;
   UILabel *label = [[UILabel alloc] init];
@@ -193,8 +199,11 @@
 {
   NSURL *reportURL = problemReportViewController.book.reportURL;
   if (reportURL) {
-    NSURLRequest *r = [NSURLRequest postRequestWithProblemDocument:@{@"type":type} url:reportURL];
-    [[NYPLSession sharedSession] uploadWithRequest:r completionHandler:nil];
+    NSURLRequest *request = [NSURLRequest
+                             postRequestWithProblemDocument:@{@"type":type}
+                             url:reportURL];
+    
+    [self.executor POST:request completion:nil];
   }
   if (problemReportViewController.navigationController) {
     [problemReportViewController.navigationController popViewControllerAnimated:YES];
--- a/Simplified/Network/NYPLNetworkExecutor.swift
+++ b/Simplified/Network/NYPLNetworkExecutor.swift
@@ -46,7 +46,7 @@ protocol NYPLRequestExecuting {
   /// - Parameter credentialsProvider: The object responsible with providing cretdentials
   /// - Parameter cachingStrategy: The strategy to cache responses with.
   /// - Parameter delegateQueue: The queue where callbacks will be called.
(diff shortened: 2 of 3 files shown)


ErnestFan (approved):

Simplified/NYPLBookDetailViewController.m
```diff
@@ -193,8 +193,10 @@ - (void)problemReportViewController:(NYPLProblemReportViewController *)problemRe
 {
   NSURL *reportURL = problemReportViewController.book.reportURL;
   if (reportURL) {
-    NSURLRequest *r = [NSURLRequest postRequestWithProblemDocument:@{@"type":type} url:reportURL];
-    [[NYPLSession sharedSession] uploadWithRequest:r completionHandler:nil];
+    NSURLRequest *request = [NSURLRequest
+                             postRequestWithProblemDocument:@{@"type":type}
+                             url:reportURL];
+    [NYPLNetworkExecutor.shared POST:request completion:nil];
```
ettore, line 199: as we discussed in private, can you create a new executor instance in the initializer  and save that as a "private" property? The `shared` instance is just a convenience (mainly to reuse the same caching storage) but here it makes sense to use `.ephemeral` caching:
```
let x = NYPLNetworkExecutor(cachingStrategy: .ephemeral)
```

ettore (changes requested):

vladimirfedorov (approved):

Simplified/Network/NYPLNetworkExecutor.swift
```diff
@@ -219,4 +219,30 @@ extension NYPLNetworkExecutor {
     }
     return executeRequest(req, completion: completionWrapper)
   }
+    
+  /// Performs a POST request using the specified request
+  /// - Parameters:
+  ///   - request: Request to be posted..
+  ///   - completion: Always called when the resource is either fetched from
```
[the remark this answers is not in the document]
singhraman4282: Great catch. Fixed. Who would had thought copy pasting the documentation could lead to such an error 😬.

Simplified/NYPLBookDetailViewController.m
singhraman4282: Sorry I forgot about doing that. Done.

Simplified/NYPLBookDetailViewController.m
```diff
@@ -193,8 +193,15 @@ - (void)problemReportViewController:(NYPLProblemReportViewController *)problemRe
 {
   NSURL *reportURL = problemReportViewController.book.reportURL;
   if (reportURL) {
-    NSURLRequest *r = [NSURLRequest postRequestWithProblemDocument:@{@"type":type} url:reportURL];
-    [[NYPLSession sharedSession] uploadWithRequest:r completionHandler:nil];
+    NSURLRequest *request = [NSURLRequest
+                             postRequestWithProblemDocument:@{@"type":type}
+                             url:reportURL];
+    NYPLNetworkExecutor *executor = [[NYPLNetworkExecutor alloc]
```
ettore, line 199: I'm unsure if this can work reliably, because the `executor` instance goes out of scope right after the POST call. Unless there is something that binds the request to a global queue or other object, but I don't think our Executor stack does that.

A strongly retained property sounds more reliable to me.

Simplified/NYPLBookDetailViewController.m
```diff
@@ -19,6 +19,7 @@ @interface NYPLBookDetailViewController () <NYPLBookDetailViewDelegate, NYPLProb
 
 @property (nonatomic) NYPLBook *book;
 @property (nonatomic) NYPLBookDetailView *bookDetailView;
+@property (nonatomic, retain) NYPLNetworkExecutor *executor;
```
ettore, line 22: minor: `retain/`strong` is unneeded since it's the default (also I think `retain` is deprecated unless you're not using ARC?)

ettore (approved):

Simplified/NYPLBookDetailViewController.m
singhraman4282: Done

Simplified/NYPLBookDetailViewController.m
singhraman4282: Fixed

PR #1337, opened 2021-03-19
SIMPLY-3621 Remove duplication between BookButtonsView and BookDetailButtonsView

ettore:
**What's this do?**
NYPLBookDetailButtonsView was a superset of NYPLBookButtonsView, with the exception of the handling of the clock icon, which is done correctly by NYPLBookButtonsView. So this class removes the latter and does the appropriate refactors.

A positive side effect of this is that now the book cells show the Return button instead of needing to go to the book details page, which is something that we want for the Finder quick-wins project. 

It also moves in new files some protocols and enums that were inside NYPLBookButtonsView.h. So even if NYPLBookButtonsState.h and NYPLBookButtonsView.h appears new, they are just cut and paste. That's also why they are not swift.

I also re-orged all the boo-related sources into physical folders. This will make browsing the code base in the finder more manageable. I left `NYPLProblemReportViewController` out of this re-org because it is currently being worked on by @singhraman4282 in another PR, so I left it as it was to avoid conflicts.

https://jira.nypl.org/browse/SIMPLY-3621

I'll add steps in the ticket.

n/a

yes

yes

@ettore


ErnestFan (approved):

vladimirfedorov (approved):
Looks good

Simplified/Book/UI/NYPLBookDetailView.m
```diff
@@ -192,18 +193,23 @@ - (void)createBookDescriptionViews
...
-                                   initWithData:htmlData
-                                   options:@{NSDocumentTypeDocumentAttribute:
-                                               NSHTMLTextDocumentType}
-                                   documentAttributes:nil
-                                   error:&error];
-  if (error) {
-    NYPLLOG_F(@"Attributed string rendering error for %@ book description: %@",
-              [self.book loggableShortString], error);
+  NSAttributedString *attrString;
+  if (htmlData != nil) {
```
singhraman4282, line 197: if (htmlData)

singhraman4282 (approved):

Simplified/Book/UI/NYPLBookDetailView.m
ettore: rebased and fixed

PR #1338, opened 2021-03-19
[SIMPLY-3620] Fix AccountsManager completion handler logic bug for AgeGate

ErnestFan:
**What's this do?**
The reason why the AgeGate did not show up is because the associated `Account` in `AccountsManager` is being overwritten, and the `needsAgeCheck` check fails because the value is empty.
The overwrite on the Account happens since `loadCatalogs` is called twice and the first one with the nil completion handler is not added to the list. Therefore, the second `loadCatalogs` continue after seeing the list is empty.

[SIMPLY-3620](https://jira.nypl.org/browse/SIMPLY-3620)

See ticket

N/A

Yes, bumped build numbers

N/A

@ErnestFan


ettore:
> loadCatalogs is called twice and the first one with the nil completion handler is not added to the list. Therefore, the second loadCatalogs **_continue_** after seeing the list is empty.

sorry, I don't understand where we continue exactly here: can you point to the area of the code where not having the empty completion the code segues into the wrong flow?

I remember doing this (I mean not adding an empty completion handler) because it was causing some issues.... although after looking at the commit history I didn't write down what issue exactly (shame on me!), so hopefully I was just doing that because it seemed useless.

ErnestFan:
> > loadCatalogs is called twice and the first one with the nil completion handler is not added to the list. Therefore, the second loadCatalogs **_continue_** after seeing the list is empty.
> 
> sorry, I don't understand where we continue exactly here: can you point to the area of the code where not having the empty completion the code segues into the wrong flow?
> 
> I remember doing this (I mean not adding an empty completion handler) because it was causing some issues.... although after looking at the commit history I didn't write down what issue exactly (shame on me!), so hopefully I was just doing that because it seemed useless.

If you look into `loadCatalog(completion:)`, we have the line `guard !wasAlreadyLoading else { return }` to make sure `loadCatalog` only happens one at a time. However, if we skip adding a completion handler when it's nil (first loadCatalog), the list will be empty when the second `loadCatalog` happens and the guard statement will be useless.

ettore:
> If you look into `loadCatalog(completion:)`, we have the line `guard !wasAlreadyLoading else { return }` to make sure `loadCatalog` only happens one at a time. However, if we skip adding a completion handler when it's nil (first loadCatalog), the list will be empty when the second `loadCatalog` happens and the guard statement will be useless.

I see now. I think I did this because it was causing issues when the catalog didn't load for an error or when the URL was nil (during the app launch sequence). At that point the Try Again button will show up, and empty completion blocks I believe were preventing the flow to go through, but I don't recall well, sadly. After dismissing the age gate, did you notice any errors in loading the catalog? Or if you force an error, does the Try Again button show up, and does hitting that make it reload the page correctly?

ErnestFan:
> > If you look into `loadCatalog(completion:)`, we have the line `guard !wasAlreadyLoading else { return }` to make sure `loadCatalog` only happens one at a time. However, if we skip adding a completion handler when it's nil (first loadCatalog), the list will be empty when the second `loadCatalog` happens and the guard statement will be useless.
> 
> I see now. I think I did this because it was causing issues when the catalog didn't load for an error or when the URL was nil (during the app launch sequence). At that point the Try Again button will show up, and empty completion blocks I believe were preventing the flow to go through, but I don't recall well, sadly. After dismissing the age gate, did you notice any errors in loading the catalog? Or if you force an error, does the Try Again button show up, and does hitting that make it reload the page correctly?

I am not seeing any error after Age Verification.
Forcing an error when loadCatalog (with invalid URL) does trigger the `Try Again` button. Hitting the button again does trigger another request in NYPLNetworkExecutor and the catalog seems to load without any error, at least that's what I see from the debugger.

singhraman4282 (approved):

Simplified/Accounts/Library/AccountsManager.swift
```diff
@@ -145,9 +145,11 @@ let currentAccountIdentifierKey  = "NYPLCurrentAccountIdentifier"
         loadingCompletionHandlers[key] = [(Bool)->()]()
       }
       wasEmpty = loadingCompletionHandlers[key]!.isEmpty
-      if let handler = handler {
-        loadingCompletionHandlers[key]!.append(handler)
-      }
+      // On the previous implementation, we do not add the handler to the list if it is nil.
+      // If the first handler is nil, the second handler would find the list empty and continue loading
+      // And the same loading request would happen more than once, adding a empty completion block void such situation
+      let h: ((Bool) -> ()) = handler != nil ? handler! : { _ in }
```
ettore, line 151: minor: you can shorten this as `handler ?? { _ in }`

ettore (approved):

PR #1339, opened 2021-03-23
Reorg for files related to book models/UI, holds, OPDS, logging, settings, utilities: no functional changes

ettore:
**What's this do?**
This PR includes no functional changes.

It just moves files around in the repo. Folder structure is also replicated in Xcode groups. The only ungrouped files are now the R1 sources, which are deprecated.

There is one rename: NYPLBookAcquisitionPath to NYPLOPDSAcquisitionPath (since that file is about OPDS stuff).

Browsing files in the repo is painful because there are still a ton of sources in the main Simplified folder for many different responsibilities. This is hard to manage, so the point of this change is to make that easier.

As long as it builds we should be good, because there are no functional changes here.

n/a

n/a

n/a

@ettore


Simplified/Book/Models/NYPLBook+Additions.swift
```diff
@@ -0,0 +1,17 @@
...
+//  NYPLBook+Additions.swift
+//  SimplyE
+//
+//  Created by Ettore Pasquini on 7/9/20.
+//  Copyright © 2020 NYPL Labs. All rights reserved.
+//
+
+import Foundation
+
+extension NYPLBook {
```
ettore, line 11: NB this is not actually new of course. Previously it was in NYPLBook+Logging.swift.

singhraman4282 (approved):

ErnestFan (approved):

PR #1341, opened 2021-03-26
SIMPLY-3644 Refactor bookmarks creation from server and R2 locator

ettore:
**What's this do?**
- Adds the spec for the bookmarks format as a submodule so that in the (near) future we can write unit tests against it.
- Moves code that parses bookmarks from server response to a Factory class. Previously it was in the code submitting the GET bookmarks request (see `NYPLAnnotations.swift`)
- Also moves code that created a bookmark from an R2 location to the same Factory class.  Previously this was in the bookmarks business logic class

https://jira.nypl.org/browse/SIMPLY-3644
This is a first PR for the bookmarks refactor. I have a lot more code but it's starting to add up, and I didn't want to submit a huge PR.  

Verify bookmarks still work as intended. I will add more details in the ticket.

This is pretty safe to merge into `develop`. There are no functional changes: beside code relocation, there's only some minor improvements for readability.

yes

yes

@ettore


Simplified/Reader2/BusinessLogic/NYPLBookmarkFactory.swift
```diff
@@ -0,0 +1,149 @@
...
+    guard source == bookID else {
+      NYPLErrorLogger.logError(withCode: .bookmarkReadError,
+                               summary: "Got bookmark for a different book",
+                               metadata: [
+                                "requestedBookID": bookID,
+                                "serverAnnotation": annotation])
+      return nil
+    }
+
+    guard motivation.contains("bookmarking") else {
```
ettore, line 103: I am refactoring these hardcoded strings in a later PR

Simplified/Reader2/BusinessLogic/NYPLReaderBookmarksBusinessLogic.swift
```diff
@@ -83,78 +88,27 @@ class NYPLReaderBookmarksBusinessLogic: NSObject {
...
-      location: registryLoc?.locationString,
-      progressWithinChapter: chapterProgress,
-      progressWithinBook: totalProgress,
-      time: (bookmarkLoc.creationDate as NSDate).rfc3339String(),
-      device: drmDeviceID) else {
-        return nil
+    guard let bookmark =
+      bookmarksFactory.make(from: bookmarkLoc,
+                            usingBookRegistry: bookRegistry) else {
+                              //TODO: log error
```
ettore, line 94: this will also come in a later PR

ErnestFan (approved):

Simplified/Reader2/BusinessLogic/NYPLBookmarkFactory.swift
```diff
@@ -0,0 +1,149 @@
...
+      idref = idrefFromR2
+    }
+
+    let chapter: String?
+    if let locatorChapter = bookmarkLoc.locator.title {
+      chapter = locatorChapter
+    } else if let tocLink = publication.tableOfContents.first(withHREF: bookmarkLoc.locator.href) {
+      chapter = tocLink.title
+    } else {
+      chapter = nil
```
singhraman4282, line 67: I think we can remove `chapter = nil` here

singhraman4282 (approved):

Simplified/Reader2/BusinessLogic/NYPLBookmarkFactory.swift
ettore: actually it's giving me a compile error because `chapter` is a `let` constant, so it requires explicit initialization. If it was a `var` then the `chapter = nil` statement would be redundant.

I didn't know this detail so thank you! There seem to be a proposal to make `var` work like `let` does: https://forums.swift.org/t/prepitch-optional-variables-should-require-explicit-initialization/26077

PR #1343, opened 2021-03-29
SIMPLY-3644 Refactor bookmarks code using spec

ettore:
**What's this do?**
This PR is larger than I wanted but on the other hand it's mostly renames, reformatting for readability, some code extraction. So reviewing should go pretty quickly.

- introduce new **`NYPLBookmarkSpec`** that collects all the key names and constants that need to appear in the JSON sent to the server instead of having them hardcoded.
- remove code duplication for Reading position serialization by using NYPLBookmarkFactory class.
- functionally, parsing and serializing code remains the same so that it's still compatible with R1.
- make transparent the fact that the app internally writes to disk a dictionary representation of bookmarks as part of the BookRegistry logic. See **`NYPLBookmarkDictionaryRepresentation`**.
- minor refactor of networking layer to express the default timeout value.

https://jira.nypl.org/browse/SIMPLY-3644

This PR does not introduce functional changes in the app behavior, so a regression should be able to cover this refactoring.

This can be merged into develop, but I won't add it to the upcoming 3.6.4 release.

no

yes

@ettore


Simplified/NYPLReadiumViewSyncManager.m
```diff
@@ -111,23 +112,28 @@ - (void)syncReadingPositionForBook:(NSString *)bookID
 {
   __weak NYPLReadiumViewSyncManager *const weakSelf = self;
 
-  [NYPLAnnotations syncReadingPositionOfBook:bookID toURL:URL completionHandler:^(NSDictionary * _Nullable responseObject) {
```
ettore, line 114: the completion callback was using a loose object to pass along the CFI and a few other things, but the parsing of the api response was identical to parsing a bookmark. So we might as well parse a bookmark object and return that instead, and this way reuse the bookmark parsing code as-is.

Simplified/NYPLReadiumViewSyncManager.m
```diff
@@ -194,12 +200,13 @@ - (void)presentSyncReadingPositionAlertForResponse:(NSDictionary *)response
...
 
-                              NSString *contentCFI = locationDictionary[@"contentCFI"];
-                              if (!contentCFI) {
-                                contentCFI = @"";
-                              }
-                              dictionary[@"openPageRequest"] =
-                              @{@"idref": locationDictionary[@"idref"], @"elementCfi": contentCFI};
+                              NSString *contentCFI = locationDictionary[NYPLBookmarkDictionaryRepresentation.cfiKey];
+
+                              dictionary[@"openPageRequest"] = @{
```
ettore, line 205: I didn't bother about some of the keys because this whole file is R1 only and will be removed

Simplified/Reader2/Bookmarks/NYPLBookmarkFactory.swift
```diff
@@ -80,13 +81,25 @@ class NYPLBookmarkFactory {
...
 
+  class func make(fromServerBookmark annotation: [String: Any],
+                  bookID: String) -> NYPLReadiumBookmark? {
+    return make(fromServerAnnotation: annotation,
+                annotationType: .bookmark,
+                bookID: bookID)
+  }
+
   class func make(fromServerAnnotation annotation: [String: Any],
+                  annotationType: NYPLBookmarkSpec.Motivation,
```
ettore, line 92: this is now the main parsing function from server data for both reading position and explicit bookmarks

Simplified/Reader2/Bookmarks/NYPLBookmarkFactory.swift
```diff
@@ -100,47 +113,50 @@ class NYPLBookmarkFactory {
...
+      let body = annotation[NYPLBookmarkSpec.Body.key] as? [String: AnyObject],
+      let device = body[NYPLBookmarkSpec.Body.Device.key] as? String,
+      let time = body[NYPLBookmarkSpec.Body.Time.key] as? String,
 
-    guard let device = body["http://librarysimplified.org/terms/device"] as? String,
-      let time = body["http://librarysimplified.org/terms/time"] as? String,
+      // TODO: SIMPLY-3644 fix
       let progressWithinChapter = (body["http://librarysimplified.org/terms/progressWithinChapter"] as? NSNumber)?.floatValue,
-      let progressWithinBook = (body["http://librarysimplified.org/terms/progressWithinBook"] as? NSNumber)?.floatValue else {
+      let progressWithinBook = (body["http://librarysimplified.org/terms/progressWithinBook"] as? NSNumber)?.floatValue
```
ettore, line 127: these strings are R1 only format so I left them as is... they will be removed in SIMPLY-3655

Simplified/Reader2/Bookmarks/NYPLAnnotations.swift
```diff
@@ -216,133 +212,111 @@ import UIKit
...
-      }
-      Log.error(#file, "No Annotation Item found for this title.")
-      completionHandler(nil)
-      return
+
+      let readPos = items
+        .compactMap { NYPLBookmarkFactory.make(fromServerAnnotation: $0,
+                                               annotationType: .readingProgress,
+                                               bookID: bookID) }
+        .first
```
singhraman4282, line 263: we're creating n number of objects with compact map and are using only the first object. Using a for loop can help us avoid all this extra work.

e.g.
```
var readPos: T?
for item in items {
if let a = NYPLBookmarkFactory.make(fromServerAnnotation: item, annotationType: .readingProgress, bookID: bookID) {
readPos = a
}
```
If the items count is relatively small, then we can get away with this since it adds to code readability.

singhraman4282 (approved):

Simplified/Reader2/Bookmarks/NYPLAnnotations.swift
ettore: that's kind of what I had before (a for loop) but I changed it to be more functional for readability. But yes, typically there are not hundreds of bookmarks in a book. I can change it back to a `for` if you feel strongly about it though.

ErnestFan (approved):

vladimirfedorov (approved):

ettore:
sorry ignore the last commit I accidentally pushed. I meant to send it to a different branch.

ettore:
> sorry ignore the last commit I accidentally pushed. I meant to send it to a different branch.

fixed. The code in this PR is the same as what you already reviewed. Sorry for the confusion.

Simplified/Reader2/Bookmarks/NYPLAnnotations.swift
singhraman4282: For smaller computations it's ok. Given the amount of money people pay for an iOS device, I think it should be able to handle such small computations without any hiccups.

PR #1347, opened 2021-04-01
[WIP] Downlaod book content with Axixnow drm

singhraman4282:
**What's this do?**
Download book content with axisnow drm. As of now the code is a mess. It's working with a few tweaks here and there but it no position to be considered even close to the final code that will be submitted for the pr.


Simplified/MyBooks/NYPLMyBooksDownloadCenter.m
```diff
@@ -1429,6 +1452,69 @@ - (void)didIgnoreFulfillmentWithNoAuthorizationPresent
 
 #endif
 
+- (void)fulfillAxisNowLicense:(NSURL *)fileUrl
```
ettore, line 1455: it would be great if this could live outside of `NYPLMyBooksDownloadCenter`. NYPLMyBooksDownloadCenter is a huge obnoxious class that just keeps growing. To be fair, I don't know how feasible that is, and I don't want to derail you into a refactor, but I wanted to point it out. 

I also wonder if we can avoid having `NYPLMyBooksDownloadCenter` be the `AxisNowDelegate`. And again, the same feasibility disclaimer applies of course.

singhraman4282: Agreed. I am of the opinion that every new feature should act like a plug and play device. Getting it entangled in previously written code is a really bad idea.

Simplified/MyBooks/RSAManager.swift
```diff
@@ -0,0 +1,116 @@
...
+//  RSAManager.swift
+//  SimplyE
+//
+//  Created by Raman Singh on 2021-03-30.
+//  Copyright © 2021 NYPL Labs. All rights reserved.
+//
+
+import Foundation
+
+class RSAManager {
```
ettore, line 11: it's best to avoid "Manager" as a class suffix, since [it's an anti pattern](coding horror manager). "Managers" have the tendency to quickly become "do everything" types, and as they grow their semantics become fuzzy. 

In this case it seems what this class does is encrypting and decrypting using RSA system, so maybe NYPLRSACypher / NYPLRSACryptographer or something like that express its duties more in detail? And if it's specific to AxisNow probably the "Axis" prefix should be added too.

PR #1348, opened 2021-04-01
Added minimal change required to setup the app for AxisNow

singhraman4282:
**What's this do?**
Added minimal change required to setup the app for AxisNow

https://jira.nypl.org/browse/SIMPLY-3598

Can not be tested as of now

n/a

n/a

Not yet. Will do upon completion.

Raman Singh verified it. Will write tests and verify downloading and opening axis books.

### DIFF
--- a/Simplified/AxisService/AxisService.swift
+++ b/Simplified/AxisService/AxisService.swift
@@ -1,0 +1,43 @@
+//
+//  AxisService.swift
+//  Simplified
+//
+//  Created by Raman Singh on 2021-04-01.
+//  Copyright © 2021 NYPL Labs. All rights reserved.
+//
+
+import Foundation
+
+#if FEATURE_DRM_CONNECTOR && OPENEBOOKS
+
+@objc
+protocol NYPLBookDownloadBroadcasting {
+  func failDownloadWithAlert(forBook book: NYPLBook)
+  func replaceBook(_ book: NYPLBook,
+                   withFileAtURL sourceLocation: URL,
+                   forDownloadTask downloadtask: URLSessionDownloadTask) -> Bool
+}
+
+@objc
+class AxisService: NSObject {
+  private weak var delegate: NYPLBookDownloadBroadcasting?
+  
+  @objc
+  init(withDelegate delegate: NYPLBookDownloadBroadcasting) {
+    self.delegate = delegate
+  }
+  
+  /// Fulfill AxisNow license
+  /// - Parameters:
+  ///   - fileURL: Downloaded LCP license URL
+  ///   - book: `NYPLBook` Book
+  ///   - downloadTask: downloadTask download task
+  @objc func fulfillAxisLicense(withFileURL fileURL: URL,
+                                forBook book: NYPLBook,
+                                downloadTask: URLSessionDownloadTask) {
+    
+  }
+  
+}
+
+#endif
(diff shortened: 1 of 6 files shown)


Simplified/AxisService/AxisService.swift
```diff
@@ -0,0 +1,39 @@
...
+//  Simplified
+//
+//  Created by Raman Singh on 2021-04-01.
+//  Copyright © 2021 NYPL Labs. All rights reserved.
+//
+
+import Foundation
+
+@objc
+protocol BookDownloadBroadcasting {
```
ettore, line 12: all our protocols and classes should have the NYPL prefix, despite the swift convention)

Simplified/AxisService/AxisService.swift
```diff
@@ -0,0 +1,39 @@
...
+  func failDownloadWithAlert(forBook book: NYPLBook)
```
singhraman4282, line 13: These methods are already present in NYPLMYBooksDownloadCenter so no extra changes are required to be added.

Simplified/AxisService/AxisService.swift
```diff
@@ -0,0 +1,39 @@
...
+  func replaceBook(_ book: NYPLBook,
+                   withFileAtURL sourceLocation: URL,
+                   forDownloadTask downloadtask: URLSessionDownloadTask) -> Bool
+}
```
singhraman4282, line 17: These methods are already present in NYPLMyBooksDownloadCenter so new changes need to be introduced.

Simplified/AxisService/AxisService.swift
```diff
@@ -0,0 +1,39 @@
...
+
+@objc
+class AxisService: NSObject {
```
ettore, line 20: do we have a requirement to be able to build the app without Axis DRM? I know we have a target with no DRMs at all, but wondering if we need to be able to turn on/off each one. Currently that's what we're doing for Adobe and LCP, so probably might make sense to do the same here.

ettore (commented):
nice small PR, I like it 👍 

Question: is this a breaking change if we merge to `develop`?  if it is breaking, you can start a feature branch and merge there. The only requirement is that it builds. Our convention for feature branch names is `feature/FEATURE_NAME`. This ensures the CI delivers a build to `iOS-binaries` repo.

Simplified/AxisService/AxisService.swift
singhraman4282: Fixed.

Simplified/AxisService/AxisService.swift
singhraman4282: Great catch. Fixed.

Simplified/AxisService/AxisService.swift
```diff
@@ -0,0 +1,43 @@
...
+//  AxisService.swift
+//  Simplified
+//
+//  Created by Raman Singh on 2021-04-01.
+//  Copyright © 2021 NYPL Labs. All rights reserved.
+//
+
+import Foundation
+
+#if FEATURE_DRM_CONNECTOR && OPENEBOOKS
```
ettore, line 11: the `FEATURE_DRM_CONNECTOR` macro is used for Adobe, so it would be better to use a different name.

PR #1350, opened 2021-04-06
Raman/feature/simply 3598 wip

singhraman4282:
Please take a look at the following files
AxisService.swift
AxisContentDownloader.swift
AxisAssetWriter.swift
AxisLicenseURLGenerator.swift
AxisXML.swift


Simplified/AxisService/AxisService.swift
```diff
@@ -20,22 +20,227 @@ protocol NYPLBookDownloadBroadcasting {
...
+          print(error)
+          group.leave()
+          self.delegate?.failDownloadWithAlert(forBook: book)
+        }
+      case .failure(let error):
+        print(error)
+        group.leave()
+        self.delegate?.failDownloadWithAlert(forBook: book)
+      }
+    }
```
singhraman4282, line 138: This code is used multiple times in the same file. Definitely needs fixing 
`AxisContentDownloader().downloadContent(from: encryptionURL) { (result) in
      switch result {
      case .success(let data):
        do {
          try AxisAssetWriter().writeAsset(data, atURL: writeURL)
          group.leave()
        } catch {
          print(error)
          group.leave()
          self.delegate?.failDownloadWithAlert(forBook: book)
        }
      case .failure(let error):
        print(error)
        group.leave()
        self.delegate?.failDownloadWithAlert(forBook: book)
      }
    }`

ettore (changes requested):
I really like the DispatchGroup solution, it's clean and makes sense!

there are a few issues mainly in relation to our merge process, which I did not explain well. In short:

- this PR merges to a branch of another PR that is still open. So it's a little confusing because that other PR is still in-flux. Usually we merge to a feature branch that starts off of `develop` so that we have a common baseline, per `git-flow`. 
- if the other PR is foundational to this one, it's best to finish that one off first, merge it to a feature branch, and then work on this one. This makes sure we review all the code once and we build incrementally maintaining the same level of code quality, even if we the feature doesn't fully work yet.
- it's fine to merge temporary hacks, especially for big features, as long as we mark them with a `TODO: SIMPLY-1234` label. Again, this is to allow for incremental development while at the same time avoiding to ship workarounds/hacks/tmp code.
- we have a convention for all class and struct names to be prefixed with `NYPL`.
- unit tests: can you add a few? is it too early? Adding some unit tests is great also for exposing modularity issues. Modularity is something we want to take more into consideration moving on. And especially for Axis DRM this is something very desirable.

PR #1354, opened 2021-04-07
Basic setup to allow Axis DRM content

singhraman4282:
**What's this do?**
Basic Setup to allow Axis DRM content

Axis DRM

Can not be tested as of yet.

N/A

N/A

N/A

Raman Singh verified it.

### DIFF
--- a/Simplified/MyBooks/NYPLMyBooksDownloadCenter.m
+++ b/Simplified/MyBooks/NYPLMyBooksDownloadCenter.m
@@ -26,6 +26,11 @@
 #import <ReadiumLCP/ReadiumLCP-Swift.h>
 #endif
 
+#if defined(AXIS)
+@interface NYPLMyBooksDownloadCenter () <NYPLBookDownloadBroadcasting>
+@end
+#endif
+
 @interface NYPLMyBooksDownloadCenter ()
   <NSURLSessionDownloadDelegate, NSURLSessionTaskDelegate>
 
@@ -132,6 +137,12 @@ totalBytesExpectedToWrite:(int64_t const)totalBytesExpectedToWrite
       self.bookIdentifierToDownloadInfo[book.identifier] =
       [[self downloadInfoForBookIdentifier:book.identifier]
        withRightsManagement:NYPLMyBooksDownloadRightsManagementAdobe];
+#if defined(AXIS)
+    } else if ([downloadTask.response.MIMEType isEqualToString:ContentTypeAxis360]) {
+      self.bookIdentifierToDownloadInfo[book.identifier] =
+      [[self downloadInfoForBookIdentifier:book.identifier]
+       withRightsManagement:NYPLMyBooksDownloadRightsManagementAxis];
+#endif
     } else if([downloadTask.response.MIMEType isEqualToString:ContentTypeReadiumLCP]) {
         self.bookIdentifierToDownloadInfo[book.identifier] =
         [[self downloadInfoForBookIdentifier:book.identifier]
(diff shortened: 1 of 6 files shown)


Simplified/AxisService/AxisService.swift
```diff
@@ -0,0 +1,101 @@
...
+
+private struct AxisHelper {
+  static let isbnKey = "isbn"
+  static let bookVaultKey = "book_vault_uuid"
+  /// This is the url for downloading content for a given book with axis drm.
+  static let baseURL = URL(string: "https://node.axisnow.com/content/stream/")!
+}
+
+@objc
+class AxisService: NSObject {
```
ettore, line 38: can you rename this to `NYPLAxisService`? feel free to do in later PR if easier

Simplified/AxisService/AxisService.swift
```diff
@@ -0,0 +1,101 @@
...
+  private let bookVaultId: String
+  private let dedicatedWriteURL: URL
+  private let fileURL: URL
+  private let baseURL: URL
+  private let book: NYPLBook
+  
+  /// Failable initializer that extracts `isbn` and `book_vault_id` from the downloaded file. Returns
+  /// nil if keys are not present and notifies delegate.
+  /// - Parameters:
+  ///   - delegate: An object confirming to NYPLBookDownloadBroadcasting protocol.
```
ettore, line 50: minor: docs for the `delegate` parameter are not really adding anything not already known. If possible I'd add what do we need this delegate for, or when it comes into play.

Simplified/AxisService/AxisService.swift
```diff
@@ -0,0 +1,101 @@
...
+  ///   - fileURL: Local url of the downloaded file.
+  ///   - book: NYPLBook object
+  @objc init?(withDelegate delegate: NYPLBookDownloadBroadcasting,
```
ettore, line 53: minor: the swift convention for initializer is to not use labels, so just `init(delegate: ...etc`)

Simplified/AxisService/AxisService.swift
```diff
@@ -0,0 +1,101 @@
...
+              fileURL: URL,
+              forBook book: NYPLBook) {
+    do {
+      let data = try Data(contentsOf: fileURL)
+      let jsonObject = try JSONSerialization
```
ettore, line 58: if this is the only thing that can `throw`, perhaps it's simpler / easier on the eye to just do a `try?` here and remove the `do-catch`, since you're validating `jsonObject` anyway in the `guard` that follows.

Simplified/MyBooks/NYPLMyBooksDownloadCenter.m
```diff
@@ -335,6 +346,22 @@ - (void)URLSession:(__attribute__((unused)) NSURLSession *)session
...
+      case NYPLMyBooksDownloadRightsManagementAxis: {
+#if defined(AXIS)
+        AxisService *axis = [[AxisService alloc]
+                             initWithDelegate:self
+                             fileURL:tmpSavedFileURL
+                             forBook:book];
+        
+        // No need to log anything in case axis initialization fails since
+        // AxisService initializer takes care of that.
+        if (axis) {
```
ettore, line 358: this `if` is unneeded in objc

ettore (approved):
mostly minor things to address but looks good!

ErnestFan (approved):

Simplified/AxisService/AxisService.swift
singhraman4282: Done

Simplified/AxisService/AxisService.swift
singhraman4282: Done

Simplified/AxisService/AxisService.swift
singhraman4282: Done

Simplified/MyBooks/NYPLMyBooksDownloadCenter.m
singhraman4282: Great catch. Fixed

Simplified/AxisService/AxisService.swift
singhraman4282: Done

PR #1358, opened 2021-04-08
SIMPLY-3671: Fixed user unable to sign in using beta libraries

singhraman4282:
**What's this do?**
Fixes user unable to sign in with beta libraries. Requires https://github.com/NYPL-Simplified/Simplified-iOS/pull/1360 to be merged in order to work.

https://jira.nypl.org/browse/SIMPLY-3671

Follow the steps to reproduce in the ticket

This will only work after https://github.com/NYPL-Simplified/Simplified-iOS/pull/1360 has been merged.

N/A

N/A
Raman Singh verified it

### DIFF
--- a/Simplified/AppInfrastructure/NYPLConfiguration+OE.swift
+++ b/Simplified/AppInfrastructure/NYPLConfiguration+OE.swift
@@ -29,10 +29,10 @@ extension NYPLConfiguration {
   private static let betaFeedFileUrl = URL(fileURLWithPath:
     Bundle.main.path(forResource: "OpenEBooks_OPDS2_Catalog_Feed-QA",
                      ofType: "json")!)
-  private static let betaFeedFileUrlHash = feedFileUrl.absoluteString.md5().base64EncodedStringUrlSafe().trimmingCharacters(in: ["="])
+  private static let betaFeedFileUrlHash = betaFeedFileUrl.absoluteString.md5().base64EncodedStringUrlSafe().trimmingCharacters(in: ["="])
 
   static var betaUrl = betaFeedFileUrl
-  static var betaUrlHash = feedFileUrlHash
+  static var betaUrlHash = betaFeedFileUrlHash
 
   // MARK:-
 


Simplified/SignInLogic/NYPLSignInBusinessLogic.swift
```diff
@@ -287,7 +287,11 @@ class NYPLSignInBusinessLogic: NSObject, NYPLSignedInStateProvider, NYPLCurrentL
       switch result {
       case .success(let responseData, _):
         #if FEATURE_DRM_CONNECTOR
-        self.drmAuthorizeUserData(responseData, loggingContext: loggingContext)
+        if NYPLSettings.shared.useBetaLibraries {
```
ettore, line 290: here we are bypassing the extraction of DRM authorization info for beta libraries, which doesn't seem quite right -- AFAIK beta libraries are supposed to work the same way as prod ones (kinda like a staging system). It has also been working fine for SimplyE.  Is this maybe meant to be just for Open eBooks? Or is it a temporary workaround?

Simplified/WelcomeScreen/OETutorialChoiceViewController.swift
```diff
@@ -112,6 +112,17 @@ class OETutorialChoiceViewController : UIViewController {
     let libAccount = AccountsManager.shared.currentAccount
     let userAccount = NYPLUserAccount.sharedAccount()
     if libAccount?.details == nil {
+      
+      if (libAccount == nil) {
```
ettore, line 116: should this `if` be moved outside of the `if libAccount?.detail`? Seems more like a `guard` to be placed at the top?

ErnestFan: It looks to me it's trying to load the catalog directly, maybe because we are not retaining/getting the library account info?
If this is temporary workaround, can @singhraman4282 put a note on it?
Also, we should find out what's the underlying issue here.

Simplified/SignInLogic/NYPLSignInBusinessLogic.swift
singhraman4282: It's a temporary workaround until the PR for https://jira.nypl.org/browse/SIMPLY-3607 gets merged in

Simplified/WelcomeScreen/OETutorialChoiceViewController.swift
singhraman4282: All of a sudden the issue stopped happening. I was unable to reproduce this upon multiple attempts.

Simplified/SignInLogic/NYPLSignInBusinessLogic.swift
singhraman4282: Removed.

ettore (approved):

ErnestFan (approved):

PR #1360, opened 2021-04-11
SIMPLY-3607: Removes Adept and LCP from OpenEbooks

singhraman4282:
**What's this do?**
Removes Adept and LCP from OpenEbooks and Makes Axis360 preferred acquisition for downloading books.
With this, we no longer need any hacks to present a custom feed to the user for testing Axis content download.

https://jira.nypl.org/browse/SIMPLY-3607

Can not be QA tested by itself. For devs, when downloading a book on OpenEbooks, a link with Axis DRM will be preferred over any other link.

N/A

N/A

N/A

Raman Singh verified it.

### DIFF
--- a/Simplified/AxisService/NYPLAxisDRMAuthorizer.swift
+++ b/Simplified/AxisService/NYPLAxisDRMAuthorizer.swift
@@ -1,0 +1,61 @@
+//
+//  NYPLAxisDRMAuthorizer.swift
+//  Simplified
+//
+//  Created by Raman Singh on 2021-04-13.
+//  Copyright © 2021 NYPL Labs. All rights reserved.
+//
+
+import Foundation
+
+#if AXIS
+@objcMembers
+class NYPLAxisDRMAuthorizer: NSObject, NYPLDRMAuthorizing {
+  
+  static let deviceIDKey = "NYPLAxisDRMAuthorizerKey"
+  
+  static let sharedInstance = NYPLAxisDRMAuthorizer()
+  
+  static var deviceID: String {
+    if
+      let id = UserDefaults.standard.string(forKey: NYPLAxisDRMAuthorizer.deviceIDKey) {
+      return id
+    }
+    
+    let id = UUID().uuidString
+    UserDefaults.standard.set(id, forKey: NYPLAxisDRMAuthorizer.deviceIDKey)
+    return id
+  }
+  
+  /// Not applicable for Axis
+  var workflowsInProgress: Bool {
+    return false
+  }
+  
+  /// We're always returning true since Axis does not deauthorize a user.
+  func isUserAuthorized(_ userID: String!,
+                        withDevice device: String!) -> Bool {
+   return true
+  }
+  
+  /// As of now, we're creating a deviceID which stays saved until the user uninstalls the app.
+  func authorize(withVendorID vendorID: String!,
+                 username: String!,
+                 password: String!,
+                 completion: ((Bool, Error?, String?, String?) -> Void)!) {
+    
+    completion(true, nil, NYPLAxisDRMAuthorizer.deviceID, username)
+  }
+  
+  /// There is no mechanism for deauthorizing the user in Axis. Returning succeeding completion.
+  func deauthorize(withUsername username: String!,
+                   password: String!,
+                   userID: String!,
+                   deviceID: String!,
+                   completion: ((Bool, Error?) -> Void)!) {
+    
+    completion(true, nil)
+  }
+
+}
+#endif
--- a/Simplified/Book/Models/NYPLBook.m
+++ b/Simplified/Book/Models/NYPLBook.m
@@ -534,19 +534,33 @@ static NSString *const UpdatedKey = @"updated";
     NYPLLOG(@"ERROR: No acquisitions found when computing a default. This is an OPDS violation.");
     return nil;
   }
-
+  
+  NYPLOPDSAcquisition *fallbackAcquisition;
+  
   for (NYPLOPDSAcquisition *const acquisition in self.acquisitions) {
     NSArray *const paths = [NYPLOPDSAcquisitionPath
                             supportedAcquisitionPathsForAllowedTypes:[NYPLOPDSAcquisitionPath supportedTypes]
                             allowedRelations:NYPLOPDSAcquisitionRelationSetAll
                             acquisitions:@[acquisition]];
-
+    
     if (paths.count >= 1) {
+#if defined(AXIS)
+      for (NYPLOPDSAcquisitionPath *path in paths) {
+        if ([path.types containsObject:ContentTypeAxis360]) {
+          return acquisition;
+        }
+      }
+      
+      if (!fallbackAcquisition) {
+        fallbackAcquisition = acquisition;
+      }
+#else
       return acquisition;
+#endif
     }
   }
-
-  return nil;
+  
+  return fallbackAcquisition;
 }
 
 - (NYPLOPDSAcquisition *)defaultAcquisitionIfBorrow
--- a/Simplified/SignInLogic/NYPLSignInBusinessLogic+Adept.swift
+++ b/Simplified/SignInLogic/NYPLSignInBusinessLogic+Adept.swift
@@ -1,0 +1,125 @@
+//
+//  NYPLSignInBusinessLogic+Adept.swift
+//  Simplified
+//
+//  Created by Raman Singh on 2021-04-14.
+//  Copyright © 2021 NYPL Labs. All rights reserved.
+//
+
+import Foundation
+
+#if FEATURE_DRM_CONNECTOR
+
+extension NYPLSignInBusinessLogic {
+  
+  /// Extracts username and password from user profile's drm's client token and authorizes with Adept DRM
+  /// - Parameters:
+  ///   - profileDoc: UserProfile document from which client token is extracted
+  ///   - loggingContext: Information to report when logging errors.
+  func authorizeWithAdobe(userProfile profileDoc: UserProfileDocument,
+                          loggingContext: [String: Any]) {
+    
+    guard
+      let drm = profileDoc.drm?.first,
+      drm.vendor != nil,
+      let clientToken = drm.clientToken else {
+
+        let drm = profileDoc.drm?.first
+        Log.info(#file, "\nLicensor: \(drm?.licensor ?? ["N/A": "N/A"])")
+        
+        NYPLErrorLogger.logError(
+          withCode: .noLicensorToken,
+          summary: "SignIn: no licensor token in user profile doc",
+          metadata: loggingContext)
+        
+        finalizeSignIn(
+          forDRMAuthorization: false,
+          errorMessage: "No credentials were received to authorize access to books with DRM.")
+        return
+    }
+    
+    Log.info(#file, "\nLicensor: \(drm.licensor)")
+    userAccount.setLicensor(drm.licensor)
+    
+    var licensorItems = clientToken
+      .replacingOccurrences(of: "\n", with: "")
+      .components(separatedBy: "|")
+    
+    let tokenPassword = licensorItems.last
+    licensorItems.removeLast()
+    let tokenUsername = (licensorItems as NSArray).componentsJoined(by: "|")
+    
+    drmAuthorize(username: tokenUsername,
+                 password: tokenPassword,
+                 loggingContext: loggingContext)
+    
+  }
+  
+  /// Perform the DRM authorization request with the given credentials
+  ///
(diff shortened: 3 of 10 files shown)


Simplified/Book/Models/NYPLBook.m
```diff
@@ -534,19 +534,33 @@ - (NYPLOPDSAcquisition *)defaultAcquisition
...
+#else
       return acquisition;
+#endif
     }
   }
-
-  return nil;
+  
+  return fallbackAcquisition;
 }
```
singhraman4282, line 565: Here's another alternative which specifically filters out Adept and LCP

```
- (NYPLOPDSAcquisition *)defaultAcquisition
{
  if (self.acquisitions.count == 0) {
    NYPLLOG(@"ERROR: No acquisitions found when computing a default. This is an OPDS violation.");
    return nil;
  }
  
  for (NYPLOPDSAcquisition *const acquisition in self.acquisitions) {
    NSArray *const paths = [NYPLOPDSAcquisitionPath
                            supportedAcquisitionPathsForAllowedTypes:[NYPLOPDSAcquisitionPath supportedTypes]
                            allowedRelations:NYPLOPDSAcquisitionRelationSetAll
                            acquisitions:@[acquisition]];
    
    if (paths.count >= 1) {
#if defined(AXIS)
      for (NYPLOPDSAcquisitionPath *path in paths) {
        if ([path.types containsObject:ContentTypeReadiumLCP]
            || [path.types containsObject:ContentTypeAdobeAdept]) {
          continue;
        }
      }
#endif
      return acquisition;
    }
  }
  
  return nil;
}
```

singhraman4282:
Regarding the failing test NYPLSingInBusinessLogicTests::testLoginFlow, it's failing because user.deviceID and user.userID are nil. 

Originally we were setting these values upon calling ```NYPLSingInBusinessLogic+DRM::drmAuthorizeData``` method or from ```NYPLMigrationManager::migrate_1_8_1``` method but since these methods are no longer being called, we're no longer setting these values. However, we will need deviceID for Axis DRM.

We can use the current device ID with ```UIDevice.current.identifierForVendor?.uuidString``` but this seems wrong for some reason. Any suggestions @ErnestFan @ettore ?

killerpuppytails:
> We can use the current device ID with `UIDevice.current.identifierForVendor?.uuidString` but this seems wrong for some reason. Any suggestions @ErnestFan @ettore ?

How did @qnga solve this for Android?

ettore:
> Regarding the failing test `NYPLSingInBusinessLogicTests::testLoginFlow`, it's failing because `user.deviceID` and `user.userID` are nil.
> 
> Originally we were setting these values upon calling `NYPLSingInBusinessLogic+DRM::drmAuthorizeData` method or from `NYPLMigrationManager::migrate_1_8_1` method but since these methods are no longer being called, we're no longer setting these values. However, we will need deviceID for Axis DRM.

I think `drmAuthorizeUserData` should still be called, but it will need to be refactored to be agnostic from a particular DRM system. Currently it set up to only work for Adobe, but there are things in there that are more generic, for instance parsing the UserProfileDocument and extracting the `authorizationIdentifier`, which we will still need.

> 
> We can use the current device ID with `UIDevice.current.identifierForVendor?.uuidString` but this seems wrong for some reason. Any suggestions @ErnestFan @ettore ?

That's a different UUID and I don't know if it can be used for DRM purposes. I suppose it could depending if Axis cares or not on the format of the ID? Maybe you can touch base with the android guys (Mark and Quentin) to see what they use?

qnga:
I actually did nothing about that! Please check with @io7m.

singhraman4282:
There's a difference between the response data we get when signing in with beta libraries vs when signing in using production library

Response data form default libraries

`{\"simplified:authorization_identifier\": \"MX8GNTIL6I\", \"drm\": [{\"drm:vendor\": \"NYPL\", \"drm:scheme\": \"http://librarysimplified.org/terms/drm/scheme/ACS\", \"drm:clientToken\": \"USOEI|1618269934|3c2ecd48-d2c4-11e6-a6bd-0e5401542847|TVoCnc3FH9Qh5fmmmVYU5F2s2Qjy3gcYYnuMi0OJncQ@\"}], \"links\": [{\"href\": \"https://circulation.openebooks.us/USOEI/AdobeAuth/devices\", \"rel\": \"http://librarysimplified.org/terms/drm/rel/devices\"}, {\"href\": \"https://circulation.openebooks.us/USOEI/annotations/\", \"type\": \"application/ld+json; profile=\\\"http://www.w3.org/ns/anno.jsonld\\\"\", \"rel\": \"http://www.w3.org/ns/oa#annotationService\"}], \"settings\": {\"simplified:synchronize_annotations\": true}}`

Response data form beta libraries

`{\"simplified:authorization_identifier\": \"MX8GNTIL6I\", \"settings\": {\"simplified:synchronize_annotations\": null}}`

I'm guessing someone on the backend side needs to change this. What do you think @ettore ?

ettore:
> There's a difference between the response data we get when signing in with beta libraries vs when signing in using production library

yeah, probably.... I mean some things there seem to indicate some client side settings were different (for example the sync annotations part), but you're right, all things being equal on the client I'm not sure about the other discrepancies here.

Simplified/AxisService/NYPLAxisDRMAuthorizer.swift
```diff
@@ -0,0 +1,81 @@
...
+    return false
+  }
+  
+  /// We're always returning true since Axis does not deauthorize a user.
+  func isUserAuthorized(_ userID: String!,
+                        withDevice device: String!) -> Bool {
+   return true
+  }
+  
+  /// TODO: SIMPLY-3692 will have code that fixes the issue.
```
ettore, line 41: sorry, which issue are you referring to? 

regarding possibly clearing the deviceID from UserDefaults, I sort of think we don't have to worry about that. Even if the user logs out, I don't see any issues with just leaving it there. It might actually be better to leave it, so it's consistently unique per device (unless user uninstalls the app, but that another story, and it's unavoidable)

Simplified/AxisService/NYPLAxisDRMAuthorizer.swift
```diff
@@ -0,0 +1,81 @@
...
+  /// As of now, we're creating a deviceID which stays saved until the user uninstalls the app.
+  func authorize(withVendorID vendorID: String!,
+                 username: String!,
+                 password: String!,
+                 completion: ((Bool, Error?, String?, String?) -> Void)!) {
+    
+    /*
+     In NYPLSignInBusinessLogic+OE::drmAuthorize:username:password:loggingContext,
+     we expect this method to take some time. After calling this method, we
```
ettore, line 50: since this implementation of `authorize(withVendorID:username:password:completion:)` does not need any delay and its execution is synchronous (we just need to call `completion(true, nil, NYPLAxisDRMAuthorizer.deviceID, username)`, correct?) I think we should just avoid the horrible hack we have in `NYPLSignInBusinessLogic+DRM::drmAuthorize(username:password:loggingContext:)` entirely. Can we refactor that code so that for OE we don't need to run any of that code? There's no need to comply to that craziness now that we don't need Adobe anymore. 

One quick way could be to move that `drmAuthorize(username:password:loggingContext:)` function out into an extension, and compile that only for SimplyE, and then write another extension for OE. That's just the first I could think of though.

singhraman4282: Agreed. That makes a lot more sense. However, the only thing we're doing different from Adobe is calling the selector after delay and cancelling the selector in completion handler. I was of the opinion that the following code should do the job.

```
private func drmAuthorize(username: String,
                            password: String?,
                            loggingContext: [String: Any]) {

    let vendor = userAccount.licensor?["vendor"] as? String

    Log.info(#file, """
      ***DRM Auth/Activation Attempt***
      Token username: \(username)
      Token password: \(password ?? "N/A")
      VendorID: \(vendor ?? "N/A")
      """)

    drmAuthorizer?
      .authorize(withVendorID: vendor,
                 username: username,
                 password: password) { success, error, deviceID, userID in
                  #if FEATURE_DRM_CONNECTOR
                  // make sure to cancel the previously scheduled selector
                  // from the same thread it was scheduled on
                  NYPLMainThreadRun.asyncIfNeeded { [weak self] in
                    if let self = self {
                      NSObject.cancelPreviousPerformRequests(withTarget: self)
                    }
                  }
                  #endif

                  Log.info(#file, """
                    Activation success: \(success)
                    Error: \(error?.localizedDescription ?? "N/A")
                    DeviceID: \(deviceID ?? "N/A")
                    UserID: \(userID ?? "N/A")
                    ***DRM Auth/Activation completion***
                    """)

                  var success = success

                  if success, let userID = userID, let deviceID = deviceID {
                    NYPLMainThreadRun.asyncIfNeeded {
                      self.userAccount.setUserID(userID)
                      self.userAccount.setDeviceID(deviceID)
                    }
                  } else {
                    success = false
                    NYPLErrorLogger.logLocalAuthFailed(error: error as NSError?,
                                                       library: self.libraryAccount,
                                                       metadata: loggingContext)
                  }

                  self.finalizeSignIn(forDRMAuthorization: success,
                                      error: error as NSError?)
    }

    #if FEATURE_DRM_CONNECTOR
    NYPLMainThreadRun.asyncIfNeeded { [weak self] in
      self?.perform(#selector(self?.dismissAfterUnexpectedDRMDelay), with: self, afterDelay: 25)
    }
    #endif
  }
```

singhraman4282: However this does make the method a bit harder to understand and modify. Your approach is better and I think we should go with that.

ettore: yeah, if I understand correctly it doesn't seem like we need that complex implementation of `drmAuthorize(username:password:loggingContext:)` on Open eBooks.

Simplified/AxisService/NYPLAxisDRMAuthorizer.swift
singhraman4282: Great catch. This is leftover comment from when we were not sure what to do with deviceID.  Removed.

Simplified/AxisService/NYPLAxisDRMAuthorizer.swift
```diff
@@ -0,0 +1,61 @@
...
+   return true
+  }
+  
+  /// As of now, we're creating a deviceID which stays saved until the user uninstalls the app.
+  func authorize(withVendorID vendorID: String!,
+                 username: String!,
+                 password: String!,
+                 completion: ((Bool, Error?, String?, String?) -> Void)!) {
+    
+    completion(true, nil, NYPLAxisDRMAuthorizer.deviceID, username)
```
ettore, line 47: I love it! #simplicity #byeadept

Simplified/SignInLogic/NYPLSignInBusinessLogic+Adept.swift
```diff
@@ -0,0 +1,125 @@
...
+#if FEATURE_DRM_CONNECTOR
+
+extension NYPLSignInBusinessLogic {
+  
+  /// Exttracts username and password from user profile's drm's client token and authorizes with Adept DRM
+  /// - Parameters:
+  ///   - profileDoc: UserProfile document from which client token is extracted
+  ///   - loggingContext: Information to report when logging errors.
+  func authorizeWithAdobe(userProfile profileDoc: UserProfileDocument,
+                                  loggingContext: [String: Any]) {
```
ettore, line 20: minor: indentation and typo in docs

ettore (approved):
Looks good to me!

Simplified/AxisService/NYPLAxisDRMAuthorizer.swift
singhraman4282, line 47: Since the NYPLDRMAuthorizing object does not do anything here, do you think we're violating( or not properly following) the Liskov Substitution Principal here @ettore @ErnestFan?

ettore: ok I confess I had to look that up 😅 but I don't think so we're violating it... considering NYPLAxisDRMAuthorizer is a subtype of NYPLDRMAuthorizing, and the latter is a protocol with no implementation, all we can ask is, can we use the concrete class anywhere we need the protocol? The answer is definitely yes.

Also, it's not entirely true this authorizer does nothing. It creates a device ID, persists it, and returns it consistently. The fact that there's no authorization process seems ok to me. A detail if you will.... basically Axis is saying, if you can use the DRM stuff on device X, you can also use it on device Y, whatever that is.

ErnestFan (approved):

PR #1361, opened 2021-04-14
SIMPLY-3670 Reconciliate old bookmark format with new format

ettore:
**What's this do?**
Enforces creation of new bookmarks following the spec. This now includes:
- when creating a bookmark in a ereader context, we use the R2 Publication object to derive the `href` upfront and store it in the bookmark, instead of doing it later. 
- when creating a bookmark from the dictionary representation, the `href` may be nil if the bookmark was created in R1. In this case the `href` can be derived only when we open the book. However in this case the `idref` MUST be not nil. This is now expressed at compile time via the `ChapterID` enum.
- when creating a dictionary representation, always save all fields required by the spec, and not old ones
- mark the total progression as non-optional, per bookmark spec.

https://jira.nypl.org/browse/SIMPLY-3670

too early to test, this is just a subtask. 

merging to feature branch.

not yet

yes, expanded existing docs.

@ettore

### DIFF
--- a/Simplified/Reader2/Bookmarks/NYPLReadiumBookmark+R2.swift
+++ b/Simplified/Reader2/Bookmarks/NYPLReadiumBookmark+R2.swift
@@ -17,8 +17,8 @@ extension NYPLReadiumBookmark {
   /// for this conversion: only what's strictly necessary to be able to point
   /// at the same location inside the `Publication`.
   ///
-  /// - Complexity: O(*n*) where *n* is the length of internal `Publication`
-  /// data structures, such as list of chapters, resources, links.
+  /// - Complexity: O(*n*) where *n* is the length of internal
+  /// `Publication.readingOrder` data structure.
   ///
   /// - Parameter publication: The R2 publication object where the bookmark is
   /// located.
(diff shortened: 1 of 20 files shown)


Simplified/Reader2/Bookmarks/NYPLReadiumBookmark+R2.swift
```diff
@@ -17,28 +17,34 @@ extension NYPLReadiumBookmark {
...
+  /// `Publication.readingOrder` data structure.
   ///
   /// - Parameter publication: The R2 publication object where the bookmark is
   /// located.
 
   /// - Returns: An object with R2 location information pointing at the same
   /// position the bookmark model is pointing to.
   func convertToR2(from publication: Publication) -> NYPLBookmarkR2Location? {
-    guard let link = publication.link(withIDref: self.idref) else {
+    let href: String!
```
singhraman4282, line 29: I think we can avoid implicit unwrapping here

ettore: this implicit unwrapping is 100% safe in the code below, but I couldn't find a way that is more readable. I'm open to ideas though, do you have something specific in mind?

singhraman4282 (approved):

ErnestFan (approved):

Simplified/Reader2/Bookmarks/NYPLReadiumBookmark+R2.swift
ettore: this implicit unwrapping is 100% safe in the code below, but I couldn't find a way that is more readable. I'm open to ideas though, do you have something specific in mind?

ettore: fixed and rebased

PR #1362, opened 2021-04-15
SIMPLY-3668 Refactor GET annotations api calls

ettore:
**What's this do?**
The `syncReadingPosition` and `getServerBookmarks` functions call the same api, so the parsing of the api response is identical. This PR moves the parsing code into its own function, so it can be reused by the two api call wrappers. Also adds related unit tests.

Minor: in NYPLAnnotations, `postAnnotation` and `annotationID(fromNetworkData:)` have been moved as-is.

https://jira.nypl.org/browse/SIMPLY-3668

will be able to test once all story subtasks have been completed. 

merging to feature branch

not needed

yes

@ettore

### DIFF
--- a/Simplified/Reader2/Bookmarks/NYPLAnnotations.swift
+++ b/Simplified/Reader2/Bookmarks/NYPLAnnotations.swift
@@ -228,7 +228,6 @@ import R2Shared
 
   /// Reads the current reading position from the server, parses the response
   /// and returns the result to the `completionHandler`.
-  // TODO: SIMPLY-3668 Refactor with `syncReadingPosition(...)`
   class func syncReadingPosition(ofBook bookID: String?,
                                  publication: Publication?,
                                  toURL url:URL?,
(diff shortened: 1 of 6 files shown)


ErnestFan (approved):

Simplified/Reader2/Bookmarks/NYPLAnnotations.swift
```diff
@@ -590,7 +465,116 @@ import R2Shared
...
+        completionHandler(false, nil)
+        return
+      }
+      guard let statusCode = (response as? HTTPURLResponse)?.statusCode else {
+        Log.error(#file, "Annotation POST error: No response received from server")
+        completionHandler(false, nil)
+        return
+      }
+
+      if statusCode == 200 {
```
singhraman4282, line 550: Is there a chance that the server might return 201 upon success at some point? I think we should consider any status code within the range 200...299 as success.

singhraman4282 (approved):

Simplified/Reader2/Bookmarks/NYPLAnnotations.swift
ettore: I'm not sure -- this is just code I moved from one place to another, and it's out of scope of this ticket, so I was just going to leave it as is. Really what should be done here is a bigger refactor for the whole request execution, i.e. going through out NYPLNetworkExecutor layer. That way all of this error handling will be done for free (including proper status code detection)

PR #1363, opened 2021-04-15
SIMPLY-3672: Get DeviceID and Client IP for AXIS license URL generation 

singhraman4282:
**What's this do?**
Gets DeviceID and Client IP for AXIS license URL generation 

We need deviceID and clientIP in order to generate a url for downloading the license file for AXIS books.
https://jira.nypl.org/browse/SIMPLY-3672

Too early to test

N/A

N/A

N/A

Raman Singh verified it

### DIFF
--- a/Simplified/AxisService/NYPLAxisDRMAuthorizer.swift
+++ b/Simplified/AxisService/NYPLAxisDRMAuthorizer.swift
@@ -9,14 +9,27 @@
 import Foundation
 
 #if AXIS
+
+@objc protocol NYPLDeviceInfoProviding {
+  var deviceID: String { get }
+  var clientIP: String { get }
+}
+
 @objcMembers
-class NYPLAxisDRMAuthorizer: NSObject, NYPLDRMAuthorizing {
+class NYPLAxisDRMAuthorizer: NSObject, NYPLDRMAuthorizing, NYPLDeviceInfoProviding {
   
-  static let deviceIDKey = "NYPLAxisDRMAuthorizerKey"
+  static private let deviceIDKey = "NYPLAxisDRMAuthorizerKey"
   
   static let sharedInstance = NYPLAxisDRMAuthorizer()
   
-  static var deviceID: String {
+  // MARK: - NYPLDeviceInfoProviding
+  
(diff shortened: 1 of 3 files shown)


ErnestFan (approved):

Simplified/AxisService/NYPLAxisDRMAuthorizer.swift
```diff
@@ -9,14 +9,23 @@
...
+class NYPLAxisDRMAuthorizer: NSObject, NYPLDRMAuthorizing, NYPLDeviceInfoProviding {
   
-  static let deviceIDKey = "NYPLAxisDRMAuthorizerKey"
+  static private let deviceIDKey = "NYPLAxisDRMAuthorizerKey"
   
   static let sharedInstance = NYPLAxisDRMAuthorizer()
   
-  static var deviceID: String {
+  // MARK: - NYPLDeviceInfoProviding
+  let clientIP = "192.168.1.254"
```
ettore, line 26: this hardcoded IP should probably be left out ?

ettore (commented):
looks good, just unclear about the hardcoded IP address

Simplified/AxisService/NYPLAxisDRMAuthorizer.swift
singhraman4282: Device IP address is only needed when generating a url for downloading license. Android is also hardcoding the ip address. Looking at it from a Cost vs Benefit perspective, I'm under the impression that generating device IP address would be more costly and would provide a result no different than what the hard coded IP address already provides. 

However I'm open to suggestions and will be more than happy to not having to hardcode it if you think we should generate the ip address for the device.

ettore: I see.... so this IP can be any number? If so can you add some documentation explaining why we need this and that it doesn't actually have to be the real client IP
 (assuming I understood correctly)?  Otherwise developers reading this code won't know why we have this constant here.

singhraman4282: Agreed. Done.

ettore (approved):

PR #1364, opened 2021-04-15
SIMPLY-3667 Consolidate locator string creation, fix resume reading bugs

ettore:
**What's this do?**
When syncing the reading progress we were converting a R2 Locator to our NYPLBookLocation data structure and then converting it back to a Locator, which is unneeded.

Also:
- now NYPLBookLocation uses the shared factory methods to serialize a R2 Locator instead of a its own custom way.
- Add more tolerance when converting NYPLBookLocation to Locator.
- NYPLReadiumBookmark can create its own R2 Locator.
- I lowered the throttling threshold because I was incurring in missed updates.

https://jira.nypl.org/browse/SIMPLY-3667

I will add QA info in the parent ticket.

merging to feature branch

yes

yes

@ettore


Simplified.xcodeproj/project.pbxproj
```diff
@@ -6928,7 +6928,7 @@
 				CLANG_WARN_NON_LITERAL_NULL_CONVERSION = YES_ERROR;
 				CODE_SIGN_ENTITLEMENTS = Simplified/SimplyE.entitlements;
 				CODE_SIGN_IDENTITY = "iPhone Distribution";
-				CURRENT_PROJECT_VERSION = 13;
+				CURRENT_PROJECT_VERSION = 14;
```
ErnestFan, line 6931: We already have 3.7.0(14) on Firebase, should this be 15 instead?

ettore: you're right, I missed it somehow... fixed

ErnestFan (approved):

singhraman4282 (approved):

PR #1365, opened 2021-04-15
SIMPLY-3646 Fix announcement UX

ErnestFan:
**What's this do?**
Use UIAlertController instead of ViewController when presenting announcements

[SIMPLY-3646](https://jira.nypl.org/browse/SIMPLY-3646)

Login and switch to NYPL QA library
Kill SimplyE and restart
Announcements should show up while the catalog being loaded in the background
Dismiss the announcements and restart the app
There should be no more announcements to show

N/A

Hold until 3.6.4 is released

N/A

@ErnestFan


Simplified/Accounts/Library/Account.swift
```diff
@@ -425,14 +425,14 @@ class OPDS2SamlIDP: NSObject, Codable {
         do {
           self.authenticationDocument = try
             OPDS2AuthenticationDocument.fromData(serverData)
+          completion(true)
```
ErnestFan, line 428: Calling the `completion` before presenting the announcements, so the catalog can load in the background

Simplified/ErrorHandling/NYPLAlertUtils.swift
```diff
@@ -197,3 +197,42 @@ import UIKit
...
+    var currentAlert: UIAlertController? = nil
+    
+    let alerts = announcements.map {
+      UIAlertController.init(title: title, message: $0.content, preferredStyle: .alert)
+    }
+    
+    // Present another alert when the current alert is being dismiss
+    // Add the presented announcement to the presentedAnnouncement document
+    for (i, alert) in alerts.enumerated() {
+      if i > 0 {
```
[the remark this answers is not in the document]
ErnestFan: This `if` statement is only called when there is more than 1 announcement, and we present the next announcement when the current one being dismissed.
Otherwise, we just `addPresentedAnnouncement(id:)` without presenting another announcement which happens in the `if let` statement below.
I can use `currentAlert != nil` instead of `i < 0` if it makes the logic more understandable.

Simplified/ErrorHandling/NYPLAlertUtils.swift
```diff
@@ -197,3 +197,42 @@ import UIKit
...
+        let action = UIAlertAction.init(title: NSLocalizedString("OK", comment: ""),
+                                        style: .default) { _ in
+          NYPLRootTabBarController.shared()?.safelyPresentViewController(alert, animated: true, completion: nil)
+          NYPLAnnouncementBusinessLogic.shared.addPresentedAnnouncement(id: announcements[i - 1].id)
```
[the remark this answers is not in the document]
ErnestFan: Updated

Simplified/Announcements/NYPLAnnouncementBusinessLogic.swift
```diff
@@ -69,6 +74,45 @@ class NYPLAnnouncementBusinessLogic {
...
   }
+  
+  // MARK: - Helper
+  
+  /**
+  Generates an alert view that presents another alert when being dismissed
+  - Parameter announcements: an array of announcements that goes into alert message.
+  - Returns: The alert controller to be presented.
+  */
+  private func alert(announcements: [Announcement]) -> UIAlertController? {
```
ErnestFan, line 85: Moved this function as part of NYPLAnnouncementBusinessLogic as @ettore suggest, since it is only being called by the business logic and to avoid tangling.

singhraman4282 (approved):

Simplified/Announcements/NYPLAnnouncementBusinessLogic.swift
```diff
@@ -2,8 +2,12 @@ import Foundation
 
 private let announcementsFilename: String = "NYPLPresentedAnnouncementsList"
 
+protocol NYPLAnnouncementProvider: class {
```
ettore, line 5: do we need this protocol? it seems unneeded since it's only used inside NYPLAnnouncementBusinessLogic... is it maybe for future unit tests?

ettore (approved):
looks good

Simplified/Announcements/NYPLAnnouncementBusinessLogic.swift
ErnestFan: Forgot to remove that, thanks for the catch

PR #1366, opened 2021-04-19
OE-16: Move downloaded file to content directory

singhraman4282:
**What's this do?**
Moves downloaded book content from downloads to content directory.

https://jira.nypl.org/browse/OE-16

Too early to test.

N/A

N/A

N/A

Raman Singh verified it.


singhraman4282:
Found a bug

PR #1368, opened 2021-04-20
SIMPLY-3608 Remove come code duplication, minor cleanup

ettore:
**What's this do?**
- remove duplication between convertToR2(from:) and `locator(forPublication:)` functions
- made properties of NYPLReadiumBookmark object `let`s instead of `var`s
- removed unused "page" property
- rename `NYPLReadiumBookmark+R2.swift` to `NYPLReadiumBookmark+Compare.swift` and added `isEqual` and `lessThan` (no changes)

Cleanup for SIMPLY-3608

I'll  add steps in ticket.

merging to feature branch

yes

yes

@ettore


singhraman4282 (approved):

Simplified/Reader2/Bookmarks/NYPLReadiumBookmark+Compare.swift
```diff
@@ -0,0 +1,71 @@
...
+      locatorChapterProgress = Float(chapterProgress)
+    } else {
+      locatorChapterProgress = nil
+    }
+
+    guard self.href == locator.href else {
+      return false
+    }
+
+    if self.progressWithinChapter =~= locatorChapterProgress {
```
ErnestFan, line 37: We can just `return self.progressWithinChapter =~= locatorChapterProgress` here

ErnestFan (approved):

Simplified/Reader2/Bookmarks/NYPLReadiumBookmark+Compare.swift
ettore: lol good catch, yes we can and should :) will fix

PR #1370, opened 2021-04-21
IOS 41 Save chapter percentage in R1 progress update

ettore:
**What's this do?**
Reading progress updates in R1 only use the CFI, which is useless in R2.  This adds the progress as a percentage, allowing us to migrate the reading progress to the new format we use in R2 once we switch over.

https://jira.nypl.org/browse/IOS-41

will add in ticket

this is going to be released as a hot fix after 3.6.4 ships.

yes

yes

@ettore


Simplified/NYPLReaderReadiumView.m
```diff
@@ -871,8 +878,30 @@ - (void)readiumPaginationChangedWithDictionary:(NSDictionary *const)dictionary
           spineItemTitle:weakSelf.spineItemDetails[@"tocElementTitle"]];
        }];
 
-       NYPLBookLocation *const location = [[NYPLBookLocation alloc] initWithLocationString:locationJSON renderer:renderer];
-       NSString *const bookID = weakSelf.book.identifier;
+      float chapterProgress = [self progressWithinChapter];
```
ErnestFan, line 881: minor: indentation is off

ErnestFan (approved):

Simplified/NYPLReaderReadiumView.m
ettore: yeah... actually the off part is in the rest of the code, which I almost fixed but decided not to, to avoid conflicts when I merge this code with the newer R2 changes. Since this whole file will be deleted soon once R2 is done, I just left it as is.

Simplified/NYPLReaderReadiumView.m
```diff
@@ -871,8 +878,30 @@ - (void)readiumPaginationChangedWithDictionary:(NSDictionary *const)dictionary
...
+      json[@"progressWithinChapter"] = @(chapterProgress);
+      NSError *serializationError = nil;
+      NSData *jsonData = [NSJSONSerialization dataWithJSONObject:json
+                                                         options:0
+                                                           error:&serializationError];
+      NSString *updatedJSON;
+      if (serializationError != nil || jsonData == nil) {
+        [NYPLErrorLogger logError:serializationError
```
singhraman4282, line 889: minor: `if (serializationError || jsonData) `

singhraman4282 (approved):

PR #1371, opened 2021-04-26
OE-62: Created a downloader responsible for downloading Axis content

singhraman4282:
**What's this do?**
Creates a downloader responsible for downloading Axis content

https://jira.nypl.org/browse/OE-62

This cannot be thoroughly tested by front-end QA. If you're able to download an axis book, it means it's working. Also, additional Unit tests have been written.

N/A

N/A

N/A

Raman Singh verified it

### DIFF
--- a/Simplified/AxisService/NYPLAxisContentDownloader.swift
+++ b/Simplified/AxisService/NYPLAxisContentDownloader.swift
@@ -1,0 +1,165 @@
+//
+//  NYPLAxisContentDownloader.swift
+//  Simplified
+//
+//  Created by Raman Singh on 2021-04-22.
+//  Copyright © 2021 NYPL Labs. All rights reserved.
+//
+
+import Foundation
+
+protocol NYPLAxisContentDownloading {
+  func downloadItem(from url: URL,
+                    _ completion: @escaping (Result<Data, Error>) -> Void)
+}
+
+
+/// A content downloader which stops downloading content upon receiving a failure.
+class NYPLAxisContentDownloader: NYPLAxisContentDownloading {
+  
+  private let tasks: ThreadSafeDictionaryContainer<URL, URLSessionDataTask>
+  private let handlers: ThreadSafeDictionaryContainer<URL, (Result<Data, Error>) -> Void>
+  private let numberOfAttempts: ThreadSafeDictionaryContainer<URL, Int>
+  private let errored: ThreadSafeValueContainer<Bool>
+  let networkExecutor: NYPLAxisNetworkExecuting
+  
+  init(networkExecuting: NYPLAxisNetworkExecuting) {
+    let queue = DispatchQueue(label: "AxixThreadSafeContainer")
+    self.tasks = ThreadSafeDictionaryContainer(queue: queue)
+    self.handlers = ThreadSafeDictionaryContainer(queue: queue)
+    self.numberOfAttempts = ThreadSafeDictionaryContainer(queue: queue)
+    self.errored = ThreadSafeValueContainer(value: false, queue: queue)
+    self.networkExecutor = networkExecuting
+  }
+  
+  func downloadItem(from url: URL,
+                    _ completion: @escaping (Result<Data, Error>) -> Void) {
+    
+    // We dont want to download anything if it's already being downloaded or if
+    // we experienced an error downloading something before.
+    let isErrored = errored.value ?? false
+    guard !isErrored, tasks[url] == nil, handlers[url] == nil else {
+      return
+    }
+    
+    numberOfAttempts[url] = (self.numberOfAttempts[url] ?? 0) + 1
+    handlers[url] = completion
+    tasks[url] = self.getTask(for: url)
+  }
+  
+  private func getTask(for url: URL) -> URLSessionDataTask {
+    
+    let request = URLRequest(url: url,
+                             cachePolicy: .reloadIgnoringLocalCacheData,
+                             timeoutInterval: networkExecutor.requestTimeout)
+    
+    let task = networkExecutor.GET(request) { [weak self] (result) in
+      guard let self = self else { return }
+      switch result {
+      case .success(let data):
+        self.handlers[url]?(.success(data))
+        self.tasks[url] = nil
+        self.handlers[url] = nil
+      case .failure(let error):
+        // In poor network conditions, the download might fail on first or
+        // second attempt. We retry 2 more times to make sure we don't
+        // prematurely stop the download process.
+        if
+          let attempts = self.numberOfAttempts[url],
+          attempts < 3,
+          let handler = self.handlers[url] {
+          print("NYPLAxisContentDownloader downloading item from \(url.absoluteString) after \(attempts) failed attempts")
+          self.tasks[url] = nil
+          self.handlers[url] = nil
+          self.downloadItem(from: url, handler)
+          return
+        }
+        
+        // Here we perform short-circuiting so that we don't download anything
+        // else after receiving a failure since a book with a missing file is
+        // unusable.
+        self.handlers[url]?(.failure(error))
+        self.tasks.forEach {
+          $1.cancel()
+        }
+        self.tasks.removeAll()
+        self.handlers.removeAll()
+        self.errored.value = true
+      }
+    }
+    
+    return task
+  }
+  
+}
+
+private class ThreadSafeContainer {
+  fileprivate let queue: DispatchQueue
+  
+  init(queue: DispatchQueue) {
+    self.queue = queue
+  }
+  
+}
+
+private class ThreadSafeDictionaryContainer<S: Hashable,T>: ThreadSafeContainer {
+  
+  private var dictionary: [S:T]
+  
+  init(dictionary:[S:T] = [:], queue: DispatchQueue) {
+    self.dictionary = dictionary
+    super.init(queue: queue)
+  }
+  
+  subscript(key: S) -> T? {
+    get {
+      return queue.sync { [weak self] in
+        self?.dictionary[key]
+      }
+    }
+    set(newValue) {
+      queue.sync { [weak self] in
+        self?.dictionary[key] = newValue
+      }
+    }
+  }
+  
+  func forEach(closure: (S, T) -> Void) {
+    queue.sync {
+      self.dictionary.forEach {
+        closure($0, $1)
+      }
+    }
+  }
+  
+  func removeAll() {
+    queue.sync { [weak self] in
+      self?.dictionary.removeAll()
+    }
+  }
+  
+}
+
+private class ThreadSafeValueContainer<T>: ThreadSafeContainer {
+  private var _value: T
+  
+  init(value: T, queue: DispatchQueue) {
+    self._value = value
+    super.init(queue: queue)
+  }
+  
+  var value: T? {
+    get {
+      return queue.sync { [weak self] in
+        self?._value
+      }
+    }
+    set(newValue) {
+      queue.sync { [weak self] in
+        guard let newValue = newValue else { return }
+        self?._value = newValue
+      }
+    }
+  }
+  
+}
--- a/Simplified/AxisService/NYPLAxisService.swift
+++ b/Simplified/AxisService/NYPLAxisService.swift
@@ -19,8 +19,26 @@ import Foundation
                    forDownloadTask downloadtask: URLSessionDownloadTask) -> Bool
 }
 
-@objc
-class NYPLAxisService: NSObject {
+/**
+ Here's how book downloading works:
+ 1. When we send a download request to server, we get a file with `isbn` & `book_vault_uuid` key..
+ 2. If the book's management rights are the ones with axis, we initiate NYPLAxisService and call the
+ `NYPLAxisService::fulfillAxisLicense:downloadTask` method.
+ 3. Using `isbn` & `book_vault_uuid` from the file, we create a url to download a license file for the axis
+ book using the `NYPLAxisService::downloadLicense` method
+ 4. If the license download is successful, we download two files, `Container.xml` & `Encryption.xml`,
+ from Axis using the isbn key. (`NYPLAxisService::downloadContainer` & `NYPLAxisService::downloadEncryption`)
+ 5. From the `container.xml file`, we extract the url for a `package.opf` file and download the
+ package.opf file (`NYPLAxisService::downloadPackage`).
+ 6. From the `package.opf` file (which is more of an xml really), we extract urls for all the required files for
+ the book (e.g. chapter1.xtml, cover.html, fonts, images, styles etc.). We pass those urls in a for loop and tell
+ the downloader to download all the files. (`NYPLAxisService::downloadAssetsFromPackage`)
+ 7. When all the files are downloaded, we consider this as book download success and call the method associated
+ with success on our NYPLBookDownloadBroadcasting delegate. (`replaceBook:withFileAtURL:forDownloadTask`)
+ 8. If we experience a failure at any point during the download process, we stop the process, delete the
+ downloaded files, and call our delegate's `failDownloadWithAlert:` method.
+ */
+@objc class NYPLAxisService: NSObject {
   private weak var delegate: NYPLBookDownloadBroadcasting?
   private let isbn: String
   private let bookVaultId: String
@@ -34,6 +52,15 @@ class NYPLAxisService: NSObject {
   private let axisKeysProviding: NYPLAxisKeysProviding
   private let dispatchGroup: DispatchGroup
   private let assetWriter: NYPLAssetWriting
+  private var downloader: NYPLAxisContentDownloading?
+  private var cypher: NYPLRSACryptographing?
+  private let globalBackgroundSyncronizeDataQueue = DispatchQueue(label: "NYPLAxis")
+  
+  private var shouldContinue: Bool {
+    return globalBackgroundSyncronizeDataQueue.sync { [weak self] in
+      self?.downloader != nil
+    }
+  }
   
   // MARK: - Static Constants
   static let reason = "reason"
@@ -60,7 +87,10 @@ class NYPLAxisService: NSObject {
        packageEndpointproviding: NYPLAxisPackageEndpointProviding,
        packagePathProviding: NYPLAxisPackagePathPrefixProviding,
        axisKeysProviding: NYPLAxisKeysProviding,
-       assetWriter: NYPLAssetWriting) {
+       assetWriter: NYPLAssetWriting,
+       cypher: NYPLRSACryptographing?,
+       dedicatedWriteURL: URL,
+       downloader: NYPLAxisContentDownloading?) {
     
     self.axisKeysProviding = axisKeysProviding
     self.isbn = isbn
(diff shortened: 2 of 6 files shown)


SimplifiedTests/NYPLContentDownloaderTests.swift
```diff
@@ -0,0 +1,156 @@
...
+  }
+
+}
+
+private class MockAxisNetworkExecutor: NYPLAxisNetworkExecuting {
+  
+  var deinitialzed: (() -> Void)?
+  
+  func GET(_ request: URLRequest, completion: @escaping (Result<Data, Error>) -> Void) -> URLSessionDataTask {
+    let task = URLSession.shared.dataTask(with: request)
```
ettore, line 129: I'm struggling to understand this part. This `task` is intentionally not resumed, correct? however you are returning it to the caller, which is the `NYPLContentDownloader` instance at the top of this file, and I think that does actually execute it. Am I reading this right?

I was going to say that it is preferable to avoid hitting the internet in unit tests, but I'm not sure we actually do that here. But if we don't, I don't quite understand the reason for the Timer.

The approach I've taken in the past for mocking server / networking behavior is something like in the NYPLNetworkExecutorMock, but I'm not entirely sure that applies here.

Simplified/AxisService/DefaultAxisNetworkExecutor.swift
```diff
@@ -0,0 +1,40 @@
...
+//
+//  Created by Raman Singh on 2021-04-22.
+//  Copyright © 2021 NYPL Labs. All rights reserved.
+//
+
+import Foundation
+
+#if AXIS
+
+struct DefaultAxisNetworkExecutor: NYPLAxisNetworkExecuting {
```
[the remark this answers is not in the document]
singhraman4282: Done

Simplified/AxisService/NYPLContentDownloader.swift
```diff
@@ -0,0 +1,76 @@
...
+                    _ completion: @escaping (Result<Data, Error>) -> Void) {
+    
+    // We dont want to download anything if it's already being downloaded or if
+    // we experienced an error downloading something before.
+    
+    guard !errored, tasks[url] == nil, handlers[url] == nil else {
+      return
+    }
+    
+    handlers[url] = completion
```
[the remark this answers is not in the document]
singhraman4282: I created a separate container class that takes care of thread safety. Will highly appreciatte yours and @ErnestFan 's input on this.

Simplified/AxisService/NYPLContentDownloader.swift
```diff
@@ -0,0 +1,76 @@
...
+  }
+  
+  private func getTask(for url: URL) -> URLSessionDataTask {
+    
+    let request = URLRequest(url: url,
+                             cachePolicy: .reloadIgnoringLocalCacheData,
+                             timeoutInterval: 30.0)
+    
+    let task = networkExecuter.GET(request) { [weak self] (result) in
+      guard let self = self else { return }
```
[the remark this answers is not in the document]
singhraman4282: Great catch. Fixed.

Simplified/AxisService/NYPLContentDownloader.swift
```diff
@@ -0,0 +1,137 @@
...
+    
+    return task
+  }
+  
+}
+
+private class ThreadSafeDictionaryContainer<S: Hashable,T> {
+  
+  private var dictionary: [S:T]
+  private let queue = DispatchQueue(label: "AxixThreadSafeDictionaryContainer")
```
ErnestFan, line 81: minor: typo - Axis, also on the one below

Simplified/AxisService/NYPLContentDownloader.swift
```diff
@@ -0,0 +1,137 @@
...
+import Foundation
+
+protocol NYPLAxisContentDownloading {
+  func downloadItem(from url: URL,
+                    _ completion: @escaping (Result<Data, Error>) -> Void)
+}
+
+
+/// A content downloader which stops downloading content upon receiving a failure.
+class NYPLContentDownloader: NYPLAxisContentDownloading {
```
ErnestFan, line 18: I have a question about this class, is this an Axis specific class or are we going to expand this for other DRMs?
`NYPLContentDownloader` sounds like a general class name but the `init` and `networkExecutor` below are tied to an Axis related protocol.

Simplified/AxisService/NYPLContentDownloader.swift
```diff
@@ -0,0 +1,76 @@
...
+    
+    handlers[url] = completion
+    tasks[url] = self.getTask(for: url)
+  }
+  
+  private func getTask(for url: URL) -> URLSessionDataTask {
+    
+    let request = URLRequest(url: url,
+                             cachePolicy: .reloadIgnoringLocalCacheData,
+                             timeoutInterval: 30.0)
```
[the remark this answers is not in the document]
singhraman4282: Done

Simplified/AxisService/NYPLContentDownloader.swift
```diff
@@ -0,0 +1,76 @@
...
+}
+
+
+/// A content downloader which stops downloading content upon receiving a failure.
+class NYPLContentDownloader: NYPLAxisContentDownloading {
+  
+  private var tasks: [URL: URLSessionDataTask]
+  private var handlers: [URL: (Result<Data, Error>) -> Void]
+  private var errored: Bool
+  let networkExecuter: NYPLAxisNetworkExecuting
```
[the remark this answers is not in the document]
singhraman4282: hahaha done

SimplifiedTests/NYPLContentDownloaderTests.swift
singhraman4282: Now that I think about it, if I have to explain why I did something in a certain way it means I did not do it well, and if we merge that code then maintaining it will be harder than it has to be. Switched to using NYPLNetworkExecutorMock.

Simplified/AxisService/NYPLContentDownloader.swift
singhraman4282: Great point. Changed to NYPLAxisContentDownloader

Simplified/AxisService/NYPLAxisContentDownloader.swift
```diff
@@ -0,0 +1,153 @@
...
+      switch result {
+      case .success(let data):
+        self.handlers[url]?(.success(data))
+        self.tasks[url] = nil
+        self.handlers[url] = nil
+      case .failure(let error):
+        // In poor network conditions, the download might fail on first or
+        // second attempt. We retry 2 more times to make sure we don't
+        // prematurely stop the download process.
+        if
```
singhraman4282, line 65: @ErnestFan @ettore In my experiments with downloading book content while having poor network conditions, there were 130 items that failed to download in the first attempt, and 56 of those items failed to download the second time, but all those 56 items successfully downloaded on the 3rd attempt.

So, to make sure we're doing everything we can to ensure a book download succeeds in all network conditions, should I up the retry count to 5?

Simplified/AxisService/NYPLContentDownloader.swift
singhraman4282: haha great catch thanks

Simplified/AxisService/NYPLAxisContentDownloader.swift
ErnestFan: I am not sure if we should up the retry count or let the user hit the download button again when they feel ready gives better user experience. Let's see what's Ettore's input.

Simplified/AxisService/NYPLAxisContentDownloader.swift
```diff
@@ -0,0 +1,153 @@
...
+          let attempts = self.numberOfAttempts[url],
+          attempts < 3,
+          let handler = self.handlers[url] {
+          print("NYPLAxisContentDownloader downloading item from \(url.absoluteString) after \(attempts) failed attempts")
```
ErnestFan, line 69: Should we use `Log` instead of `print` here?

Simplified/AxisService/NYPLAxisContentDownloader.swift
ettore: this is a tough one, I would say that with poor network conditions all bets are off. With more automatic retries the user might be seeing a spinner for a long time and then it might _still_ fail, and that would be even worse. 

I'd like to understand a few things first: 
- Is there one call to `getTask(for:)` for each of the 130 items? 
- And are those 130 items for just one book?

I'm wondering if these failures are caused because we are flooding the network with too many requests at the same time, but I don't know exactly how these requests are submitted and how Axis works. (In my ignorance I was expecting to download a zip file or something like that.)

Simplified/AxisService/NYPLAxisContentDownloader.swift
```diff
@@ -0,0 +1,153 @@
...
+protocol NYPLAxisContentDownloading {
+  func downloadItem(from url: URL,
+                    _ completion: @escaping (Result<Data, Error>) -> Void)
+}
+
+
+/// A content downloader which stops downloading content upon receiving a failure.
+class NYPLAxisContentDownloader: NYPLAxisContentDownloading {
+  
+  private let tasks: ThreadSafeDictionaryContainer<URL, URLSessionDataTask>
```
ettore, line 20: do these properties need to stay consistent with each other?  the container is nice but will only ensure thread safety of each individual instance, since the queue it's using to synchronize access is different for each one.

Simplified/AxisService/NYPLAxisContentDownloader.swift
singhraman4282: @ettore This is on me. I should had written a brief overview of how Axis works at the top of **AxisService.swift** file. Here's how it works:

1. When we send a download request to server, we get a file with isbn & book_vault_uuid key. (`NYPLMyBooksDownloadCenter::URLSession:session:downloadTask:didFinishDownloadingToURL`)
2. If the book's management rights are the ones with axis, we initiate NYPLAxisService and call the `NYPLAxisService::fulfillAxisLicense:downloadTask` method.
3. Using isbn & book_vault_uuid from the file, we create a url to download a license file for the axis book. (NYPLAxisService::downloadLicense).
4. If the license download is successful, we download two files **Container.xml** & **Encryption.xml**) from Axis using the isbn key. (`NYPLAxisService::downloadContainer` & `NYPLAxisService::downloadEncryption`)
5. From the **container.xml** file, we extract the url for a **package.opf** file and download the **package.opf** file (`NYPLAxisService::downloadPackage`).
6. From the **package.opf** file (which is more of an xml really), we extract urls for all the required files for the book (e.g. chapter1.xtml, cover.html, fonts, images, styles etc.). We pass those urls in a for loop and tell the downloader to download all the files. (`NYPLAxisService::downloadAssetsFromPackage`)
7. When all the files are downloaded, we consider this as book download success.

singhraman4282: So to answer your questions,
_Is there one call to getTask(for:) for each of the 130 items?_
We are storing tasks in a dictionary using the associated url as the key. We achieve two things by doing this.
1. We don't download from the same url twice (if for any reason the package.opf file contains two files with the exact same url). We do this by making sure tasks[url] is nil before initiating the download in `NYPLAxisContentDownloader::downloadItem:from completion`.
2. We can cancel all the pending downloads if a file fails to download (after 3 tries).

_And are those 130 items for just one book?_
Yes. An Axis book folder structure looks like this
fonts (directory)
images (directory)
styles (directory)
toc.ncx
xhtml (directory)

The file size can vary depending on the book.

Simplified/AxisService/NYPLAxisContentDownloader.swift
singhraman4282: We're already logging download failure in `AxisService::downloadItem:from at` method. This print statement is more or less redundant and was placed to make it easy to spot book download failure while debugging.

Simplified/AxisService/NYPLAxisContentDownloader.swift
```diff
@@ -0,0 +1,153 @@
+//
+//  NYPLAxisContentDownloader.swift
+//  Simplified
+//
+//  Created by Raman Singh on 2021-04-22.
+//  Copyright © 2021 NYPL Labs. All rights reserved.
+//
+
+import Foundation
```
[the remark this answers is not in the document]
singhraman4282: All the Axis related files are only available for OpenEbooks target. If you believe it should be behind an  ` #ifdef AXIS` we can do that too.

Simplified/AxisService/NYPLAxisContentDownloader.swift
singhraman4282: These properties are supposed to stay consistent. Now that you mentioned it, I think we should have a single source of truth instead of relying on multiple objects . A single class that stores all these values (handlres, tasks, errored) should be more appropriate. Let me know if that's something you'd prefer and we can go with that.

Simplified/AxisService/NYPLAxisContentDownloader.swift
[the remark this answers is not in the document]
ettore: no that sounds good, either way works. I was just wondering because in other source files you used the `#ifdef`, so I was just checking :)

Simplified/AxisService/NYPLAxisContentDownloader.swift
ettore: thank you for the detailed explanation, this helps a lot. It would be great if you can add this write-up to the code base for future reference.

going back to the original doubt...I'm not opposed to the retry strategy, to me 3 attempts seem reasonable, I was just wondering if we are inducing that by flooding the network. Hundreds of requests could do that -- although come to think of it, I believe NYPLNetworkExecutor has already infrastructure to cap to 4 at a time max. Perhaps verify that, my memory is faulty... 😭

More than everything it would be good to double check the network behavior with a proxy just to be sure we are actually issuing 4 - 8 requests at a time. (I use Charles, it has a free tier, but it costs only $50 and it's worth every penny... there are many others though)

Simplified/AxisService/NYPLAxisContentDownloader.swift
ettore: there are many ways to achieve the same thing, your approach seem fine. In a way you already have a class that store all the values (the Downloader itself) you could just protect access with a NSRecursiveLock... honestly the simplest is perhaps to parametrize your nice container classes on the same queue. This just needs the ThreadContainers to receive a queue from the caller in the initializer, and you're done I think...

Simplified/AxisService/NYPLAxisContentDownloader.swift
singhraman4282: That's a good point. Just checked and even though by the default value is 6 in macOS and 4 in iOS, we're using a URLSessionConfiguration created in NYPLCaching which has this value amped up to maximum 8 connections per host.

`NYPLCaching::makeURLSessionConfiguration:caching:requestTimeout`

Simplified/AxisService/NYPLAxisContentDownloader.swift
singhraman4282: Done

Simplified/AxisService/NYPLAxisService.swift
```diff
@@ -19,8 +19,26 @@ import Foundation
                    forDownloadTask downloadtask: URLSessionDownloadTask) -> Bool
 }
 
-@objc
-class NYPLAxisService: NSObject {
+/**
+ Here's how book downloading works:
```
ettore, line 23: this is awesome, thank you!

Simplified/AxisService/NYPLAxisContentDownloader.swift
```diff
@@ -0,0 +1,163 @@
...
+  private var dictionary: [S:T]
+  
+  init(dictionary:[S:T] = [:], queue: DispatchQueue) {
+    self.dictionary = dictionary
+    super.init(queue: queue)
+  }
+  
+  subscript(key: S) -> T? {
+    get {
+      return queue.sync {
```
ettore, line 115: I think these methods have retain cycles (sorry I didn't notice this sooner): self > queue property > closure > self

Simplified/AxisService/NYPLAxisService.swift
```diff
@@ -34,6 +52,15 @@ class NYPLAxisService: NSObject {
   private let axisKeysProviding: NYPLAxisKeysProviding
   private let dispatchGroup: DispatchGroup
   private let assetWriter: NYPLAssetWriting
+  private var downloader: NYPLAxisContentDownloading?
+  private var cypher: NYPLRSACryptographing?
+  private let globalBackgroundSyncronizeDataQueue = DispatchQueue(label: "NYPLAxis")
+  
+  private var shouldContinue: Bool {
+    return globalBackgroundSyncronizeDataQueue.sync {
```
ettore, line 60: this has also a retain cycle

Simplified/AxisService/NYPLAxisService.swift
```diff
@@ -216,37 +263,35 @@ class NYPLAxisService: NSObject {
...
     
-    // TODO: OE-62: Axis code might need specific requirements for caching and
-    // using the shared network executor will result in whatever we choose to
-    // user there.
+    guard shouldContinue else {
+      return
+    }
     
-    NYPLNetworkExecutor.shared.GET(url) { (result) in
+    downloader?.downloadItem(from: url) { (result) in
```
ettore, line 270: retain cycle, I think, because downloader retains the closure

Simplified/AxisService/NYPLAxisService.swift
```diff
@@ -258,6 +303,22 @@ class NYPLAxisService: NSObject {
...
+  /// Stops the download process, leaves the dispatch group, deletes downloaded files, and notifies the
+  /// delegate of download failure.
+  private func leaveGroupAndStopDownload() {
+    // Already processed failure. No need to do it again.
+    if !shouldContinue {
+      return
+    }
+    
+    globalBackgroundSyncronizeDataQueue.sync {
+      self.downloader = nil
```
ettore, line 315: retain cycle

ettore (approved):
some retain cycles I missed previously, but other than that this looks good

Simplified/AxisService/NYPLAxisService.swift
singhraman4282: You're welcome

Simplified/AxisService/NYPLAxisContentDownloader.swift
singhraman4282: @ettore I originally considered using weak self but didn't since these blocks are short lived and `self` would be released as soon as the value is returned/modified thus breaking the retain cycle. However, not having retain cycles is still better than short lived retain cycles. Fixed.

Simplified/AxisService/NYPLAxisService.swift
singhraman4282: Fixed

Simplified/AxisService/NYPLAxisService.swift
singhraman4282: Fixed

Simplified/AxisService/NYPLAxisService.swift
singhraman4282: Fixed

ErnestFan (approved):

PR #1378, opened 2021-05-12
IOS-193 Fix AdobeDRM decoding after recent changes in R2Shared

ettore:
**What's this do?**
This is addressing the breakage we experienced after upgrading the latest R2 version, `2.0.0-beta.2`, after which opening books with Adobe DRM started failing. The [issue](https://github.com/readium/r2-shared-swift/issues/136) I opened on r2-shared-swift adds more background. This commit follows some of the suggestions from @mickael-menu from that discussion, particularly the requirement to have a `ContentProtectionService` implementation and making sure to close the `fetcher` once done. 
I decided for now to stick with our current fetcher implementation because it has been QA'ed for a while now and so far it has been reliable. The other solutions (such as using `TransformingFetcher` instead of our own custom implementation) are probably more correct but it seems risky introducing more fundamental changes right now.

@vladimirfedorov adding you too since you know more about this code, if you can, it would be great to have your input. Not sure if you're still using adobe drm but if so you might encounter this issue too.

https://jira.nypl.org/browse/IOS-193

I will add instructions in the ticket.

This fix doesn't need to be in 3.6.5, since R2 is still under a feature flag, so it's not blocking release.

yes

yes

@ettore

### DIFF
--- a/Simplified/Reader2/ReaderStackConfiguration/AdobeDRM/NYPLAdobeContentProtectionService.swift
+++ b/Simplified/Reader2/ReaderStackConfiguration/AdobeDRM/NYPLAdobeContentProtectionService.swift
@@ -1,0 +1,51 @@
+//
+//  NYPLAdobeContentProtectionService.swift
+//  Simplified
+//
+//  Created by Ettore Pasquini on 5/11/21.
+//  Copyright © 2021 NYPL Labs. All rights reserved.
+//
+
+#if FEATURE_DRM_CONNECTOR
+
+import Foundation
+import R2Shared
+import R2Streamer
+import R2Navigator
+
+final class NYPLAdobeContentProtectionService: ContentProtectionService {
+  var error: Error?
+  let context: PublicationServiceContext
+
+  init(context: PublicationServiceContext) {
+    self.context = context
+    self.error = nil
+    if let adobeFetcher = context.fetcher as? AdobeDRMFetcher {
+      if let drmError = adobeFetcher.container.epubDecodingError {
+        self.error = NSError(domain: "Adobe DRM decoding error",
+                             code: NYPLErrorCode.adobeDRMFulfillmentFail.rawValue,
+                             userInfo: [
+                              "AdobeDRMContainer error msg": drmError
+                             ])
+      }
+    }
+  }
+
+  /// A restricted publication has a limited access to its manifest and
+  /// resources and can’t be rendered with a Navigator. It is usually
+  /// only used to import a publication to the user’s bookshelf.
+  var isRestricted: Bool {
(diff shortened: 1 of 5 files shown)


Simplified/Reader2/ReaderStackConfiguration/AdobeDRM/NYPLAdobeContentProtectionService.swift
```diff
@@ -0,0 +1,51 @@
...
+
+  /// A restricted publication has a limited access to its manifest and
+  /// resources and can’t be rendered with a Navigator. It is usually
+  /// only used to import a publication to the user’s bookshelf.
+  var isRestricted: Bool {
+    context.publication.ref == nil || error != nil
+  }
+
+  var rights: UserRights {
+    isRestricted ? AllRestrictedUserRights() : UnrestrictedUserRights()
```
mickael-menu, line 42: This might be an issue for Adobe's certification, as the navigator will allow any text copy. I suggest always using `AllRestrictedUserRights()` until you implement your own `UserRights` for ACS. [It's implemented like this with LCP](https://github.com/readium/r2-lcp-swift/blob/b8adc686c392b3279f57259d6904af599b9737d3/readium-lcp-swift/License/License.swift#L67-L105).

ettore: that's a good point... well my understanding was (please correct me I'm wrong, @killerpuppytails )  that we had those rights on SimplyE but Open eBooks has more restrictions  i.e. disabling the "Copy" / "Look up" pop up menu -- that currently we address elsewhere. I have to say I only recently found the more declaratively options for doing so inside R2, so I should file a ticket for improving that.

singhraman4282 (approved):

ettore:
Filed https://jira.nypl.org/browse/IOS-196 for using TransformingFetcher instead of our custom AdobeDRMFetcher.

ErnestFan (approved):

Simplified/Reader2/ReaderStackConfiguration/AdobeDRM/NYPLAdobeContentProtectionService.swift
killerpuppytails: You're absolutely correct, @ettore . Open eBooks has many more restrictions and we explicitly were asked to turn it off by partners. However, since we will be removing Adobe from OE for the next launch, we have more options we can explore. I'll set up a conversation to confirm.

ettore:
(had to rebase to fix merge conflicts)

vladimirfedorov (approved):
Thanks for adding me, looks great

PR #1381, opened 2021-05-13
OE-14: Decrypt book content with Axis DRM

singhraman4282:
**What's this do?**
Prepares book download code to accommodate  decryption of Axis book content

https://jira.nypl.org/browse/OE-14

User should be able to open Axis books

N/A

Will do after adding tests

N/A

Raman Singh verified it

### DIFF
--- a/Simplified.xcodeproj/project.pbxproj
+++ b/Simplified.xcodeproj/project.pbxproj
@@ -221,12 +221,10 @@
 		52592BB821220A1100587288 /* NYPLLocalization.m in Sources */ = {isa = PBXBuildFile; fileRef = 52592BB721220A1100587288 /* NYPLLocalization.m */; };
 		5916E7D4262791B20021CD67 /* NYPLSignInBusinessLogic+Adept.swift in Sources */ = {isa = PBXBuildFile; fileRef = 5916E7D3262791B20021CD67 /* NYPLSignInBusinessLogic+Adept.swift */; };
 		5916E7D5262791B20021CD67 /* NYPLSignInBusinessLogic+Adept.swift in Sources */ = {isa = PBXBuildFile; fileRef = 5916E7D3262791B20021CD67 /* NYPLSignInBusinessLogic+Adept.swift */; };
-		5916E7D6262791B20021CD67 /* NYPLSignInBusinessLogic+Adept.swift in Sources */ = {isa = PBXBuildFile; fileRef = 5916E7D3262791B20021CD67 /* NYPLSignInBusinessLogic+Adept.swift */; };
 		5916E7D7262791B20021CD67 /* NYPLSignInBusinessLogic+Adept.swift in Sources */ = {isa = PBXBuildFile; fileRef = 5916E7D3262791B20021CD67 /* NYPLSignInBusinessLogic+Adept.swift */; };
 		5916E7D8262791B20021CD67 /* NYPLSignInBusinessLogic+Adept.swift in Sources */ = {isa = PBXBuildFile; fileRef = 5916E7D3262791B20021CD67 /* NYPLSignInBusinessLogic+Adept.swift */; };
 		5916E7DF2627922B0021CD67 /* NYPLSignInBusinessLogic+Axis.swift in Sources */ = {isa = PBXBuildFile; fileRef = 5916E7DE2627922B0021CD67 /* NYPLSignInBusinessLogic+Axis.swift */; };
 		5916E7E02627922B0021CD67 /* NYPLSignInBusinessLogic+Axis.swift in Sources */ = {isa = PBXBuildFile; fileRef = 5916E7DE2627922B0021CD67 /* NYPLSignInBusinessLogic+Axis.swift */; };
-		5916E7E12627922B0021CD67 /* NYPLSignInBusinessLogic+Axis.swift in Sources */ = {isa = PBXBuildFile; fileRef = 5916E7DE2627922B0021CD67 /* NYPLSignInBusinessLogic+Axis.swift */; };
 		5916E7E22627922B0021CD67 /* NYPLSignInBusinessLogic+Axis.swift in Sources */ = {isa = PBXBuildFile; fileRef = 5916E7DE2627922B0021CD67 /* NYPLSignInBusinessLogic+Axis.swift */; };
 		5916E7E32627922B0021CD67 /* NYPLSignInBusinessLogic+Axis.swift in Sources */ = {isa = PBXBuildFile; fileRef = 5916E7DE2627922B0021CD67 /* NYPLSignInBusinessLogic+Axis.swift */; };
 		59453E9D263210E400B63E73 /* NYPLAxisNetworkExecutor.swift in Sources */ = {isa = PBXBuildFile; fileRef = 59453E98263210E400B63E73 /* NYPLAxisNetworkExecutor.swift */; };
@@ -239,6 +237,10 @@
 		599CB79C2631EC140024B5F1 /* NYPLAxisKeysProvider.swift in Sources */ = {isa = PBXBuildFile; fileRef = 599CB7972631EC140024B5F1 /* NYPLAxisKeysProvider.swift */; };
 		599CB7AC263200040024B5F1 /* NYPLAxisContentDownloader.swift in Sources */ = {isa = PBXBuildFile; fileRef = 599CB7A7263200040024B5F1 /* NYPLAxisContentDownloader.swift */; };
 		59A78B8A261656C20038BCDC /* NYPLAxisService.swift in Sources */ = {isa = PBXBuildFile; fileRef = 59A78B89261656C20038BCDC /* NYPLAxisService.swift */; };
+		59A80DC4264C9CF400458C84 /* NYPLAxisItemDownloader.swift in Sources */ = {isa = PBXBuildFile; fileRef = 59A80DC3264C9CF400458C84 /* NYPLAxisItemDownloader.swift */; };
+		59C46FAB264B21F800BBCC96 /* NYPLAxisLicenseService.swift in Sources */ = {isa = PBXBuildFile; fileRef = 59C46FAA264B21F800BBCC96 /* NYPLAxisLicenseService.swift */; };
+		59C46FB0264B388200BBCC96 /* NYPLAxisMetadataService.swift in Sources */ = {isa = PBXBuildFile; fileRef = 59C46FAF264B388200BBCC96 /* NYPLAxisMetadataService.swift */; };
+		59C46FB5264B852600BBCC96 /* NYPLAxisPackageService.swift in Sources */ = {isa = PBXBuildFile; fileRef = 59C46FB4264B852600BBCC96 /* NYPLAxisPackageService.swift */; };
 		59DE643326372B0800CFCF14 /* NYPLContentDownloaderTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 59DE643226372B0800CFCF14 /* NYPLContentDownloaderTests.swift */; };
 		59DE644C26372E1300CFCF14 /* NYPLAxisNetworkExecuting.swift in Sources */ = {isa = PBXBuildFile; fileRef = 59DE644B26372E1300CFCF14 /* NYPLAxisNetworkExecuting.swift */; };
 		59DE644D26372E1300CFCF14 /* NYPLAxisNetworkExecuting.swift in Sources */ = {isa = PBXBuildFile; fileRef = 59DE644B26372E1300CFCF14 /* NYPLAxisNetworkExecuting.swift */; };
@@ -402,7 +404,6 @@
 		732F0490263379BA0018A82E /* R2Shared.xcframework in Frameworks */ = {isa = PBXBuildFile; fileRef = 732F048F263379BA0018A82E /* R2Shared.xcframework */; };
 		732F0492263379BA0018A82E /* R2Shared.xcframework in Frameworks */ = {isa = PBXBuildFile; fileRef = 732F048F263379BA0018A82E /* R2Shared.xcframework */; };
 		732F0493263379BA0018A82E /* R2Shared.xcframework in Frameworks */ = {isa = PBXBuildFile; fileRef = 732F048F263379BA0018A82E /* R2Shared.xcframework */; };
-		732F049A26337BAB0018A82E /* NYPLAEToolkit.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = 732F03C3263350DB0018A82E /* NYPLAEToolkit.framework */; };
 		732F04A8263739A50018A82E /* CryptoSwift.xcframework in Frameworks */ = {isa = PBXBuildFile; fileRef = 732F04A7263739A40018A82E /* CryptoSwift.xcframework */; };
 		732F04A9263739A50018A82E /* CryptoSwift.xcframework in Frameworks */ = {isa = PBXBuildFile; fileRef = 732F04A7263739A40018A82E /* CryptoSwift.xcframework */; };
 		732F04AB263739A50018A82E /* CryptoSwift.xcframework in Frameworks */ = {isa = PBXBuildFile; fileRef = 732F04A7263739A40018A82E /* CryptoSwift.xcframework */; };
@@ -497,7 +498,6 @@
 		7369A38B2649C07F0029D8AB /* SwiftSoup.xcframework in Embed Frameworks */ = {isa = PBXBuildFile; fileRef = 732F0441263376530018A82E /* SwiftSoup.xcframework */; settings = {ATTRIBUTES = (CodeSignOnCopy, RemoveHeadersOnCopy, ); }; };
 		7369A38C2649C07F0029D8AB /* ZIPFoundation.xcframework in Embed Frameworks */ = {isa = PBXBuildFile; fileRef = 732F0433263374FC0018A82E /* ZIPFoundation.xcframework */; settings = {ATTRIBUTES = (CodeSignOnCopy, RemoveHeadersOnCopy, ); }; };
 		7369A38D2649C07F0029D8AB /* ZXingObjC.xcframework in Embed Frameworks */ = {isa = PBXBuildFile; fileRef = 732F0480263378EF0018A82E /* ZXingObjC.xcframework */; settings = {ATTRIBUTES = (CodeSignOnCopy, RemoveHeadersOnCopy, ); }; };
-		7369A3912649C0AF0029D8AB /* NYPLAEToolkit.framework in Embed Frameworks */ = {isa = PBXBuildFile; fileRef = 732F03C3263350DB0018A82E /* NYPLAEToolkit.framework */; settings = {ATTRIBUTES = (CodeSignOnCopy, RemoveHeadersOnCopy, ); }; };
 		736A54B8254B3C680031C3AA /* NYPLSAMLHelper.m in Sources */ = {isa = PBXBuildFile; fileRef = 730263AB2540DE4200A53891 /* NYPLSAMLHelper.m */; };
 		736A54BD254B3C6D0031C3AA /* NYPLSignInBusinessLogic+DRM.swift in Sources */ = {isa = PBXBuildFile; fileRef = 737F4A6A254A137900A3C34B /* NYPLSignInBusinessLogic+DRM.swift */; };
 		736A54BE254B3C730031C3AA /* NYPLSignInBusinessLogic+UI.swift in Sources */ = {isa = PBXBuildFile; fileRef = 73A794A525477D8F00C59CC1 /* NYPLSignInBusinessLogic+UI.swift */; };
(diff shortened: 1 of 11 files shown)


Simplified/AxisService/NYPLAxisLicenseService.swift
```diff
@@ -0,0 +1,192 @@
...
+      baseURL: axisKeysProvider.licenseBaseURL,
+      bookVaultId: bookVaultId,
+      clientIP: deviceInfoProvider.clientIP,
+      cypher: cypher,
+      deviceID: deviceInfoProvider.deviceID,
+      isbn: isbn
+    ).generateLicenseURL()
+    
+    let writeURL = parentDirectory
+      .appendingPathComponent(axisKeysProvider.desiredNameForLicenseFile)
```
singhraman4282, line 74: it should be `localLicenseURL`

Simplified/AxisService/NYPLAxisItemDownloader.swift
```diff
@@ -0,0 +1,103 @@
...
+  var shouldContinue: Bool { get }
+  func downloadItem(from url: URL, at writeURL: URL)
+  func leaveGroupAndStopDownload()
+}
+
+class NYPLAxisItemDownloader: NYPLAxisItemDownloading {
+  
+  // Content Downloading constants
+  static private let writeFailureSummary = "AXIS: Failed writing item to specified write URL"
+  static private let failedDownloadingItemSummary = "AXIS: Failed downloading item"
```
ErnestFan, line 27: `writeFailureSummary` should follow the naming convention of this property

singhraman4282: Agreed. Fixed.

ettore:
about that failed test, since it's a preexisting test and you're reporting it succeeds locally, maybe there's a memory effect in it somehow, perhaps caused by not nilling out all variables in `tearDown`? or maybe there are tests running in parallel that interfere with each other? I'm just speculating because I don't see the test. The other option is to add printouts to see where it's failing exactly.

Simplified/AxisService/NYPLAxisLicenseService.swift
```diff
@@ -0,0 +1,190 @@
...
+      let jsonObject = try? JSONSerialization.jsonObject(with: data, options: .fragmentsAllowed),
+      let json = jsonObject as? [String: Any],
+      let encryptionNode = json["encryption"] as? [String: Any],
+      let userKey = encryptionNode["user_key"] as? [String: Any],
+      let encryptedBookVaultId = userKey["key_check"] as? String,
+      let base64Data = Data(base64Encoded: encryptedBookVaultId),
+      let decryptedBookVaultData = cypher.decryptWithPKCS1_OAEP(base64Data),
+      let decryptedBookVaultId = String(data: decryptedBookVaultData, encoding: .utf8)
+    else {
+      logLicenseError(reason: "Axis failed to validate license")
```
ErnestFan, line 96: The string is already defined as `licenseValidationFailureSummary`, so we should use that instead.
Same for below.

Simplified/AxisService/NYPLRSACypher.swift
```diff
@@ -126,52 +128,95 @@ struct NYPLRSACypher: NYPLRSACryptographing {
...
     }
+    return decryptData as Data
+  }
+  
+  func decryptWithAES(_ data: Data, key: Data) -> Data? {
+    let content = data.subdata(in: 4..<data.count)
+    let keyLength = key.count
+    let validKeyLengths = [kCCKeySizeAES128, kCCKeySizeAES192, kCCKeySizeAES256]
+    if (validKeyLengths.contains(keyLength) == false) {
+        return nil
```
ErnestFan, line 188: minor: indentation, same for the if statement below

Simplified/AxisService/NYPLAxisService.swift
```diff
@@ -72,253 +56,70 @@ import Foundation
...
-      self.downloadEncryption(forBook: self.book)
-      self.downloadContainer(forBook: self.book)
-      self.downloadPackage(forBook: self.book)
-      self.downloadAssetsFromPackage(forBook: self.book)
-      
-      self.dispatchGroup.notify(queue: .global(qos: .utility)) {
-        _ = self.delegate?.replaceBook(self.book,
-                                       withFileAtURL: self.dedicatedWriteURL,
-                                       forDownloadTask: downloadTask)
+      self.downloadAndValidateLicense()
```
ettore, line 85: these tasks will be running concurrently, are there concurrency issues in doing that? I suppose as long as `downloadMetadataContent` and `downloadPackage` are thread safe we are good?

Simplified/AxisService/NYPLAxisService.swift
```diff
@@ -72,253 +56,70 @@ import Foundation
...
+      self.downloadMetadataContent()
+      self.downloadPackage()
+      self.axisItemDownloader.dispatchGroup.notify(queue: .global(qos: .utility)) {
+        // weak self will result in deallocation before book completion
```
ettore, line 89: which object would get dealloced? If it's `self` this likely means there's an ownership problem on the service.

Simplified.xcodeproj/project.pbxproj
```diff
@@ -497,7 +498,7 @@
 		7369A38B2649C07F0029D8AB /* SwiftSoup.xcframework in Embed Frameworks */ = {isa = PBXBuildFile; fileRef = 732F0441263376530018A82E /* SwiftSoup.xcframework */; settings = {ATTRIBUTES = (CodeSignOnCopy, RemoveHeadersOnCopy, ); }; };
 		7369A38C2649C07F0029D8AB /* ZIPFoundation.xcframework in Embed Frameworks */ = {isa = PBXBuildFile; fileRef = 732F0433263374FC0018A82E /* ZIPFoundation.xcframework */; settings = {ATTRIBUTES = (CodeSignOnCopy, RemoveHeadersOnCopy, ); }; };
 		7369A38D2649C07F0029D8AB /* ZXingObjC.xcframework in Embed Frameworks */ = {isa = PBXBuildFile; fileRef = 732F0480263378EF0018A82E /* ZXingObjC.xcframework */; settings = {ATTRIBUTES = (CodeSignOnCopy, RemoveHeadersOnCopy, ); }; };
-		7369A3912649C0AF0029D8AB /* NYPLAEToolkit.framework in Embed Frameworks */ = {isa = PBXBuildFile; fileRef = 732F03C3263350DB0018A82E /* NYPLAEToolkit.framework */; settings = {ATTRIBUTES = (CodeSignOnCopy, RemoveHeadersOnCopy, ); }; };
+		7369A3912649C0AF0029D8AB /* (null) in Embed Frameworks */ = {isa = PBXBuildFile; settings = {ATTRIBUTES = (CodeSignOnCopy, RemoveHeadersOnCopy, ); }; };
```
ErnestFan, line 501: This probably needs a clean up? Same for line 1564

Simplified/AxisService/NYPLAxisService.swift
```diff
@@ -378,34 +180,67 @@ extension NYPLAxisService {
...
+              isbn: isbn)
+  }
+  
+  /// Convenience initialzer to be used for testing
+  /// - Parameters:
+  ///   - axisKeysProvider: NYPLAxisKeysProviding object
+  ///   - book: NYPLBook that is to be downloaded
+  ///   - bookVaultId: BookVaultID for the given book
+  ///   - cypher: NYPLRSACryptographing object to be used for downloading license
+  ///   - delegate: An object confirming to NYPLBookDownloadBroadcasting protocol
```
ettore, line 199: minor: instead of repeating the declaration from the code, can you say why we need this delegate or what are we delegating?

Generally speaking, if the docs don't provide anything additional to the code (because the meaning is obvious and there's nothing to clarify), it's fine to omit them. We don't `have to` put something in just because.

another example: I can tell `bookVaultId` is the vaultID for the given book by just reading the declaration, but what's less obvious is the meaning of "vault" in this context. IMO that's the kind of stuff that makes great docs useful. I like what you did for `cypher` and `book` for example.

Simplified/AxisService/NYPLRSACypher.swift
```diff
@@ -126,52 +128,95 @@ struct NYPLRSACypher: NYPLRSACryptographing {
...
-        NYPLRSACypher
-          .logFailedEncryptionError(NYPLRSACypher.nilEncryptedDataFailure)
+        .logFailedInitializationError(
+          "Failed to get exponent and or modulus for public key")
         return nil
     }
     
-    return (encryptedData as Data).base64EncodedString()
+    let modulus = mod.base64EncodedString()
+    guard modulus.count >= 342 else {
```
ettore, line 142: why 342? can you add a comment about this hardcoded constant?

Simplified/AxisService/NYPLRSACypher.swift
```diff
@@ -126,52 +128,95 @@ struct NYPLRSACypher: NYPLRSACryptographing {
...
+  static private func parsePublicSecKey(publicKey: SecKey) -> (mod: Data, exp: Data)? {
     
-    guard let messageData = _messageData else {
-      NYPLRSACypher
-        .logFailedDecryptionError(NYPLRSACypher.stringToDataConversionFailure)
+    guard let pubAttributes = SecKeyCopyAttributes(publicKey) as? [String: Any] else {
       return nil
     }
     
+    let keySize = pubAttributes[kSecAttrKeySizeInBits as String] as! Int
```
ettore, line 158: why are these 2 force unwrap safe? (I'm assuming they are)

Simplified/AxisService/NYPLRSACypher.swift
```diff
@@ -126,52 +128,95 @@ struct NYPLRSACypher: NYPLRSACryptographing {
...
+    let pubData = pubAttributes[kSecValueData as String] as! Data
+    var modulus = pubData.subdata(in: 8..<(pubData.count - 5))
```
ettore, line 160: I don't know what's going on with these magic numbers...

or why we removeFirst at line 164

Not sure if it's part of a standard  algorithm, if so it would be good to paste a link to that reference, otherwise this is very cryptic.

Simplified/AxisService/NYPLRSACypher.swift
```diff
@@ -126,52 +128,95 @@ struct NYPLRSACypher: NYPLRSACryptographing {
...
+    }
+
+    let ivSize = kCCBlockSizeAES128;
+    let clearLength = size_t(content.count - ivSize)
+    var clearData = Data(count:clearLength)
+
+    var numBytesDecrypted :size_t = 0
+    let options   = CCOptions(kCCOptionPKCS7Padding)
+
+    let cryptStatus = clearData.withUnsafeMutableBytes {cryptBytes in
```
ettore, line 198: if you have reference for this process, it would be great to add it to the code -- I'm sure this works, but I have no idea why.

ettore (changes requested):

Simplified/AxisService/NYPLAxisLicenseService.swift
```diff
@@ -0,0 +1,190 @@
...
+  /// however store the book_vault_id and isbn for downloading license again when the user wants to open
+  /// the book for reading.
+  func validateLicense() {
+    axisItemDownloader.dispatchGroup.wait()
+    axisItemDownloader.dispatchGroup.enter()
+    guard
+      let data = try? Data(contentsOf: localLicenseURL),
+      let jsonObject = try? JSONSerialization.jsonObject(with: data, options: .fragmentsAllowed),
+      let json = jsonObject as? [String: Any],
+      let encryptionNode = json["encryption"] as? [String: Any],
```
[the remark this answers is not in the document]
singhraman4282: Great catch. Done

Simplified/AxisService/NYPLAxisLicenseService.swift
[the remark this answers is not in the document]
singhraman4282: Agreed. Done

singhraman4282: Great catch. Done

Simplified/AxisService/NYPLAxisLicenseService.swift
```diff
@@ -0,0 +1,190 @@
...
+      let decryptedBookVaultId = String(data: decryptedBookVaultData, encoding: .utf8)
+    else {
+      logLicenseError(reason: "Axis failed to validate license")
+      axisItemDownloader.leaveGroupAndStopDownload()
+      return
+    }
+    
+    let isValid = decryptedBookVaultId == bookVaultId
+    guard isValid else {
+      logLicenseError(reason: "Axis failed to validate license")
```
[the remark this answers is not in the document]
singhraman4282: Yes. That would make a huge difference. Done

Simplified/AxisService/NYPLAxisLicenseService.swift
```diff
@@ -0,0 +1,190 @@
...
+    }
+    
+    defer {
+      axisItemDownloader.dispatchGroup.leave()
+    }
+    
+    return Data(base64Encoded: encryptedContentValue)
+  }
+  
+  private func logLicenseError(reason: String) {
```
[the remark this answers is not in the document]
singhraman4282: That would be awesome. Done

Simplified/AxisService/NYPLAxisPackageService.swift
```diff
@@ -0,0 +1,118 @@
...
+    else {
+      return
+    }
+    axisItemDownloader.dispatchGroup.enter()
+    let packageLocation = parentDirectory.appendingPathComponent(endpoint)
+    guard
+      let data = try? Data(contentsOf: packageLocation),
+      let axisXML = NYPLAxisXML(data: data)
+    else {
+      // TODO: Log error here
```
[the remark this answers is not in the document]
singhraman4282: 😬 

Done

Simplified/AxisService/NYPLRSACypher.swift
singhraman4282: Done

Simplified/AxisService/NYPLRSACypher.swift
singhraman4282: Fixed

Simplified/AxisService/NYPLRSACypher.swift
singhraman4282: Agreed. Added documentation along with the link to where I got the inspiration from.

Simplified/AxisService/NYPLRSACypher.swift
singhraman4282: Fixed

Simplified/AxisService/NYPLRSACypher.swift
singhraman4282: done

Simplified.xcodeproj/project.pbxproj
singhraman4282: Fixed

Simplified.xcodeproj/project.pbxproj
```diff
@@ -4298,250 +4274,6 @@
 			);
 			runOnlyForDeploymentPostprocessing = 0;
 		};
-		7347F037245A4DE200558D7F /* Sources */ = {
```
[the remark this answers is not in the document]
singhraman4282: It looks like it happened during rebasing my current commit on the latest develop branch

Simplified/AxisService/NYPLAxisService.swift
singhraman4282: Each method in license service depends on dispatch group. Before starting, it waits for the dispatch group to finish doing whatever it was doing, enters the dispatch group, and doesn't leave it until it's done or encounters an error. This ensures that no two methods run at the same time.

singhraman4282: All the methods in packageService and metadataService are controlled by dispatchGroup. Before executing, they wait for the dispatchGroup to finish whatever it was doing, then enter the dispatchGroup, and only leave it upon success or failure. This ensures that package service doesn't actually start until metadataService is finished.

Simplified/AxisService/NYPLAxisService.swift
singhraman4282: It's AxisService. The reason for that is the only other strong reference to AxisService exists inside NYPLMyBooksDownloadCenter::URLSession:downloadTask:didFinishLoadingToURL method because we need a different AxisService for each book download.

Simplified/AxisService/NYPLAxisService.swift
singhraman4282: Done

Simplified.xcodeproj/project.pbxproj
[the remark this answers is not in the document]
ettore: I checked that key on develop and it's not there anymore, so I think this just means your feature branch is behind and your `task/OE-14` branch includes more changes from `develop`. This is a bit unusual but it's ok. 

The more usual git-flow process does this instead:
- routinely merge `develop` on the feature branch
- start or rebase your new commit branches (like `task/OE-14`) on top of the _feature_ branch, not `develop`.

This also reduces the amount of changes in PR -- for example this would not have showed up here.

Simplified/AxisService/NYPLAxisService.swift
```diff
@@ -25,42 +25,26 @@ import Foundation
...
+  private weak var delegate: NYPLBookDownloadBroadcasting?
   
-  private var shouldContinue: Bool {
-    return globalBackgroundSyncronizeDataQueue.sync { [weak self] in
-      self?.downloader != nil
-    }
-  }
   
   // MARK: - Static Constants
   static let reason = "reason"
```
ettore, line 50: since you're using this in crashlytics reports, can you change it to "axisDRMReason" or something like that? So that it's easier to search for it.

Simplified/AxisService/NYPLAxisService.swift
ettore: do you mean the reference [created here](https://github.com/NYPL-Simplified/Simplified-iOS/blob/01749343a8eee678e1e711d5b91c0fc720688c0e/Simplified/MyBooks/NYPLMyBooksDownloadCenter.m#L352)? If so I see what you mean: the object created there is immediately released so there's no remaining strong reference beside the capture done by the closure here.

This works as long as this closure is kept alive: once the dispatch group releases the closure the service will also be dealloced and that's probably not what you want. If you didn't observe this it's likely because the run time wasn't aggressive in reclaiming memory. 

Either way though the ownership hierarchy here looks reversed to me: it shouldn't be a low level closure holding the controlling service alive.

Since this PR is already big, I'd prefer actually if you file a ticket to fix this separately rather than fixing it now.

Simplified/AxisService/NYPLRSACypher.swift
```diff
@@ -126,52 +128,131 @@ struct NYPLRSACypher: NYPLRSACryptographing {
...
+   
+   These integers are padded with a zero byte if they are positive, and have a leading 1-bit. Therefore we
+   might get one extra byte. If that happens, we just cut it away.
+   
+   Here's what the data looks like
+   First 8 bits - some bits we do not need. Anything after that upto the 5th last bit is modulus. And the last 3
+   bits are the exponent.
+   
+   Our public key is 2048 bits long. Modulus for a public key is 1/8th the key size. But since we might have a
+   leading 1-bit, in which case our modulus will be 257 bits, we remove the extra leading bit.
```
ettore, line 174: from reading the SO answer, is it possible you meant "257 _bytes_", and then "we remove the extra leading _byte_"?

Simplified/AxisService/NYPLRSACypher.swift
```diff
@@ -126,52 +128,131 @@ struct NYPLRSACypher: NYPLRSACryptographing {
...
+    modulus INTEGER,
+    publicExponent INTEGER
+   }
+   
+   These integers are padded with a zero byte if they are positive, and have a leading 1-bit. Therefore we
+   might get one extra byte. If that happens, we just cut it away.
+   
+   Here's what the data looks like
+   First 8 bits - some bits we do not need. Anything after that upto the 5th last bit is modulus. And the last 3
+   bits are the exponent.
```
ettore, line 171: from reading the stack overflow doc, is this maybe 3 bytes instead of 3 bits?

Simplified/AxisService/NYPLRSACypher.swift
```diff
@@ -126,52 +128,131 @@ struct NYPLRSACypher: NYPLRSACryptographing {
...
+   SEQUENCE {
+    modulus INTEGER,
+    publicExponent INTEGER
+   }
+   
+   These integers are padded with a zero byte if they are positive, and have a leading 1-bit. Therefore we
+   might get one extra byte. If that happens, we just cut it away.
+   
+   Here's what the data looks like
+   First 8 bits - some bits we do not need. Anything after that upto the 5th last bit is modulus. And the last 3
```
ettore, line 170: are these first 8 bits the same of the padding 0-byte mentioned at line 166?

ettore (approved):

Simplified/AxisService/NYPLRSACypher.swift
singhraman4282: yes.

ErnestFan (approved):

PR #1383, opened 2021-05-14
IOS-190 Access property via accessor to ensure consistency

ettore:
**What's this do?**
this crash is caused by an unrecognized message sent to an invalid `DefaultAudiobookManager` instance. While the class has some possible threading issues, I don't think they are the cause of this crash, because the incriminated `manager` property is being deallocated as part of the timer behavior (which appears correct to me). However, at the point of crash we were accessing the lower level pointer instead of the property, and the property was not `atomic`, so it is possible (depending on the optimizations done by the runtime) to access a garbage pointer. By using `atomic` and the property we should avoid this situation.

If the crash persists, I think it will be time (no pun intended 😄 ) for a larger refactor (ensuring thread safety, perhaps avoiding the timer entirely) that I don't have bandwidth for right now.

https://jira.nypl.org/browse/IOS-190

will add in ticket

no, this will go out after 3.6.5

not needed right now

n/a

@ettore

### DIFF
--- a/Simplified/Book/UI/NYPLBookCellDelegate.m
+++ b/Simplified/Book/UI/NYPLBookCellDelegate.m
@@ -30,7 +30,7 @@
 
 @property (nonatomic) NSTimer *timer;
 @property (nonatomic) NYPLBook *book;
-@property (nonatomic) id<AudiobookManager> manager;
+@property DefaultAudiobookManager *manager;
 @property (nonatomic, weak) AudiobookPlayerViewController *audiobookViewController;
 @property (strong) NSLock *refreshAudiobookLock;
 
@@ -238,9 +238,9 @@
       AudiobookMetadata *const metadata = [[AudiobookMetadata alloc]
                                            initWithTitle:book.title
                                            authors:@[book.authors]];
-      id<AudiobookManager> const manager = [[DefaultAudiobookManager alloc]
-                                            initWithMetadata:metadata
-                                            audiobook:audiobook];
+      DefaultAudiobookManager *const manager = [[DefaultAudiobookManager alloc]
+                                                initWithMetadata:metadata
+                                                audiobook:audiobook];
       manager.refreshDelegate = self;
 
       AudiobookPlayerViewController *const audiobookVC = [[AudiobookPlayerViewController alloc]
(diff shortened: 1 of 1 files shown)


Simplified/Book/UI/NYPLBookCellDelegate.m
```diff
@@ -330,12 +331,14 @@ - (void)scheduleTimerForAudiobook:(NYPLBook *)book
...
-                                              selector:@selector(pollAudiobookReadingLocation)
+                                              selector:@selector(pollAudiobookReadingLocation:)
                                               userInfo:nil
                                                repeats:YES];
 }
 
-- (void)pollAudiobookReadingLocation
+// non-thread safe: must be called on the same thread as
+// scheduleTimerForAudiobook:manager:viewController: each time.
+- (void)pollAudiobookReadingLocation:(NSTimer *)timer
```
ErnestFan, line 341: Does adding the `NSTimer` as the parameter do anything here? Just curious

ErnestFan (approved):

Simplified/Book/UI/NYPLBookCellDelegate.m
ettore: no, it does nothing... I merely added it to comply with the docs. But it worked without too.

Simplified/Book/UI/NYPLBookCellDelegate.m
```diff
@@ -30,7 +30,7 @@ @interface NYPLBookCellDelegate () <RefreshDelegate>
 
 @property (nonatomic) NSTimer *timer;
 @property (nonatomic) NYPLBook *book;
-@property (nonatomic) id<AudiobookManager> manager;
+@property DefaultAudiobookManager *manager;
```
singhraman4282, line 33: There's a huge change that I'm missing something here but it looks like we're breaking the dependency inversion principle.

ettore: My changes are pretty minimal, so I'm guessing you mean with the existing implementation?

singhraman4282: oops sorrry I meant CHANCE not change @ettore

singhraman4282 (approved):

singhraman4282 (approved):

Simplified/Book/UI/NYPLBookCellDelegate.m
ettore: hey Raman can you explain what you mean? this is for a private ("private" in objc terms that is) property so calling things with their actual name is not making things more tightly dependent than using a protocol.

singhraman4282: you're right. I missed the 'private' part.

PR #1384, opened 2021-05-18
iOS-36 Update OPDS2 parser

ErnestFan:
**What's this do?**
Refactor OPDS2 related class
Update parser, sample json and unit tests

[ios-36](https://jira.nypl.org/browse/IOS-36)

Not ready to be tested

N/A

N/A

No

@ErnestFan


ErnestFan:
Below are some references to make it less confusing when reviewing this PR
- Renaming `OPDS2CatalogsFeed` to `OPDS2LibraryRegistryFeed`
- Renaming `OPDS2Publication` to `OPDS2LibraryCatalog`

Since we are adopting the new API schema for LibraryRegistry, the old json data in `OPDS2CatalogsFeed.json` would not work with the new parser. Therefore, I have only converted the json data of a couple libraries that we are using in unit tests.

singhraman4282 (approved):

SimplifiedTests/Mocks/NYPLLibraryAccountsProviderMock.swift
```diff
@@ -60,21 +60,27 @@ class NYPLLibraryAccountMock: NSObject, NYPLLibraryAccountsProvider {
...
+  func createOPDS2LibraryCatalog() -> OPDS2LibraryCatalog {
     let link = OPDS2Link(href: "href\(arc4random())",
       type: "type\(arc4random())",
       rel: "rel\(arc4random())",
       templated: false,
       displayNames: nil,
       descriptions: nil)
-    let metadata = OPDS2Publication.Metadata(updated: Date(),
+    let subject = OPDS2LibraryCatalog.Metadata.Subject(name: "subjectName",
+                                                    code: "subjectCode",
```
ettore, line 71: minor: indentation, also below

SimplifiedTests/OPDS2LibraryRegistryFeed.json
```diff
@@ -0,0 +1,347 @@
...
+      "type":"application/opds+json;profile=http://librarysimplified.org/rel/profile/directory"
+    }
+  ],
+  "catalogs":[
+    {
+      "metadata":{
+        "title":"Boston Public Library",
+        "identifier":"urn:uuid:edef2358-9f6a-4ce6-b64f-9b351ec68ac4",
+        "modified":"2021-01-12T17:48:06Z",
+        "schema:distance":"14 km",
```
ettore, line 19: @leonardr would it be possible to put the unit of measure in a separate field? The client will have to do conversions between km and miles depending on the user's locale settings, so it would save us some parsing. If it's always going to be in km I suppose we could even omit the `km` part, but that seems less future-proof.

ettore (approved):
Looks good to me

SimplifiedTests/OPDS2LibraryRegistryFeed.json
leonardr: This comes from the definition of https://schema.org/Distance:

```
Properties that take Distances as values are of the form '<Number> <Length unit of measure>'. E.g., '7 ft'.
```

So we would need to make it a custom field rather than schema:distance. I would prefer to leave things as is unless parsing out the unit of measurement is a big problem.

PR #1385, opened 2021-05-20
Wrote tests for license service

singhraman4282:
**What's this do?**
Tests for NYPLAxisLicenseService

Cannot be tested by QA

N/A

N/A

N/A
Raman Singh verified it


SimplifiedTests/NYPLAxisLicenseServiceTests.swift
```diff
@@ -0,0 +1,218 @@
...
+    }
+    wait(for: [validationExpectation], timeout: 10.0)
+  }
+  
+  func testLicenseShouldSaveBookInfoForLaterUse() {
+    let savedInfoExpectation = XCTestExpectation(
+      description: "Book info should be saved for later use")
+    
+    
+    let cypher = NYPLRSACypher()!
```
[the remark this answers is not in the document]
singhraman4282: That's a very good point. Done.

ErnestFan (approved):

SimplifiedTests/NYPLAxisLicenseServiceTests.swift
```diff
@@ -0,0 +1,204 @@
...
+      }
+    }
+    wait(for: [validationExpectation], timeout: 10.0)
+  }
+  
+  func testLicenseShouldSaveBookInfoForLaterUse() {
+    let savedInfoExpectation = XCTestExpectation(
+      description: "Book info should be saved for later use")
+    
+    let license = createLicenseService(bookVaultId: TestBook.book1.bookVaultId,
```
ettore, line 107: can you rename `license` to `service`? if I understand correctly that's what we are creating here...

SimplifiedTests/NYPLAxisLicenseServiceTests.swift
```diff
@@ -0,0 +1,204 @@
...
+        let data = try? Data(contentsOf: designatedBookInfoURL),
+        let jsonObject = try? JSONSerialization.jsonObject(with: data, options: .fragmentsAllowed),
+        let json = jsonObject as? [String: String],
+        let bookVaultId = json[self.axisKeysProvider.bookVaultKey],
+        bookVaultId == TestBook.book1.bookVaultId,
+        let isbn = json[self.axisKeysProvider.isbnKey],
+        isbn == TestBook.book1.isbn
+      {
+        savedInfoExpectation.fulfill()
+      }
```
ettore, line 126: should we do an `else XCTFail()` so that we don't wait 10 seconds for the expectation to time out? Not sure if this can apply to other tests, but if so this would speed up test execution.

It's also probably ok to reduce the timeout?

SimplifiedTests/NYPLAxisLicenseServiceTests.swift
```diff
@@ -0,0 +1,204 @@
...
+private struct TestBook {
+  let isbn: String
+  let bookVaultId: String
+  
+  static let book1 = TestBook(isbn: "9781619634015", bookVaultId: "23593B59-F4A2-4722-A3D3-4DF348037757")
+  static let book2 = TestBook(isbn: "9781250109699", bookVaultId: "872D8384-7F33-4B2C-81A5-267415C0170B")
+}
+
+
+private class CustomNYPLAxisItemDownloader: NYPLAxisItemDownloader {
```
ettore, line 182: since this is subclassing the actual downloader class, does this mean these unit tests actually reach the network? I see you have a `NYPLAxisNetworkExecutorMock` so if the downloader is using that under the hood we are good (the internet can be unpredictable and tests will start failing randomly), but I don't see how the downloader is configured with it. I'm likely missing something, just checking.

SimplifiedTests/NYPLContentDownloaderTests.swift
```diff
@@ -122,14 +125,14 @@ class NYPLContentDownloaderTests: XCTestCase {
       description: "NYPLAxisContentDownloader and NYPLAxisNetworkExecutor should deinitialize upon failure")
     expectation.expectedFulfillmentCount = 2
     
-    
-    var customContentDownloader: CustomNYPLContentDownloader? = CustomNYPLContentDownloader(networkExecuting: MockAxisNetworkExecutor())
+    var customContentDownloader: CustomNYPLContentDownloader? = CustomNYPLContentDownloader(
```
ettore, line 128: minor: we shouldn't need `: CustomNYPLContentDownloader?`

ettore (commented):
looks fine except one question about networking -- also could you group all the axis DRM unit test files into one directory? to avoid reorgs down the line

SimplifiedTests/NYPLAxisLicenseServiceTests.swift
singhraman4282: The only time licenseService actually downloads something is when we download the license. Since we're mocking downloaded license in our tests, we're not making any real network calls.

SimplifiedTests/NYPLContentDownloaderTests.swift
singhraman4282: we're testing to make sure CustomNYPLContentDownloader gets deinitialized properly without any retain cycles or memory leaks by first creating an optional object and later nullifying it. Without the explicit `CustomNYPLContentDownloader?` we can't make it nil to test it.

SimplifiedTests/NYPLAxisLicenseServiceTests.swift
singhraman4282: Done.

SimplifiedTests/NYPLAxisLicenseServiceTests.swift
singhraman4282: That makes more sense. Done.

singhraman4282:
@ettore All axis related tests moved to separate directory. Will do the same in other PRs

ErnestFan:
> @ettore All axis related tests moved to separate directory. Will do the same in other PRs

@singhraman4282 The number of Axis-related mock classes are getting larger from all these PRs combined, we should group them in a sub-directory `Mocks` of the AxisUnitTest directory. What do you think?

P.S. This can be done after all the PRs are merged.

SimplifiedTests/NYPLAxisTests/license-9781250109699.json
```diff
@@ -0,0 +1 @@
...":{"copy":{"copyright":true,"days":365,"notice":"Copyright held by Publisher.  All rights Reserved.","percentage":50,"permission":false,"watermark":true},"devices":6,"preview":null,"print":{"copyright":false,"days":5,"notice":"","percentage":50,"permission":false,"watermark":false},"reflow":true,"ttsread":true},"signature":{"algorithm":"http://www.w3.org/2000/09/xmldsig#rsa-sha1","certificate":"BgIAAACkAABSU0ExAAgAAAEAAQAlOqGroR57awwqgtZgPCIrT2+fwq3bxzyF9HSDghHxaqECT2Aaf60c0usWGAJRbi00fS5OYoCox753bc6w4Ii8F7OooNuqTFfILUk1qDcyDoMKRVjuPf1oDcwmygt/CSnbngUDCtrLmTGUxzjac8KSIZ3mGBVBPr+FV9hBLQhWcv4srhlHpOvwQQGR4aftRyi2hUhJWqECqVGqxY5cp94pRnlYiOy0GCFCYjkQI9T+wEbBKHSGQsabYTJO2u51wDhgYZEL+t0ec6F/F6i/zfGI+hBxFSnZfgj/NhcNvjDOho2YCIGGfH0Cv4xz3cCcj8EQOES9n9PwjVKq9/Ai8ifM","value":"I7Ydnew5Yc5KmCyxoky5qXnTVqGGt54dxsZ4g00dlbGLSxrgV7tuTLskRn/3n8ywRDZZDt9VnAaMQDDRyQOlVONm/G9yQcHr0PxL7U+lGDhyHUIYKa5VA27uUEe2UB37+YZmU8I68nppT+hgq1aqRpcDZtud6JOIQK2CyMjxxjA71f7mMAHs+IxwM8NCKISIS/IRpxXOwc9U5V/5ygZgo1xEqoVlw/CXnXy7D1RBser2NUqRWYVnlDkIAfE6mtE4ml2aO9eEdV0uGpLLU+5XZTlq6YPcmFv4MxPz6BkOmMHzEpoGC1g5/3/oPlBmd6AyKobXM3D1ZBQ3f59rLAudbQ==","LicenseConicalForm":null,"RawHashLicenseConicalForm":null,"RawAESkey":null}}
```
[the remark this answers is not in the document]
singhraman4282: yes these are. To download and decrypt a book, one needs the bookId and bookVault id. The license contains bookVaultID but it's encrypted with the key we sent before downloading the license. Since there is no bookId and the bookVaultID in the license is encrypted with our private key, nobody can have unauthorized access to the books.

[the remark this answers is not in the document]
ettore: That's good, however it's still a hacking vector for a malicious person: this might be ok for us or it might not. I've definitely be in situations where this would not be ok, so I should ask if this was cleared by Risa aka @killerpuppytails and how comfortable we are with this. There might be legal requirements I'm not aware of.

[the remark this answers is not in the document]
singhraman4282: Modified license file. None of the values in the license are real anymore. The base64 string is no longer pointing to the key.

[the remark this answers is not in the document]
killerpuppytails: So this flow wasn't cleared by me, no, but this is a known B&T process. Let's bring @leonardr into this conversation, because I believe there's a way to make this more secure.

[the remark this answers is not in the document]
leonardr: In a situation like this, masking the real values is the right call. Since the old values are now under version control, it would also be good to rebase this branch so they don't get merged into the history. Based on my knowledge of AxisNow, I agree with Raman that these values don't themselves compromise security, but it's best to be conservative.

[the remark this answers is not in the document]
singhraman4282: @ettore @leonardr @killerpuppytails the branch has been rebased with masked values.

ettore (approved):

PR #1386, opened 2021-05-20
NYPL Axis Service Tests

singhraman4282:
**What's this do?**
Adds tests for NYPLAxisService

N/A

Cannot be QA tested

N/A

N/A

N/A

Raman Singh verified it

### DIFF
--- a/SimplifiedTests/NYPLAxisTests/NYPLAxisServiceTests.swift
+++ b/SimplifiedTests/NYPLAxisTests/NYPLAxisServiceTests.swift
@@ -1,0 +1,168 @@
+//
+//  NYPLAxisServiceTests.swift
+//  OpenEbooksTests
+//
+//  Created by Raman Singh on 2021-05-14.
+//  Copyright © 2021 NYPL Labs. All rights reserved.
+//
+
+import XCTest
+@testable import SimplyE
+
+class NYPLAxisServiceTests: XCTestCase {
+  
+  lazy private var downloadsDirectory: URL = {
+    let paths = FileManager.default.urls(for: .documentDirectory, in: .userDomainMask)
+    let documentsDirectory = paths[0]
+    return documentsDirectory.appendingPathComponent("NYPLAxisLicenseServiceTests")
+  }()
+  
+  lazy private var book: NYPLBook = {
+    let acquisitions = [NYPLFake.genericAcquisition.dictionaryRepresentation()]
+    
+    return NYPLBook(dictionary: ["acquisitions": acquisitions,
+                                 "categories" : ["Fantasy"],
+                                 "id": "666",
+                                 "title": "The Lord of the Rings",
+                                 "updated": "2020-09-08T09:22:45Z"
+    ])!
+  }()
+  
+  private let itemDownloader = NYPLAxisItemDownloader()
+  private let downloadBroadcaster = NYPLAxisBookDownloadBroadcasterMock()
+  
+  func testAxisServiceShouldFailUponLicenseDownloadFailure() {
+    let downloadFailedExpectation = self.expectation(
+      description: "Failed license process should result in failed download!")
+      
+    let axisService = createAxisService(licenseService: failingLicenseService,
+                                        metadataService: succeedingMetadataService,
+                                        packageService: succeedingPackageService)
+    
+    let task = URLSessionDownloadTask()
+    
+    downloadBroadcaster.downloadFailed = {
+      downloadFailedExpectation.fulfill()
+    }
+    
+    downloadBroadcaster.downloadSuccessful = {
+      XCTFail()
+    }
+    
+    axisService.fulfillAxisLicense(downloadTask: task)
+    
+    waitForExpectations(timeout: 4, handler: nil)
+  }
+  
+  func testAxisServiceShouldFailUponMetadataContentDownloadFailure() {
+    let downloadFailedExpectation = self.expectation(
+      description: "Failed metadata download should result in failed download!")
+      
+    let axisService = createAxisService(licenseService: succeedingLicenseService,
+                                        metadataService: failingMetadataService,
+                                        packageService: succeedingPackageService)
+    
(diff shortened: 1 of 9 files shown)


SimplifiedTests/MockServiceUsingAxisItemDownloader.swift
```diff
@@ -0,0 +1,31 @@
...
+//  OpenEbooksTests
+//
+//  Created by Raman Singh on 2021-05-17.
+//  Copyright © 2021 NYPL Labs. All rights reserved.
+//
+
+import Foundation
+@testable import SimplyE
+
+class MockServiceUsingAxisItemDownloader {
```
ErnestFan, line 12: I think the name of this class is kinda confusing.
Can it be named something like `NYPLAxisItemDownloaderMockResponseService / NYPLAxisItemDownloaderMockSucceedService`? Or anything describe this class better.
@ettore any thoughts?

ettore: I'm not sure... I don't think this is mocking a service (it's not implementing any protocols), it just provides functionality to other mocks as part of their own class hierarchy. Mock ideally should be as simple as possible to avoid getting into situations where the flow of the test is hard to follow.

Maybe I would call it NYPLDownloadRunnerMock since it's triggering the downloader? but I'm actually not sure how this works to make a good suggestion.

singhraman4282: NYPLDownloadRunnerMock makes more sense. This class acts as a base class for NYPLAxisMetadataServiceMock,  NYPLAxisPackageServiceMock, & NYPLAxisLicenseServiceMock. If you believe this complicates the code I'll be more than happy to make changes to it.

ErnestFan (approved):

SimplifiedTests/MockServiceUsingAxisItemDownloader.swift
ettore: generally speaking in tests is totally fine to repeat the same code over and over. The objective is to write tests that are easy to understand -- we don't want to look for logic bugs in tests, essentially. I do get it that like in this case there is some test logic that is repeated.

But no I'm not asking you to change it because I don't have better suggestions right now (I would have to look at it more deeply and I can't right now).

For the future though I would limit code reuse to things like helper functions but not much else. I'd try to avoid any logic like conditional branches, loops, subclassing-- not always possible but something to strive for.

SimplifiedTests/NYPLAxisTests/NYPLAxisServiceTests.swift
```diff
@@ -0,0 +1,182 @@
...
+                                        packageService: succeedingPackageService)
+    
+    let task = URLSessionDownloadTask()
+    
+    downloadBroadcaster.downloadFailed = {
+      downloadFailedExpectation.fulfill()
+    }
+    
+    downloadBroadcaster.downloadSuccessful = {
+      downloadSucceededExpectation.fulfill()
```
ettore, line 53: wouldn't this mean that we have to wait 10 seconds for this inverted expectation to be... not fulfilled? I haven't used inverted expectations often. 

wouldn't it be simpler to just do a XCFail() inside the success block instead? Same for the other tests, wherever we have explicit fail / success blocks.

SimplifiedTests/NYPLAxisTests/NYPLAxisServiceTests.swift
```diff
@@ -0,0 +1,182 @@
...
+    }
+    
+    axisService.fulfillAxisLicense(downloadTask: task)
+    
+    wait(
+      for: [downloadFailedExpectation, downloadSucceededExpectation],
+      timeout: 10.0)
```
ettore, line 60: just wondering if these can be reduced to 2-3 secs

SimplifiedTests/NYPLAxisTests/NYPLAxisServiceTests.swift
```diff
@@ -0,0 +1,182 @@
...
+    let axisService = createAxisService(licenseService: succeedingLicenseService,
+                                        metadataService: succeedingMetadataService,
+                                        packageService: succeedingPackageService)
+    
+    let task = URLSessionDownloadTask()
+    
+    downloadBroadcaster.downloadSuccessful = {
+      downloadSucceededExpectation.fulfill()
+    }
+
```
ettore, line 134: can we have a  `downloadFailed = { XCTFail() }` so we can fail faster?

SimplifiedTests/NYPLAxisTests/NYPLAxisServiceTests.swift
singhraman4282: Agreed. Fixed.

SimplifiedTests/NYPLAxisTests/NYPLAxisServiceTests.swift
singhraman4282: Done

SimplifiedTests/NYPLAxisTests/NYPLAxisServiceTests.swift
singhraman4282: Dine

ettore: I don't see the downloadFailed callback anymore, maybe it got lost while rebasing?

ettore (approved):

SimplifiedTests/NYPLAxisTests/NYPLAxisServiceTests.swift
singhraman4282: rebases 😬

ettore: I see it now, looks good 👍

PR #1389, opened 2021-05-24
IOS-205 Fix inconsistencies returning books

ettore:
**What's this do?**
- make sure to save the book registry after returning the book successfully and when there was no active loan
- make sure to present an error if returning loan to ACS fails
- first return loan to ACS; if that succeeds revoke loan with Circulation Manager

@leonardr do you see any issues with this synchronization? I just want to make sure the CM doesn't expect an active loan still open on the ACS side when the call to revoke the loan is received. If so I can flip the sequence of the 2 calls.

https://jira.nypl.org/browse/IOS-205

see ticket

none, this will merge into develop and will appear in the SimplyE release after 3.6.5 

yes

yes

@ettore


Simplified/MyBooks/NYPLMyBooksDownloadCenter.m
```diff
@@ -607,66 +606,116 @@ - (void)returnBookWithIdentifier:(NSString *)identifier
...
 #endif
+}
 
+- (void)revokeLoanAndRemoveBook:(NYPLBook *)book state:(NYPLBookState)state
+{
+  BOOL didDownloadBook = (state == NYPLBookStateDownloadSuccessful
+                          || state == NYPLBookStateUsed);
+  NSString *const identifier = book.identifier;
+
+  // maybe this is the case when book has no DRM so there's no loan to revoke?
```
leonardr, line 635: The main time you won't see a revoke link is when the library doesn't authenticate its users. In that case, yes, there is no server-side loan to revoke.

Simplified/MyBooks/NYPLMyBooksDownloadCenter.m
```diff
@@ -607,66 +606,116 @@ - (void)returnBookWithIdentifier:(NSString *)identifier
...
 
-      [[NYPLBookRegistry sharedRegistry] setProcessing:NO forIdentifier:book.identifier];
-      
-      if(feed && feed.entries.count == 1)  {
-        NYPLOPDSEntry *const entry = feed.entries[0];
-        if(downloaded) {
-          [self deleteLocalContentForBookIdentifier:identifier];
-        }
-        NYPLBook *returnedBook = [NYPLBook bookWithEntry:entry];
-        if(returnedBook) {
```
ettore, line 636: I removed this `if` because it's irrelevant. It's impossible to get a nil book from `bookWithEntry` if `entry` is not nil, and we know that from `NYPLOPDSEntry *const entry = feed.entries[0];` (NSArrays cannot contain nil values).

leonardr:
I was not able to follow the code 100% but based on your description what you have is right. For loans through ACS, the "revoke" link doesn't revoke anything: the client uses it to _notify_ the server that it returned the book. For that reason it should be hit after the ACS revoke succeeds; otherwise the book will immediately show up again in the loans feed.

ettore:
> I was not able to follow the code 100% 

sorry about that-- this code is old and in need of a refactor. But it's a huge refactor that will require at least a sprint.

> but based on your description what you have is right. For loans through ACS, the "revoke" link doesn't revoke anything: the client uses it to _notify_ the server that it returned the book. For that reason it should be hit after the ACS revoke succeeds; otherwise the book will immediately show up again in the loans feed.

Sounds good then, this maybe will solve some of the issues we've had in the past with book loans showing up again in My Books for mysterious reasons.

leonardr:
I would really like to know the answer to this question:

* Return a book through ACS (and hit the revoke link)
* Fetch the loans feed
* Is the book you just returned in the loans feed or not?

If so, there's a server-side problem we need to fix. I can't test this on my own because I can't do an ACS return.

ettore:
> I would really like to know the answer to this question:
> 
> * Return a book through ACS (and hit the revoke link)
> * Fetch the loans feed
> * Is the book you just returned in the loans feed or not?

It's not. Here's what I did in detail though.

I signed in and I already had a book in My Books. I returned that using the new code. I verified GET https://circulation.librarysimplified.org/NYNYPL/loans/441963/revoke  is hit and returns a 200 (not sure why it's a GET, in my mind that should be a POST but that's how the code was written). Then I pulled to refresh to refetch My Books (`/NYNYPL/loans`) and the feed was empty.

leonardr:
That's good news -- it means hitting the revoke endpoint invalidates the server-side loans cache, forcing the server to check with the vendor to see whether the book was actually returned.

Simplified/Book/Models/NYPLBookRegistry.m
```diff
@@ -461,7 +461,7 @@ - (void)updateBook:(NYPLBook *const)book
 - (void)updateAndRemoveBook:(NYPLBook *)book
 {
   if(!book) {
-    @throw NSInvalidArgumentException;
+    return;
```
singhraman4282, line 464: this was much needed

singhraman4282 (approved):

ErnestFan (approved):

ettore:
had to rebase to fix conflicts.

PR #1390, opened 2021-05-24
IOS-199 Error out if Adobe fulfillment fails

ettore:
**What's this do?**
If Adobe DRM fails fulfilling a book loan request for whatever reason, SimplyE silently swallows the error and proceeds as nothing happened. This changes the behavior to fail immediately and report an error to the user.

https://jira.nypl.org/browse/IOS-199

We should verify fulfillment works as before in the and that there are not false positives reported by Adobe. A regression should cover this.

Requires https://github.com/NYPL-Simplified/DRM-iOS-Adobe/pull/20 to be merged first, then I have to update the commit hash for this PR to match that.

this could go together with PR https://github.com/NYPL-Simplified/Simplified-iOS/pull/1389

yes

@ettore

### DIFF
--- a/Simplified/MyBooks/NYPLMyBooksDownloadCenter.m
+++ b/Simplified/MyBooks/NYPLMyBooksDownloadCenter.m
@@ -267,7 +267,16 @@ didFinishDownloadingToURL:(NSURL *const)tmpSavedFileURL
            fulfillWithACSMData:ACSMData
            tag:book.identifier
            userID:[[NYPLUserAccount sharedAccount] userID]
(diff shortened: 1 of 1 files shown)


ettore:
Adding @vladimirfedorov in case he can see any reasons for not doing this since he's more familiar with Adobe DRM.

singhraman4282 (approved):

ErnestFan (approved):

vladimirfedorov (approved):
That's good, no reason to mute any error message here

ettore:
updated commit for adept-ios

PR #1396, opened 2021-05-25
iOS-37 Call new endpoints for LibraryRegistry

ErnestFan:
**What's this do?**
Dependency inject `AccountsManager` into `NYPLFinderBusinessLogic` in order to load registry feed
Construct request URL with new endpoints, user's location and search keyword
Some minor UI fixes and clean up

[iOS-37](https://jira.nypl.org/browse/IOS-37)

Not ready to be tested

N/A

N/A

N/A

@ErnestFan

### DIFF
--- a/Simplified/AppInfrastructure/NYPLConfiguration+OE.swift
+++ b/Simplified/AppInfrastructure/NYPLConfiguration+OE.swift
@@ -20,7 +20,7 @@ extension NYPLConfiguration {
                      ofType: "json")!)
   private static let prodFeedKey = "OpenEBooksLibraryRegistryProdFeedKey"
   
-  static var prodUrl = feedFileUrl
+  static let prodUrl = feedFileUrl
   private static let prodFeedKeyHash = prodFeedKey.md5().base64EncodedStringUrlSafe().trimmingCharacters(in: ["="])
 
   // MARK:- Beta library catalog
@@ -32,7 +32,7 @@ extension NYPLConfiguration {
                      ofType: "json")!)
   private static let betaFeedKey = "OpenEBooksLibraryRegistryBetaFeedKey"
 
-  static var betaUrl = betaFeedFileUrl
+  static let betaUrl = betaFeedFileUrl
   private static let betaFeedKeyHash = betaFeedKey.md5().base64EncodedStringUrlSafe().trimmingCharacters(in: ["="])
   
   static let feedKeyHash = {
--- a/Simplified/AppInfrastructure/NYPLConfiguration+OE.swift
+++ b/Simplified/AppInfrastructure/NYPLConfiguration+OE.swift
@@ -21,7 +21,7 @@ extension NYPLConfiguration {
   private static let prodFeedKey = "OpenEBooksLibraryRegistryProdFeedKey"
   
   static var prodUrl = feedFileUrl
-  static let prodFeedKeyHash = prodFeedKey.md5().base64EncodedStringUrlSafe().trimmingCharacters(in: ["="])
+  private static let prodFeedKeyHash = prodFeedKey.md5().base64EncodedStringUrlSafe().trimmingCharacters(in: ["="])
 
   // MARK:- Beta library catalog
 
@@ -32,7 +32,7 @@ extension NYPLConfiguration {
                      ofType: "json")!)
   private static let betaFeedKey = "OpenEBooksLibraryRegistryBetaFeedKey"
 
-  private static var betaUrl = betaFeedFileUrl
+  static var betaUrl = betaFeedFileUrl
   private static let betaFeedKeyHash = betaFeedKey.md5().base64EncodedStringUrlSafe().trimmingCharacters(in: ["="])
   
   static let feedKeyHash = {
--- a/Simplified/AppInfrastructure/NYPLConfiguration+OE.swift
+++ b/Simplified/AppInfrastructure/NYPLConfiguration+OE.swift
@@ -18,9 +18,10 @@ extension NYPLConfiguration {
   private static let feedFileUrl = URL(fileURLWithPath:
     Bundle.main.path(forResource: "OpenEBooks_OPDS2_Library_Registry_Feed",
                      ofType: "json")!)
-  private static let feedFileUrlHash = feedFileUrl.absoluteString.md5().base64EncodedStringUrlSafe().trimmingCharacters(in: ["="])
+  private static let prodFeedKey = "OpenEBooksLibraryRegistryProdFeedKey"
+  
   static var prodUrl = feedFileUrl
-  static var prodUrlHash = feedFileUrlHash
+  static let prodFeedKeyHash = prodFeedKey.md5().base64EncodedStringUrlSafe().trimmingCharacters(in: ["="])
 
   // MARK:- Beta library catalog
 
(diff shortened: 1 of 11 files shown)


singhraman4282 (approved):

Simplified/Settings/NYPLSettingsAccountsList.swift
```diff
@@ -254,6 +202,7 @@
     var array = userAddedSecondaryAccounts!
```
ettore, line 202: maybe I missed this in a previous review, but is this forced unwrap safe here?

ettore (commented):
looks good overall, mostly minor comments/questions

oh, one non minor thought was about unit tests: it seems like some of this is unit testable, particularly the business logic?

Simplified/AppInfrastructure/NYPLConfiguration+OE.swift
```diff
@@ -18,9 +18,10 @@ extension NYPLConfiguration {
   private static let feedFileUrl = URL(fileURLWithPath:
     Bundle.main.path(forResource: "OpenEBooks_OPDS2_Library_Registry_Feed",
                      ofType: "json")!)
-  private static let feedFileUrlHash = feedFileUrl.absoluteString.md5().base64EncodedStringUrlSafe().trimmingCharacters(in: ["="])
+  private static let prodFeedKey = "OpenEBooksLibraryRegistryProdFeedKey"
+  
   static var prodUrl = feedFileUrl
-  static var prodUrlHash = feedFileUrlHash
+  static let prodFeedKeyHash = prodFeedKey.md5().base64EncodedStringUrlSafe().trimmingCharacters(in: ["="])
```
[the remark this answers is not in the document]
ErnestFan: Yes, just the key we use for storing

Simplified/Settings/NYPLSettingsAccountsList.swift
ErnestFan: Not sure why it's implemented that way, I assign an empty array to `userAddedSecondaryAccounts` in `init` so it's no longer optional

ErnestFan:
> looks good overall, mostly minor comments/questions
> 
> oh, one non minor thought was about unit tests: it seems like some of this is unit testable, particularly the business logic?

I am working on the unit test now, will let you know once it's ready

singhraman4282:
Tests look good

Simplified/LibraryFinder/UI/NYPLLibraryFinderViewController.swift
```diff
@@ -48,7 +49,6 @@ class NYPLLibraryFinderViewController: UICollectionViewController, UICollectionV
   }
 
   override func collectionView(_ collectionView: UICollectionView, didSelectItemAt indexPath: IndexPath) {
-    super.collectionView(collectionView, didSelectItemAt: indexPath)
```
ettore, line 51: shouldn’t this stay? just wondering if there’s a reason i’m missing … even if super class impl. does nothing, it’s good to future proof it by calling super….

SimplifiedTests/NYPLFinderBusinessLogicTests.swift
```diff
@@ -0,0 +1,82 @@
...
+    libraryRegistryMock.requestUrl = nil
+    libraryRegistryMock.libraryAccounts = [Account]()
+    
+    businessLogic.userLocation = nil
+    businessLogic.newLibraryAccounts = [Account]()
+  }
+
+  func testLibraryRegistryRequestUrlWithSearchKeyword() throws {
+    businessLogic.requestLibraryList(searchKeyword: "test") { (success) in
+      XCTAssertEqual(self.libraryRegistryMock.requestUrl?.absoluteString, "http://librarysimplified.org/terms/rel/search?query=test&stage=all")
```
ettore, line 47: we should add an expectation that the closure is called. Otherwise the test might still pass if we don’t call the completion.

Simplified/LibraryFinder/UI/NYPLLibraryFinderViewController.swift
ErnestFan: This line is causing a `NSInvalidArgumentException` crash for some reason, even if this is only line being called in this method. Any thoughts?

ettore: that's pretty weird, I'm guessing the index path somehow became invalid? even that doesn't make a lot of sense, but the exception hints at something to do with the arguments. is it at all possible that the collection view was dealloced and recreated? or some other thread is modifying the number of items to be displayed?

singhraman4282 (approved):

Simplified/LibraryFinder/UI/NYPLLibraryFinderViewController.swift
ErnestFan: I checked everything you mentioned above and none of them seems to be the case.
Could it be because `collectionView(_:didSelectItemAt:))` is an optional implementation for `UICollectionViewController`?
ref: https://stackoverflow.com/a/11403413

ErnestFan: I can add an if-statement `UICollectionViewController.responds(to: #selector(collectionView(_:didSelectItemAt:)))` before calling the `super.collectionView(_:didSelectItemAt:))`. What do you think?

ettore: that's an interesting thought.... it's weird that the method though requires `override`... but if you're saying that with the check it doesn't crash that's probably it. I think though you wanna use `instancesRespond(to:)`?

Simplified/AppInfrastructure/NYPLConfiguration+OE.swift
```diff
@@ -29,10 +30,14 @@ extension NYPLConfiguration {
   private static let betaFeedFileUrl = URL(fileURLWithPath:
     Bundle.main.path(forResource: "OpenEBooks_OPDS2_Library_Registry_Feed-QA",
                      ofType: "json")!)
-  private static let betaFeedFileUrlHash = feedFileUrl.absoluteString.md5().base64EncodedStringUrlSafe().trimmingCharacters(in: ["="])
+  private static let betaFeedKey = "OpenEBooksLibraryRegistryBetaFeedKey"
```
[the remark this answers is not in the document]
ettore: good point, this is true for other classes too. Make sure to periodically merge `develop` into your feature branches, you will probably  still have conflicts but they will be manageable.

ettore (approved):

Simplified/AppInfrastructure/NYPLConfiguration+OE.swift
```diff
@@ -32,8 +32,12 @@ extension NYPLConfiguration {
                      ofType: "json")!)
   private static let betaFeedKey = "OpenEBooksLibraryRegistryBetaFeedKey"
 
-  static var betaUrl = betaFeedFileUrl
-  static let betaFeedKeyHash = betaFeedKey.md5().base64EncodedStringUrlSafe().trimmingCharacters(in: ["="])
+  private static var betaUrl = betaFeedFileUrl
```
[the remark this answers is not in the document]
ErnestFan: I accidentally changed this to `private` and I changed it back, this is being called in `AccountsManager`

Simplified/LibraryFinder/UI/NYPLLibraryFinderViewController.swift
ErnestFan: You are right, updated it with `instancesRespond(to:)`

PR #1401, opened 2021-05-31
IOS-5 Remove R1 related code

ettore:
**What's this do?**
This is removing R1-related code, in particular:
- deleted all objc and JS code that was responsible for presenting books using R1
- removed dependency from the readium-shared-js and adobe-content-filters submodules
- skipped compilation of a few classes in the RDServices project from the readium-sdk (mostly classes related to parsing multiform part data responses and log formatters from CocoaLumberjack). The classes I did keep are still referenced by the Adobe SDK binaries from the ADEPT submodule (I was getting linker errors).
- Removed the R1 - R2 toggle
- other minor cleanup from unused methods.

https://jira.nypl.org/browse/IOS-5

This will be covered in the final regression. Opening a book in R2 should be unaffected. 

Merging to feature branch that contains other R1-breaking changes.

yes

yes

@ettore


ettore:
@vladimirfedorov I'd appreciate your input on the removal of adobe-content-filter. Feel free to just look at commit https://github.com/NYPL-Simplified/Simplified-iOS/commit/1c8e42cc22e7d2c5b27aa3625b62a0c6e607efb6. The [decrypt call you added in AdobeDRMContainer](https://github.com/NYPL-Simplified/Simplified-iOS/blob/560a4de55e5c3ac9e384d5f6766acd34cfcd34ad/Simplified/Reader2/ReaderStackConfiguration/AdobeDRM/AdobeDRMContainer.mm#L103) seem to match [what is in Adobe-Content-Filter](https://github.com/NYPL-Simplified/Adobe-Content-Filter/blob/c4e50bef31ff45e8e9e692fdf6277f54d52161e6/src/adept_filter.cpp#L99), so hopefully we are ok. But do you think the adobe-sdk "secretly" expects the content-filter code to be there somehow? It seems to work fine without it when using R2, but I don't know if I'm missing any edge-cases.

ErnestFan (approved):

ettore:
> I'm not 100% sure whether the adobe-sdk "secretly" expects the content-filter code to be there somehow, but it does seem to work fine without it when using R2 -- I just don't know if I'm missing any edge-cases.

I reached out to one of the developers that worked on AdobeDRM-related things for R2 and SimplyE and it's confirmed that R2 + Adobe connector does not need adobe-content-filter code. So it's safe to remove.

singhraman4282 (approved):

PR #1404, opened 2021-06-03
Wrote tests for DownloadTaskProgress

singhraman4282:
**What's this do?**
Adds tests for NYPLAxisDownloadProgress

Testing purposes

Cannot be QA tested

N/A

N/A

N/A

Raman Singh verified it

### DIFF
--- a/SimplifiedTests/NYPLAxisTests/NYPLAxisDownloadProgressTests.swift
+++ b/SimplifiedTests/NYPLAxisTests/NYPLAxisDownloadProgressTests.swift
@@ -1,0 +1,326 @@
+//
+//  NYPLAxisDownloadProgressTests.swift
+//  OpenEbooksTests
+//
+//  Created by Raman Singh on 2021-06-01.
+//  Copyright © 2021 NYPL Labs. All rights reserved.
+//
+
+import XCTest
+@testable import SimplyE
+
+class NYPLAxisDownloadProgressTests: XCTestCase {
+
+  private let keysProvider = NYPLAxisKeysProvider()
+  private let licenseURL = URL(string: "www.mock.com/license")!
+  private let containerURL = URL(string: "www.mock.com/container")!
+  private let encryptionURL = URL(string: "www.mock.com/encryption")!
+  private let packageURL = URL(string: "www.mock.com/package")!
+  private let chapter1URL = URL(string: "www.mock.com/chapter1")!
+  private let chapter2URL = URL(string: "www.mock.com/chapter2")!
+  private let chapter3URL = URL(string: "www.mock.com/chapter3")!
+
+  private let progressListener = NYPLAxisProgressListenerMock()
+
+  lazy private var sut: NYPLAxisDownloadProgress = {
+    let axisDownloadProcess = NYPLAxisDownloadProgress()
+    axisDownloadProcess.progressListener = self.progressListener
+    return axisDownloadProcess
+  }()
+
+  override func tearDown() {
+    super.tearDown()
+    sut.test_reset()
+  }
+
+  func testDownloadingFixedWeightItemsShouldUpdateDownloadProgress() {
+    sut.addFixedWeightTask(with: licenseURL, weight: 0.05)
+    sut.addFixedWeightTask(with: containerURL, weight: 0.05)
+
+    XCTAssertEqual(progressListener.currentProgress, 0.0)
+
+    sut.didFinishTask(with: licenseURL)
+    // 5 % downloaded
+    XCTAssertEqual(progressListener.currentProgress, 0.05)
+
+    sut.addFixedWeightTask(with: encryptionURL, weight: 0.4)
+    // 5 % downloaded
+    XCTAssertEqual(progressListener.currentProgress, 0.05)
+
+    sut.didFinishTask(with: encryptionURL)
+    // 45 % downloaded
+    XCTAssertEqual(progressListener.currentProgress, 0.45)
+  }
+
+  func testDownloadingFlexibleWeightItemShouldUpdateDownloadProgress() {
+    sut.addFixedWeightTask(with: licenseURL, weight: 0.05)
+    sut.addFixedWeightTask(with: containerURL, weight: 0.05)
+    sut.addFlexibleWeightTask(with: chapter1URL)
+    sut.addFlexibleWeightTask(with: chapter2URL)
+
+    XCTAssertEqual(progressListener.currentProgress, 0.0)
+
+    sut.didFinishTask(with: chapter1URL)
+    // 45 % downloaded
+    XCTAssertEqual(progressListener.currentProgress, 0.45)
+
+    sut.didFinishTask(with: chapter2URL)
+    // 90% downloaded
+    XCTAssertEqual(progressListener.currentProgress, 0.9)
+  }
+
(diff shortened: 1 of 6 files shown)


SimplifiedTests/NYPLAxisTests/NYPLAxisDownloadProgressTests.swift
```diff
@@ -0,0 +1,312 @@
...
+    XCTAssertEqual(sut.test_getWeightForProcess(withURL: licenseURL), 0.05)
+    XCTAssertEqual(sut.test_getWeightForProcess(withURL: containerURL), 0.05)
+    XCTAssertEqual(sut.test_getWeightForProcess(withURL: encryptionURL), 0.05)
+    XCTAssertEqual(sut.test_getWeightForProcess(withURL: chapter1URL), 0.85/3)
+    XCTAssertEqual(sut.test_getWeightForProcess(withURL: chapter2URL), 0.85/3)
+    XCTAssertEqual(sut.test_getWeightForProcess(withURL: chapter3URL), 0.85/3)
+
+    // ~43% completed
+    sut.didFinishTask(with: chapter2URL)
+    XCTAssertEqual(progressListener.currentProgress, (0.15 + (0.85/3)).roundedToTwoDecimalPlaces)
```
ettore, line 166: great test -- but if you round it why not just using `0.43`?

SimplifiedTests/NYPLAxisTests/NYPLAxisItemDownloaderTests.swift
```diff
@@ -198,4 +203,71 @@ class NYPLAxisItemDownloaderTests: XCTestCase {
...
+    
+    let itemDownloader = NYPLAxisItemDownloader(
+      assetWriter: assetWriter, dispatchGroup: dispatchGroup,
+      downloader: contentDownloader, weightProvider: weightProvider)
+    
+    let progressListener = NYPLAxisProgressListenerMock()
+    itemDownloader.delegate = progressListener
+    XCTAssertEqual(progressListener.currentProgress, 0)
+    
+    dispatchGroup.enter()
```
ettore, line 218: would you say this usage of the dispatchGroup assumes knowledge of how the group is used inside the  code being tested?  Or is this documented? it probably is/should, but I don't recall. From this point of view I wouldn't know why I need to "enter" the group here and wait later, for example. If it's already documented in NYPLAxisItemDownloader, this is fine.

ettore (approved):

SimplifiedTests/NYPLAxisTests/NYPLAxisDownloadProgressTests.swift
```diff
@@ -0,0 +1,312 @@
...
+    XCTAssertEqual(sut.test_getWeightForProcess(withURL: chapter1URL), 0.45)
+    XCTAssertEqual(sut.test_getWeightForProcess(withURL: chapter2URL), 0.45)
+
+    sut.addFlexibleWeightTask(with: encryptionURL)
+    // 45 % downloaded
+    XCTAssertEqual(progressListener.currentProgress, 0.45)
+    XCTAssertEqual(sut.test_getWeightForProcess(withURL: licenseURL), 0.05)
+    XCTAssertEqual(sut.test_getWeightForProcess(withURL: containerURL), 0.05)
+    XCTAssertEqual(sut.test_getWeightForProcess(withURL: chapter1URL), 0.45)
+    XCTAssertEqual(sut.test_getWeightForProcess(withURL: chapter2URL), 0.45)
```
ErnestFan, line 98: Correct me if I am wrong, we now have two unfinished flexible weight tasks, should the weight of `chapter2URL` be 0.225 ((1 - 0.05 - 0.05 - 0.45) / 2)?

SimplifiedTests/NYPLAxisTests/NYPLAxisDownloadProgressTests.swift
```diff
@@ -0,0 +1,312 @@
...
+    XCTAssertEqual(sut.test_getWeightForProcess(withURL: chapter1URL), 0.8)
+
+    sut.addFlexibleWeightTask(with: chapter3URL)
+    XCTAssertEqual(progressListener.currentProgress, 0.0)
+    XCTAssertEqual(sut.test_getWeightForProcess(withURL: licenseURL), 0.2)
+    XCTAssertEqual(sut.test_getWeightForProcess(withURL: chapter1URL), 0.4)
+    XCTAssertEqual(sut.test_getWeightForProcess(withURL: chapter3URL), 0.4)
+
+    // 20 % completed
+    sut.didFinishTask(with: licenseURL)
```
ErnestFan, line 190: This test is named `AddingMoreTasksShouldNotAffectWeightOfCompletedTasks` but there is no task added after this task is completed. Also `licenseURL` is a fixed weight task, so it should not be affected in any ways?

SimplifiedTests/NYPLAxisTests/NYPLAxisItemDownloaderTests.swift
singhraman4282: You're right. We're entering dispatchGroup from multiple places. However, NYPLAxisItemDownloader should be the only place where we enter and leave the dispatchGroup and other classes should not be aware of it. Added a TODO note in code. Will work on that.

SimplifiedTests/NYPLAxisTests/NYPLAxisDownloadProgressTests.swift
singhraman4282: You sir are absolutely right. Not only was there a problem with my math, but there was also a bug 🤮  in my logic. Thank you for pointing it out. Fixed.

SimplifiedTests/NYPLAxisTests/NYPLAxisDownloadProgressTests.swift
singhraman4282: Great catch. Fixed.

SimplifiedTests/NYPLAxisTests/NYPLAxisDownloadProgressTests.swift
singhraman4282: You're right. It is 43%.

SimplifiedTests/NYPLAxisTests/NYPLAxisDownloadProgressTests.swift
```diff
@@ -0,0 +1,314 @@
...
+  func testAddingMoreTasksShouldNotAffectWeightOfCompletedTasks() {
+    sut.addFixedWeightTask(with: licenseURL, weight: 0.2)
+    sut.addFlexibleWeightTask(with: chapter1URL)
+
+    XCTAssertEqual(progressListener.currentProgress, 0.0)
+    XCTAssertEqual(sut.test_getWeightForProcess(withURL: licenseURL), 0.2)
+    XCTAssertEqual(sut.test_getWeightForProcess(withURL: chapter1URL), 0.8)
+    
+    // 20 % completed
+    sut.didFinishTask(with: licenseURL)
```
ErnestFan, line 185: We are checking the weight of a fixed weight task here, shouldn't we test a flexible weight task instead?

singhraman4282: done

ErnestFan (approved):

PR #1409, opened 2021-06-15
OE-142: Fixed incorrect banner and button timer for some axis books

singhraman4282:
**What's this do?**
Fixes incorrect expiration date on some axis books

https://jira.nypl.org/browse/OE-142

Follow the steps in the ticket

N/A

Not yet

N/A

Raman Singh verified it

### DIFF
--- a/Simplified/Book/UI/NYPLBookButtonsView.m
+++ b/Simplified/Book/UI/NYPLBookButtonsView.m
@@ -303,9 +303,21 @@
       [self.book.defaultAcquisition.availability
        matchUnavailable:nil
        limited:^(NYPLOPDSAcquisitionAvailabilityLimited *const _Nonnull limited) {
+        
+        NSDate *untilDate = self.book.defaultAcquisition.availability.until;
+        NSDate *sinceDate = self.book.defaultAcquisition.availability.since;
+        
+        BOOL hasIncorrectDates = NO;
+        
+        if (untilDate && sinceDate) {
+          hasIncorrectDates = (untilDate.timeIntervalSinceReferenceDate - sinceDate.timeIntervalSinceReferenceDate == 3600);
+        }
+        
         if ([limited.until timeIntervalSinceNow] > 0) {
           [button setType:NYPLRoundedButtonTypeClock];
-          [button setEndDate:limited.until];
+          if (!hasIncorrectDates) {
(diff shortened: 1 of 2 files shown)


ErnestFan (approved):

Simplified/Book/UI/NYPLBookButtonsView.m
```diff
@@ -303,9 +303,21 @@ - (void)updateButtons
       [self.book.defaultAcquisition.availability
        matchUnavailable:nil
        limited:^(NYPLOPDSAcquisitionAvailabilityLimited *const _Nonnull limited) {
+        
+        NSDate *untilDate = self.book.defaultAcquisition.availability.until;
+        NSDate *sinceDate = self.book.defaultAcquisition.availability.since;
+        
+        BOOL hasIncorrectDates = FALSE;
```
ettore, line 310: minor: in objc the convention is to use `NO` instead of FALSE

Simplified/Book/UI/NYPLBookButtonsView.m
```diff
@@ -303,9 +303,21 @@ - (void)updateButtons
...
+        
+        if (untilDate && sinceDate) {
+          hasIncorrectDates = (untilDate.timeIntervalSinceReferenceDate - sinceDate.timeIntervalSinceReferenceDate == 3600);
```
ettore, line 313: all this logic should at the very least go inside a #ifdef AXIS since it's only affecting Axis. But honestly I don't think it's the job of the app to patch these issues. I'll elaborate in the ticket.

Simplified/Book/UI/NYPLBookButtonsView.m
singhraman4282: Fixed.

Simplified/Book/UI/NYPLBookButtonsView.m
singhraman4282: Agreed.

ettore:
btw the only reason why I haven't approved this PR is because I don't think we should put a patch in the client code for this Axis issue. E.g. if Axis gets around to fix the issue the app will have incorrect code.

 IMO the ticket should be converted into a server ticket and this PR closed.

singhraman4282:
Closing since it should be addressed on server side

PR #1410, opened 2021-06-15
OE-141: Fix some axis books failing download

singhraman4282:
**What's this do?**
Fixes some axis books failing to download

https://jira.nypl.org/browse/OE-141

Follow the steps in the tickets

N/A

N/A

N/A

Raman Singh verified it

### DIFF
--- a/Simplified/AxisService/NYPLAxisPackagePathPrefixProvider.swift
+++ b/Simplified/AxisService/NYPLAxisPackagePathPrefixProvider.swift
@@ -14,28 +14,30 @@ protocol NYPLAxisPackagePathPrefixProviding {
   func getPackagePathPrefix(packageEndpoint: String?) -> String?
 }
 
-class NYPLAxisPackagePathPrefixProvider: NYPLAxisPackagePathPrefixProviding {
+struct NYPLAxisPackagePathPrefixProvider: NYPLAxisPackagePathPrefixProviding {
   
-  private var packagePathPrefix: String?
-  
-  /// Generates, stores, and returns path prefix for downloadable content from the package endpoint
-  /// extracted from the package.opf file. Returns stored value on subsequent calls.
+  /// Generates and returns path prefix for downloadable content from the package endpoint extracted from
+  /// the package.opf file.
+  ///
   /// - Parameter packageEndpoint: Package endpoint
   /// - Returns: Generated or stored (if already generated in a previous call) path prefix (optional)
   func getPackagePathPrefix(packageEndpoint: String?) -> String? {
-    if let pathPrefix = packagePathPrefix {
-      return pathPrefix
-    }
-    
     guard let packageEndpoint = packageEndpoint else {
       // No need to log error here since an error is already logged when
       // pacakgeEndpoint generation returns nil
       return nil
     }
     
-    packagePathPrefix = URL(string: packageEndpoint)?
+    let packagePathPrefix = URL(string: packageEndpoint)?
       .deletingLastPathComponent()
       .absoluteString
+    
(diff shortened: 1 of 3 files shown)


ErnestFan (approved):

Simplified/AxisService/NYPLAxisPackagePathPrefixProvider.swift
```diff
@@ -37,6 +37,10 @@ class NYPLAxisPackagePathPrefixProvider: NYPLAxisPackagePathPrefixProviding {
       .deletingLastPathComponent()
       .absoluteString
     
+    if packagePathPrefix == "./" {
```
ettore, line 40: is it a concern that even if we return nil here, we have modified the state of this class with something that looks wrong?

Also, can you add a comment to explain the rationale here? E.g. why is this bad/incorrect? it seems very specific.

Simplified/AxisService/NYPLAxisPackagePathPrefixProvider.swift
```diff
@@ -37,6 +37,10 @@ class NYPLAxisPackagePathPrefixProvider: NYPLAxisPackagePathPrefixProviding {
       .deletingLastPathComponent()
```
ettore, line 37: I didn't notice this in previous reviews, but this function is called `getXYZ` which kind of implies a function with no side effects, however at line 36 we are changing the value of the `packagePathPrefix` property. It would be make it for clearer semantics IMO to rename this function to something like `calculateXYZ` or `generateXYZ`.

ettore:
btw check out your gitmodules setup in your feature branch, I think you might have merged something incorrect. `adobe-content-filter` should not be referenced anymore.

singhraman4282:
@ettore you're right. The tests will pass once https://github.com/NYPL-Simplified/Simplified-iOS/pull/1412 has been merged.

Simplified/AxisService/NYPLAxisPackagePathPrefixProvider.swift
singhraman4282: If we return nil, it means the package and it's content are supposed to be in the root directory of the book instead of a subdirectory.
Added comment explaining the reasoning.

Simplified/AxisService/NYPLAxisPackagePathPrefixProvider.swift
singhraman4282: You're right. Changed it to a struct since we're only using it once per download session and therefore do not need to cache the value of packagePathPrefix

Simplified/AxisService/NYPLAxisPackagePathPrefixProvider.swift
```diff
@@ -14,29 +14,31 @@ protocol NYPLAxisPackagePathPrefixProviding {
...
     }
     
-    packagePathPrefix = URL(string: packageEndpoint)?
+    let packagePathPrefix = URL(string: packageEndpoint)?
       .deletingLastPathComponent()
       .absoluteString
     
+    /// For most books, package content lives inside a directory e.g. abc/def/package.opf and we use the
+    /// parent directory of the opf file as package endpoint. However, for some books, the package content
+    /// is in root directory. In that case, we do not need to provide a package endpoint.
```
ettore, line 37: minor doubt: instead of `endpoint` you mean `prefix` here?

ettore (approved):

Simplified/AxisService/NYPLAxisPackagePathPrefixProvider.swift
singhraman4282: You're right. Done

PR #1413, opened 2021-06-17
IOS-222 Dismiss book details when opening eReader

ettore:
**What's this do?**
Ties the dismissal of the book details VC with the Read action. Unfortunately this meant passing a success completion block a few layers down, but this way the downstream code makes no assumptions about whether to close or not a presented VC (in fact, it does not even know / care about that).

https://jira.nypl.org/browse/IOS-222

See ticket

This issue is a blocker for 3.7.0 release.

yes

yes

@ettore tested EPUBs and audiobooks on iPad and iPhone simulators.


Simplified/Book/UI/NYPLBookDetailViewController.m
```diff
@@ -147,20 +173,6 @@ - (void)didSelectCancelDownloadingForBookDetailView:
    cancelDownloadForBookIdentifier:self.book.identifier];
 }
 
-#pragma mark - NYPLCatalogLaneCellDelegate
```
ettore, line 150: this was just moved down for better grouping. No changes.

Simplified/Book/UI/NYPLBookDetailViewController.m
```diff
@@ -195,6 +207,32 @@ - (void)didSelectMoreBooksForLane:(NYPLCatalogLane *)lane
   [self.navigationController pushViewController:viewController animated:YES];
 }
 
+- (void)didSelectViewIssuesForBook:(NYPLBook *)book sender:(id)__unused sender
```
ettore, line 210: both these methods were just moved up to group them with other methods of `NYPLBookDetailViewDelegate`.

ErnestFan (approved):

singhraman4282 (approved):

PR #1439, opened 2021-07-07
iOS-228 Add nil check for UrlPathComponent

ErnestFan:
**What's this do?**
Add nil check for UrlPathComponent to avoid crash

[iOS-228](https://jira.nypl.org/browse/IOS-228)

N/A

N/A

N/A

N/A

@ErnestFan

### DIFF
--- a/Simplified/Book/Models/NYPLBookCoverRegistry.m
+++ b/Simplified/Book/Models/NYPLBookCoverRegistry.m
@@ -82,10 +82,14 @@ static NSUInteger const memoryCacheInMegabytes = 2;
   return URL;
 }
 
-- (NSURL *)URLForPinnedThumbnailImageOfBookIdentifier:(NSString *const)bookIdentifier
+- (nullable NSURL *)URLForPinnedThumbnailImageOfBookIdentifier:(NSString *const)bookIdentifier
 {
-  return [[self pinnedThumbnailImageDirectoryURL]
-          URLByAppendingPathComponent:[bookIdentifier SHA256]];
+  NSString *encryptedBookID = [bookIdentifier SHA256];
+  if (encryptedBookID) {
+    return [[self pinnedThumbnailImageDirectoryURL]
+            URLByAppendingPathComponent:encryptedBookID];
+  }
+  return nil;
 }
 
 - (void)coverImageForBook:(NYPLBook *)book handler:(void (^)(UIImage *image))handler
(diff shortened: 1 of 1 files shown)


ettore (commented):
this fix is fine but we should look into the consequences that this fix will bring.

That is, instead of crashing here we might crash in any downstream code that did not expect a nil, but now is. For example `pinThumbnailImageForBook:` now may receive a nil URL which may cause a crash. 

So just take a look at wherever this is used and follow up with fixes where (if!) needed... this unfortunately has a cascade effect, and you'll have to follow all the paths.

Minor improvement: you have 2 commits in this PR, one of which is just for a typo: for a cleaner history it would be good to squash that into the previous.

ErnestFan:
> this fix is fine but we should look into the consequences that this fix will bring.
> 
> That is, instead of crashing here we might crash in any downstream code that did not expect a nil, but now is. For example `pinThumbnailImageForBook:` now may receive a nil URL which may cause a crash.
> 
> So just take a look at wherever this is used and follow up with fixes where (if!) needed... this unfortunately has a cascade effect, and you'll have to follow all the paths.
> 
> Minor improvement: you have 2 commits in this PR, one of which is just for a typo: for a cleaner history it would be good to squash that into the previous.

I did look at the places that calls `pinThumbnailImageForBook:` and they should be fine with a `nil` URL. I will update the return type of `pinThumbnailImageForBook` to nullable URL for reference. Does that work @ettore ?

I will squash the commits once I made the changes, thanks for the advice.

ettore:
> I did look at the places that calls `pinThumbnailImageForBook:` and they should be fine with a `nil` URL. I will update the return type of `pinThumbnailImageForBook` to nullable URL for reference. Does that work @ettore ?

for example in NYPLBookCoverRegistry::pinThumbnailImageForBook: I see the following (repeated twice):
```
@synchronized(self) {
    // We create an empty file to mark that the thumbnail is pinned even if we do not manage to
    // finish fetching the image this application run.
    [[NSFileManager defaultManager]
     createFileAtPath:[[self URLForPinnedThumbnailImageOfBookIdentifier:book.identifier] path]
     contents:[NSData data]
     attributes:nil];
  }
```
so if `URLForPinnedThumbnailImageOfBookIdentifier` can return nil, there's a problem because the NSFileManager api does not expect a nil. 

Similarly `NYPLBookCoverRegistry ::thumbnailImageForBook:` uses `URLForPinnedThumbnailImageOfBookIdentifier` and passes its output to `[UIImage imageWithContentsOfFile:`, which in swift I think translates to `init?(contentsOfFile path: String)`, and the param of that initializer is not optional.... 

I just took a deeper look, I think these are the only two cases that need some fixing, but let me know if I'm off.

ErnestFan:
> > I did look at the places that calls `pinThumbnailImageForBook:` and they should be fine with a `nil` URL. I will update the return type of `pinThumbnailImageForBook` to nullable URL for reference. Does that work @ettore ?
> 
> for example in NYPLBookCoverRegistry::pinThumbnailImageForBook: I see the following (repeated twice):
> 
> ```
> @synchronized(self) {
>     // We create an empty file to mark that the thumbnail is pinned even if we do not manage to
>     // finish fetching the image this application run.
>     [[NSFileManager defaultManager]
>      createFileAtPath:[[self URLForPinnedThumbnailImageOfBookIdentifier:book.identifier] path]
>      contents:[NSData data]
>      attributes:nil];
>   }
> ```
> 
> so if `URLForPinnedThumbnailImageOfBookIdentifier` can return nil, there's a problem because the NSFileManager api does not expect a nil.
> 
> Similarly `NYPLBookCoverRegistry ::thumbnailImageForBook:` uses `URLForPinnedThumbnailImageOfBookIdentifier` and passes its output to `[UIImage imageWithContentsOfFile:`, which in swift I think translates to `init?(contentsOfFile path: String)`, and the param of that initializer is not optional....
> 
> I just took a deeper look, I think these are the only two cases that need some fixing, but let me know if I'm off.

You are right, `removeItemAtURL:error:` in NSFileManager does take nil for URL so I assumed the others do as well.

I updated the code and squashed the commits.

Simplified/Book/Models/NYPLBookCoverRegistry.m
```diff
@@ -286,22 +295,27 @@ - (void)pinThumbnailImageForBook:(NYPLBook *const)book
...
 {
-  @synchronized(self) {
-    [[NSFileManager defaultManager]
-     removeItemAtURL:[self URLForPinnedThumbnailImageOfBookIdentifier:bookIdentifier]
-     error:NULL];
+  NSURL *url = [self URLForPinnedThumbnailImageOfBookIdentifier:bookIdentifier];
+  if (url) {
+    @synchronized(self) {
+      [[NSFileManager defaultManager]
+       removeItemAtURL:url
```
ettore, line 316: hah, good catch

ettore (approved):

PR #1446, opened 2021-07-12
iOS-180 Fix emtpy search result if acquisition of fetched feed entries not supported

ErnestFan:
**What's this do?**
This fixes an empty search result being shown if the acquisition of all the entries are not supported.
We fetch the next page in such case.

[ios-180](https://jira.nypl.org/browse/IOS-180?filter=-1)

Go to NYPL library
Search `How to do everything with`
Multiple search results should be shown

N/A

N/A

N/A

@ErnestFan

### DIFF
--- a/Simplified/Catalog/NYPLCatalogUngroupedFeed.m
+++ b/Simplified/Catalog/NYPLCatalogUngroupedFeed.m
@@ -13,7 +13,7 @@
 @interface NYPLCatalogUngroupedFeed ()
 
 @property (nonatomic) BOOL currentlyFetchingNextURL;
-@property (nonatomic) BOOL booksFromLastFetchNotSupported;
+@property (nonatomic) BOOL noSupportedBooksInLastFetch;
 @property (nonatomic) NSMutableArray *books;
 @property (nonatomic) NSArray *facetGroups;
 @property (nonatomic) NSUInteger greatestPreparationIndex;
@@ -54,7 +54,7 @@ handler:(void (^)(NYPLCatalogUngroupedFeed *category))handler
      
     NYPLCatalogUngroupedFeed *feed = [[self alloc] initWithOPDSFeed:ungroupedFeed];
     
-    if (feed.booksFromLastFetchNotSupported) {
+    if (feed.noSupportedBooksInLastFetch) {
       [feed fetchNextPageWithCompletionHandler:handler];
     } else {
       handler(feed);
@@ -91,7 +91,7 @@ handler:(void (^)(NYPLCatalogUngroupedFeed *category))handler
     }
     [self.books addObject:book];
   }
-  self.booksFromLastFetchNotSupported = (self.books.count == 0) && (feed.entries.count > 0);
+  self.noSupportedBooksInLastFetch = (self.books.count == 0) && (feed.entries.count > 0);
 
   NSMutableArray *const entryPointFacets = [NSMutableArray array];
   NSMutableArray *const facetGroupNames = [NSMutableArray array];
@@ -191,16 +191,11 @@ handler:(void (^)(NYPLCatalogUngroupedFeed *category))handler
   [self fetchNextPageWithCompletionHandler:nil];
 }
 
-// This method might trigger a recursive loop, if and only if
-// the fetched feed contains more than one entry AND none of the entries are supported AND nextURL is not nil
-// [NYPLCatalogUngroupedFeed withURL:handler] <-> [NYPLCatalogUngroupedFeed fetchNextPageWithCompletionHandler:]
-// The handler should always be called if it is not nil, in order to exit the recursive loop
+// Continues to recursively fetch results from `nextURL` if all the books from last fetch are not supported.
+// @param completion handler, executed when booksFromLastFetchNotSupported, no next url or fetch failed
 - (void)fetchNextPageWithCompletionHandler:(nullable void (^)(NYPLCatalogUngroupedFeed *category))handler
 {
   if(self.currentlyFetchingNextURL) {
-    if (handler) {
-      handler(self);
-    }
     return;
   }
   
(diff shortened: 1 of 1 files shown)


Simplified/Catalog/NYPLCatalogUngroupedFeed.m
```diff
@@ -84,6 +91,7 @@ - (instancetype)initWithOPDSFeed:(NYPLOPDSFeed *const)feed
     }
     [self.books addObject:book];
   }
+  self.booksFromLastFetchNotSupported = (self.books.count == 0) && (feed.entries.count > 0);
```
ettore, line 94: if I understand well this means that we had some books in the feed but none of them is supported by SimplyE, correct? If so, do you think `noSupportedBooksInLastFetch` is more clear? I was struggling to understand the meaning of this property until I saw the code. 

I was also wondering if it can be a computed property instead -- I'm not sure that's ok though.

Simplified/Catalog/NYPLCatalogUngroupedFeed.m
```diff
@@ -176,11 +184,30 @@ - (void)prepareForBookIndex:(NSUInteger)bookIndex
...
+  [self fetchNextPageWithCompletionHandler:nil];
+}
+
+// This method might trigger a recursive loop, if and only if
+// the fetched feed contains more than one entry AND none of the entries are supported AND nextURL is not nil
+// [NYPLCatalogUngroupedFeed withURL:handler] <-> [NYPLCatalogUngroupedFeed fetchNextPageWithCompletionHandler:]
+// The handler should always be called if it is not nil, in order to exit the recursive loop
+- (void)fetchNextPageWithCompletionHandler:(nullable void (^)(NYPLCatalogUngroupedFeed *category))handler
+{
+  if(self.currentlyFetchingNextURL) {
```
ettore, line 200: I'm not sure we want to call the handler every time we call this method and we are currently fetching... that would result in a disparity of `handler` call compared to the actual fetch.

Simplified/Catalog/NYPLCatalogUngroupedFeed.m
```diff
@@ -195,14 +222,23 @@ - (void)prepareForBookIndex:(NSUInteger)bookIndex
...
+         }
          return;
        }
        
        [self.books addObjectsFromArray:ungroupedFeed.books];
        self.nextURL = ungroupedFeed.nextURL;
        self.currentlyFetchingNextURL = NO;
        
-       [self prepareForBookIndex:self.greatestPreparationIndex];
+       if(handler) {
```
ettore, line 235: I could be misunderstanding, but shouldn't the handler be called only when we are exiting/breaking the recursive loop?

Simplified/Catalog/NYPLCatalogUngroupedFeed.m
```diff
@@ -176,11 +184,30 @@ - (void)prepareForBookIndex:(NSUInteger)bookIndex
...
-  if(self.currentlyFetchingNextURL) return;
+  if(self.books.count - bookIndex > preloadThreshold) {
+    return;
+  }
   
-  if(!self.nextURL) return;
+  [self fetchNextPageWithCompletionHandler:nil];
+}
+
+// This method might trigger a recursive loop, if and only if
```
ettore, line 194: I would rephrase this a bit because as it is it looks like this recursive loop is not intended, but it is. So it's a little bit confusing. How about 
```
Continues to recursively fetch results from `nextURL` if all the books from last fetch are not supported.
@param handler Called at the end of the recursion.
```
?

(let me know if I'm off, I'm not 100% sure what I wrote is correct)

Simplified/Catalog/NYPLCatalogUngroupedFeed.m
ErnestFan: Right, I need to have the naming convention tips open all the time.

I don't think it can be a computed property as we use the entries of `NYPLOPDSFeed` to determine the value, and we are not storing that in `NYPLCatalogUngroupedFeed`

Simplified/Catalog/NYPLCatalogUngroupedFeed.m
ErnestFan: From my understanding, we are exiting here because we have found books that are supported and need to return the `NYPLCatalogUngroupedFeed` object to the view controller, in order to update the UI and set the delegates. The view controller will call `prepareForBookIndex:` if more books are needed.

Simplified/Catalog/NYPLCatalogUngroupedFeed.m
ettore: oh I see, you're right... sounds good.

Simplified/Catalog/NYPLCatalogUngroupedFeed.m
ettore: ah, so we are supposed to exit but then it's missing a `return` statement after the `handler` call maybe?

ettore (approved):
this looks ok for a temporary fix, considering all the limitations discussed in the ticket, whose implications however reach outside the scope of IOS-180. 

So while I am approving this PR (it does fix the bug), I would prefer to wait to merge this until we get the authorization to work on IOS-239 as a quick follow-up. Essentially I don't know if it's a good idea to live with this code for a long time.

MalcolmMcFly:
@ErnestFan what are we going to do with this PR?

ErnestFan:
> @ErnestFan what are we going to do with this PR?

@MalcolmMcFly We probably will close this once we rebuild this module as described in the TAD ([iOS-243](https://jira.nypl.org/browse/IOS-243)). I am leaving this open in case we need to release this temporary fix.

ettore:
@ErnestFan  can we close this one now?

ErnestFan:
@ettore Sure
Note: This can be closed since this part of code has been refactored in this [PR](https://github.com/NYPL-Simplified/Simplified-iOS/pull/1523)

PR #1448, opened 2021-07-14
IOS-234 Add Publisher's defaults toggle

ettore:
**What's this do?**
- adds toggle for disabling publisher's epub defaults
- minor refactor of our user settings
- adds a script to easily toggle between the stable tags and the `develop` branch of the R2 repos 

https://jira.nypl.org/browse/IOS-234

see ticket

this will appear in the next SimplyE

yes

yes

@ettore

### DIFF
--- a/scripts/checkout-r2-branch.sh
+++ b/scripts/checkout-r2-branch.sh
@@ -1,0 +1,60 @@
+#!/bin/bash
+
+# SUMMARY
+#   Use this script to easily toggle all R2 repos between the stable release
+#   we use in production builds and the most recent code on `develop`.
+#
+# SYNOPSIS
+#   ./scripts/checkout-r2-branch.sh [develop]
+#
+# PARAMETERS
+#   develop: check out the develop branch of every R2 frameworks we use.
+#            If missing, it will check out the tags we officially use in
+#            SimplyE / Open eBooks.
+#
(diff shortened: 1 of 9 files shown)


scripts/checkout-r2-branch.sh
```diff
@@ -0,0 +1,58 @@
+#!/bin/bash
+
+# SUMMARY
+#   Use this script in conjuction with build-carthage-R2-integration.sh
```
ErnestFan, line 4: Do you mean we should run `build-carthage-R2-integration.sh` after running this script, or the other way around? The `build-carthage-R2-integration.sh` script checkout the versions we use in SE/OE for each R2 repo, so running the `build-carthage-R2-integration.sh` script after this script with the `develop` branch would not work.

ettore: you're 100% right, I think my mistake was that I did not run the `build-carthage-R2-integration.sh` script after the "checkout" script. But yeah the intention is to run the checkout script first, then the build script. I'll fix the build script to avoid doing the checkout, since now that's done in its own script.

ettore: fixed -- I rebased, but the only commit that changed is the first.

ErnestFan (approved):

PR #1451, opened 2021-07-21
iOS-52 Announce download completion in BookDetailView for VoiceOver

ErnestFan:
**What's this do?**
Announce download completion to Voice Over users
Focus on "Read" button after the download complete announcement is finished

[iOS-52](https://jira.nypl.org/browse/IOS-52)

Turn On "VoiceOver"
Click on a book to go into the book detail page
Click "Get" or "Download" depends on the state of the book
An announcement of "Your book is ready! It will expire in X days" should be read out when download completed
Once the announcement finishes, it should focus on the "Read" button and says "Read button"

N/A

N/A

N/A

@ErnestFan


Simplified/Book/UI/NYPLBookDetailView.m
```diff
@@ -594,6 +604,9 @@ - (void)setState:(NYPLBookState)state
       self.buttonsView.hidden = NO;
       self.normalView.state = NYPLBookButtonsStateDownloadSuccessful;
       self.buttonsView.state = NYPLBookButtonsStateDownloadSuccessful;
+      if (self.downloadStarted) {
```
ettore, line 607: why do we need to check if the download started if we are inside the `NYPLBookStateDownloadSuccessful` case?

Simplified/Book/UI/NYPLBookDetailView.m
```diff
@@ -708,4 +721,28 @@ - (void)viewIssuesTapped {
...
+  if (UIAccessibilityIsVoiceOverRunning()) {
+    [NYPLMainThreadRun asyncIfNeeded:^{
+      UIAccessibilityPostNotification(UIAccessibilityAnnouncementNotification, message);
+    }];
+  }
+}
+
+-(void)didFinishUIAccessibilityAnnouncementNotification:(NSNotification *)notification {
+  NSString *message = notification.userInfo[UIAccessibilityAnnouncementKeyStringValue];
+  if (self.downloadStarted && [message containsString:self.normalView.messageStringForNYPLBookButtonStateSuccessful]) {
```
ettore, line 742: also here, the `downloadStarted` check seems redundant but I may be missing something

Simplified/Book/UI/NYPLBookDetailView.m
```diff
@@ -708,4 +721,28 @@ - (void)viewIssuesTapped {
   [self.detailViewDelegate didSelectViewIssuesForBook:self.book sender:self];
 }
 
+- (void)setUIAccessibilityFocusToView: (UIView *)view {
```
ettore, line 724: wondering if these 3 functions could have been put into a swift extension instead.  Not a big deal -- an argument to keep them in objc is that the rest of the objc code requires these to even compile. 

But on the other hand writing objc code is adding to our tech debt. Generally speaking, I'd say if a piece of new code can be factored out into a swift extension pretty easily, I see more advantages than disadvantages. What do you think?

Simplified/Book/UI/NYPLBookDetailView.m
ErnestFan: We only want to do the announcement if the download started and completed. Without this check, we would be doing the announcement every time user goes into the book detail page of a downloaded book.

ettore: ahh good point!

Simplified/Book/UI/NYPLBookDetailView.m
ErnestFan: You are right, it's better to have more swift code than Objc code. We are talking about an extension of `NYPLBookDetailView` in swift, right?

ettore: right -- something like `NYPLBookDetailView+Accessibility`

ErnestFan: For `didFinishUIAccessibilityAnnouncementNotification:`, should we keep it in ObjC? Since it needs access to `normalView` and `buttonsView`

ErnestFan:
@ettore Should I bump the build number and release a new build as well?

Simplified/Book/UI/NYPLBookDetailView.m
ettore: i see what you mean… yeah, probably, since you would have to expose unneeded properties.

ettore:
> @ettore Should I bump the build number and release a new build as well?

sure!

Simplified/Book/UI/NYPLBookDetailView+Accessibility.swift
```diff
@@ -0,0 +1,29 @@
...
+//  Created by Ernest Fan on 2021-07-21.
+//  Copyright © 2021 NYPL Labs. All rights reserved.
+//
+
+import Foundation
+
+@objc extension NYPLBookDetailView {
+  @objc(setUIAccessibilityFocusToView:)
+  func setUIAccessibilityFocus(to view: UIView) {
+    if UIAccessibility.isVoiceOverRunning {
```
ettore, line 14: I would put the `isVoiceOverRunning` check inside the main thread closure, just to be safe. Also in the other method below.

Simplified/Book/UI/NYPLBookDetailView+Accessibility.swift
```diff
@@ -0,0 +1,29 @@
+//
+//  NYPLBookDetailView+Accessibility.swift
+//  Simplified
+//
+//  Created by Ernest Fan on 2021-07-21.
+//  Copyright © 2021 NYPL Labs. All rights reserved.
+//
+
+import Foundation
```
ettore, line 9: this should probably be UIKit since you're using UIAccessibility. Maybe we get away with `Foundation` because we might be importing UIKit in the precompiled headers.

Simplified/Book/UI/NYPLBookDetailView.m
```diff
@@ -708,4 +721,12 @@ - (void)viewIssuesTapped {
   [self.detailViewDelegate didSelectViewIssuesForBook:self.book sender:self];
 }
 
+-(void)didFinishUIAccessibilityAnnouncementNotification:(NSNotification *)notification {
```
ettore, line 724: is this called when the system has finished reading any UIAccessibility announcement, even unrelated ones? For example, you have this view on the Catalog tab, but then you moved to the My Books tab. 

And is it possible for the message to be read twice?

ErnestFan:
For future reference, instead of posting an announcement notification, we now set download completion message to the accessibility label of the read button. The accessibility label will be reset to default after its read.

Simplified/Book/UI/NYPLBookDetailView.m
```diff
@@ -594,6 +594,11 @@ - (void)setState:(NYPLBookState)state
       self.buttonsView.hidden = NO;
       self.normalView.state = NYPLBookButtonsStateDownloadSuccessful;
       self.buttonsView.state = NYPLBookButtonsStateDownloadSuccessful;
+      if (self.downloadStarted && [self.window isKeyWindow]) {
+        // The "isKeyWindow" check avoids focusing on book detail view if it is not visible
+        [self.buttonsView setReadButtonAccessibiltyLabelWithMessage:self.normalView.messageStringForNYPLBookButtonStateSuccessful];
+        [self setUIAccessibilityFocusToView:self.buttonsView];
```
ettore, line 600: this will trigger reading our custom message, correct?

Simplified/Views/NYPLRoundedButton.swift
```diff
@@ -139,16 +139,24 @@ private let NYPLRoundedButtonPadding: CGFloat = 6.0
...
   
+  private var customAccessibilityLabel: String?
+  
   override var accessibilityLabel: String? {
     get {
+      if let label = customAccessibilityLabel {
+        self.customAccessibilityLabel = nil
+        return label
+      }
       guard !self.iconView.isHidden,
```
ettore, line 150: should we remove the `isHidden` check? so that we always say the time remaining -- Otherwise once the custom label is fetched once, the next time I think it will fall inside the guard and not say the time indication.

ettore (approved):
looks good to me, as long as the accessibility label also says the word "Button" together with the announcement / time indication

gioneill:
Two quick points from an old friend:
- "Button" is never explicitly added to accessibilityLabel. One should add (|=) the .button trait to the accessibilityTraits property on the element, and VO takes care it.
- This solution for editing a label is pretty strange for getting an announcement effect. I'll trust you if you say it works, but for someone with low-vision they might get confused because they're still associating a visual indicator to correlate against the information they're hearing, and it will be a bit odd.

I remember that this code to update the state of info bar/strip had some shakiness when shifting, which would in turn not play nice with UIAccessibility notification posting so I think I understand the challenge presented.

Super pumped A11y is getting attention! Would love to see the reading experience. There is a WWDC session explicitly for reading experiences as well!

ettore:
> Two quick points from an old friend:
> 
> * "Button" is never explicitly added to accessibilityLabel. One should add (|=) the .button trait to the accessibilityTraits property on the element, and VO takes care it.

I totally forgot about `accessibilityTraits`, but NYPLRoundedButton is a subclass of UIButton so it should have that trait.

> * This solution for editing a label is pretty strange for getting an announcement effect. I'll trust you if you say it works, but for someone with low-vision they might get confused because they're still associating a visual indicator to correlate against the information they're hearing, and it will be a bit odd.

it is definitely unusual, I agree, but Ernest mentioned that our custom announcement was overlapping with the normal VoiceOver screen reading for the button, so we came up with this workaround.  Although it does seem weird that VoiceOver does not ensure sequentiality of everything it speaks. 

> I remember that this code to update the state of info bar/strip had some shakiness when shifting, which would in turn not play nice with UIAccessibility notification posting so I think I understand the challenge presented.
> 

yeah, my thought was that if our custom announcement is not too long and we do end up saying "Read button" at the end of the accessibility label, maybe it's not too confusing.

Simplified/Views/NYPLRoundedButton.swift
ErnestFan: Removed

Simplified/Book/UI/NYPLBookDetailView.m
ErnestFan: Right

Simplified/Book/UI/NYPLBookButtonsView.h
```diff
@@ -30,5 +30,6 @@
 @property (nonatomic) BOOL showReturnButtonIfApplicable;
 
 - (void)configureForBookDetailsContext;
+- (void)setReadButtonAccessibiltyLabelWithMessage:(NSString *) message;
```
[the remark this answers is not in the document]
ErnestFan: Fixed

Simplified/Views/NYPLRoundedButton.swift
```diff
@@ -139,16 +139,26 @@ private let NYPLRoundedButtonPadding: CGFloat = 6.0
     updateColors()
   }
   
+  private var customAccessibilityLabel: String?
+  
   override var accessibilityLabel: String? {
```
ettore, line 144: since the getter is kind of hacky, can you add some docs explaining that the value that we set at line 153 is used only the first time, and after that the accessibility label will be the button label?

ettore (approved):

PR #1461, opened 2021-08-13
IOS-259 Set up Card Creator framework as a submodule

ettore:
**What's this do?**
Integrate new changes from Card Creator framework and set it up as a submodule.

https://jira.nypl.org/browse/IOS-259
PROS
1. this way we can easily change the card creator code within SimplyE. 
2. We can run the card creator unit tests when we build SimplyE.
 
CONS
The added cost in build time is negligible since the framework is very small. It has one carthage dependency, which would be managed by carthage automatically, but SimplyE and other frameworks also use the same dependency, so we can simply symlink to the SimplyE carthage folder.

see ticket

requires https://github.com/NYPL-Simplified/CardCreator-iOS/pull/36

yes

yes

@ettore


Simplified.xcodeproj/project.pbxproj
```diff
@@ -1831,7 +1892,6 @@
 		597E2722268B537E00A3CD23 /* NYPLAxisBookContentDecryptionAdapter.swift */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.swift; path = NYPLAxisBookContentDecryptionAdapter.swift; sourceTree = "<group>"; };
 		597E2731268B538B00A3CD23 /* NYPLAxisServiceAdapter.swift */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.swift; path = NYPLAxisServiceAdapter.swift; sourceTree = "<group>"; };
 		59F7F769268B513800CCA592 /* NYPLAxis.xcodeproj */ = {isa = PBXFileReference; lastKnownFileType = "wrapper.pb-project"; name = NYPLAxis.xcodeproj; path = "Axis-iOS/NYPLAxis.xcodeproj"; sourceTree = "<group>"; };
-		59F7F77A268B516300CCA592 /* NYPLAxis.xcframework */ = {isa = PBXFileReference; lastKnownFileType = wrapper.xcframework; name = NYPLAxis.xcframework; path = "Axis-iOS/NYPLAxis.xcframework"; sourceTree = "<group>"; };
```
ErnestFan, line 1834: This is a clean up from the previous work I believe? Don't think the Card creator has anything to do with Axis

ErnestFan (approved):

Simplified.xcodeproj/project.pbxproj
ettore: you're right, it does not :) it was just a dead reference in the project (I should have created a separate commit but it was so small I let it slide :)

ettore:
the failures appear to be related to the breaking changes in CardCreator. Once we merge those the build should succeed.

PR #1462, opened 2021-08-16
IOS-262 Fix POST annotations logging

ettore:
**What's this do?**
- only post reading position to server if sync is enabled
- If sync is not permitted (either by the user or the library config) we should also not log an error: therefore the check has been moved up a bit from deep down the annotation API wrappers. A similar situation exists for when creating a bookmark.

https://jira.nypl.org/browse/IOS-262

- sign in and disable sync in Settings
- verify reading resumes correctly on the same device
- verify bookmarks can be created 
- enable sync in Settings
- verify reading resumes correctly on the same device and other devices for the same user
- verify sync is enabled on the other devices
- verify bookmarks can be created and visible on other devices 

n/a

yes

n/a

@ettore


Simplified/Reader2/BusinessLogic/NYPLLastReadPositionPoster.swift
```diff
@@ -56,16 +56,22 @@ class NYPLLastReadPositionPoster {
...
+    if NYPLAnnotations.syncIsPossibleAndPermitted() {
+      let selectorValue = NYPLBookmarkFactory
+        .makeLocatorString(chapterHref: locator.href,
+                           chapterProgression: Float(chapterProgress))
+
+      postReadPosition(selectorValue: selectorValue)
+    }
   }
 
-  /// Deprecated
```
ettore, line 67: @ErnestFan I removed the deprecation notice because we are still using it... unless you remember why we put it there?

ErnestFan: I don't really remember, but here is the commit ([link](https://github.com/NYPL-Simplified/Simplified-iOS/commit/3b12829c5cea3c3a67ab1f23eaf65bcc89f0f2bc#))
I guess you added the `deprecated` because it was posting hardcoded CFI, and then you refactored it to use the `locationString` and forgot to remove the `deprecated`?
Anyways, yes removing it is fine if we are still using it

ErnestFan (approved):

PR #1469, opened 2021-09-11
iOS-265 Implement new user setting view with auto layout

ErnestFan:
**What's this do?**
Implement NYPLReaderSettingsView with autolayout

[ios-265](https://jira.nypl.org/browse/IOS-265)

1. Open SimplyE
2. Choose an e-book that is not a PDF
3. Download and open the book
4. Tap the Tt or Aa menu icon
5. There should be 4 font options and a description of font beneath it when being selected
6. The "Publisher's Default" toggle should no longer show

N/A

Yes

N/A

@ErnestFan


ErnestFan:
@ettore Can you give me the Italian of the new localized strings when you review this PR? Thanks in advance.
I will bump the build number when I commit the localized strings changes.

Simplified/Reader2/UI/NYPLReaderSettingsView.swift
```diff
@@ -0,0 +1,723 @@
...
+  private lazy var fontDescriptionLabel: UILabel = {
+    let label = UILabel()
+    label.textAlignment = .left
+    label.font = .systemFont(ofSize: 16)
+    label.numberOfLines = 0
+    // NYPLReaderSettingsView is dynamic height when being initialized,
+    // so we want to initialize this label with the possible largest amount of text
+    // to ensure there is enough room.
+    label.text = NSLocalizedString("PublisherDefaultDescriptionContent", comment: "Publisher's default description")
+    label.autoSetDimension(.height, toSize: 50)
```
ettore, line 565: will this size scale if the label text has, for example, 5 lines or more, depending on the localization? from here it seems like `50` is a fixed size, which could be a problem.

Simplified/Reader2/UI/NYPLUserSettingsVC.swift
```diff
@@ -50,63 +50,54 @@ import R2Navigator
...
-    view.publisherDefault = userSettings.publisherDefault
-    
-    self.view = view;
-  }
-
   override func viewDidLoad() {
     super.viewDidLoad()
-    self.preferredContentSize = self.view.bounds.size
+    
+    let view = NYPLReaderSettingsView(width: 330,
```
ettore, line 56: not a big deal, but I was wondering why not leaving the view creation inside `loadView` (basically everything between lines 56 - 63), and instead of adding a subview you'll keep doing `self.view = view` (in loadView).

Simplified/en.lproj/Localizable.strings
```diff
@@ -148,6 +148,10 @@
 "Next Chapter" = "Next Chapter";
 
 // NYPLReaderSettingsView
+"SpecialFontDescriptionTitle" = "Special Font";
```
ettore, line 151: For all the new strings, since we don't have localizations for Spanish and German, `SpecialFontDescriptionTitle` will be used instead of the actual localization. So it's better to use the actual English localization as key, as a fallback, so:
`"Special Font" = "Special Font";`

ettore: Italian localizations:

"Special Font" = "Font speciale" (btw where is this used in the UI?)
"Sans" = "Sans";
"Serif" = "Serif";
"Open dyslexic" = "Font per dislessia";

Simplified/en.lproj/Localizable.strings
```diff
@@ -157,6 +161,9 @@
 "DecreaseFontSize" = "Decrease font size";
 "IncreaseFontSize" = "Increase font size";
 "BrightnessSlider" = "Brightness slider";
+"PublisherDefaultButtonTitle" = "Pub";
```
ettore, line 164: "Pub" = "Editore"; (if not enough space, maybe use "Edit.")
"Publisher's Default" = "Impostazioni dell'editore";
"Show the publisher-specified fonts and layout choices in this ebook" = "Usa le impostazioni specificate dall'editore per il testo e la formattazione di questo ebook";

Simplified/Reader2/UI/NYPLReaderSettingsView.swift
```diff
@@ -0,0 +1,723 @@
...
+  
+  private func setupUI(width: CGFloat) {
+    autoSetDimension(.width, toSize: width)
+    
+    let topPadding: CGFloat = 16
+    let padding: CGFloat = 10
+    
+    backgroundColor = NYPLConfiguration.backgroundColor()
+    
+    let partialHorizontalLine = newHorizontalLine()
```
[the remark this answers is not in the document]
ErnestFan: Yes, `partialHorizontalLine` should have the same width as the stack view since its a subview of the stack view

Simplified/it.lproj/Localizable.strings
```diff
@@ -143,6 +143,10 @@
 "Next Chapter" = "Capitolo successivo";
 
 // NYPLReaderSettingsView
+"Special Font" = "Font speciale";
+"Sans" = "Sans";
+"Serif" = "Serif";
+"Open dyslexic" = "Font per dislessia";
```
ettore, line 149: I think Xcode wants the same capitalization for the key, so `"Open Dyslexic"` which is what you used elsewhere if I'm not mistaken

ettore (approved):

Simplified/it.lproj/Localizable.strings
ErnestFan: Nice catch, thanks

PR #1473, opened 2021-09-23
IOS-265 Use publisher's defaults font when selecting "Pub" button

ettore:
**What's this do?**
- Disable font overriding R2 flag when using Publisher's defaults so that the publisher's original font is used. Otherwise, the last selected font will stick when using Publisher's defaults.
- Minor fixes in localizations

https://jira.nypl.org/browse/IOS-265

Selecting the publisher's defaults button should use the publisher's font, not the last selected font in our reader settings.

required before releasing IOS-265

yes

yes

@ettore


Simplified/it.lproj/Localizable.strings
```diff
@@ -143,13 +143,13 @@
 "Next Chapter" = "Capitolo successivo";
 
 // NYPLReaderSettingsView
-"Special Font" = "Font speciale";
-"Sans" = "Sans";
+"Special Font" = "Font personalizzato";
+"Sans" = "Sans serif";
```
ErnestFan, line 147: The key should be `Sans Serif` like it's in the En localizable strings

ettore: great catch, fixed

ErnestFan (approved):

PR #1483, opened 2021-10-05
[WIP] Attempt to fix crash for Overdrive notification

ettore:
**What's this do?**
Attempts to fix a crash happening while responding to a notification meant for an Overdrive audiobook, but the current audiobook is a Findaway audiobook.

NOTE: this is temporary code only meant to illustrate approach. 

@ErnestFan can you take a look at the changes in the `updateODAudiobookManifest` method? I am unsure those changes are actually ok to do. 

https://jira.nypl.org/browse/IOS-56

[Description of any tests that need to be performed once merged goes here]

n/a

not yet

n/a

@ettore


Simplified/Book/UI/NYPLBookCellDelegate.m
```diff
@@ -237,46 +237,11 @@ - (void)openAudiobook:(NYPLBook *)book withJSON:(NSDictionary *)json decryptor:(
       DefaultAudiobookManager *const manager = [[DefaultAudiobookManager alloc]
                                                 initWithMetadata:metadata
                                                 audiobook:audiobook];
-      manager.refreshDelegate = self;
```
ettore, line 240: this was just cut and pasted into the new function `createPlayerVCForAudiobook:withBook:configuringAudiobookManager:`

Simplified/Book/UI/NYPLBookCellDelegate.m
```diff
@@ -413,33 +425,50 @@ - (void)audiobookManagerDidRequestRefresh {
...
   [[NYPLMyBooksDownloadCenter sharedDownloadCenter] startDownloadForBook:self.book];
 }
 
 #if FEATURE_OVERDRIVE
 - (void)updateODAudiobookManifest {
-  if ([[NYPLBookRegistry sharedRegistry] stateForIdentifier:self.book.identifier] == NYPLBookStateDownloadSuccessful) {
+  NYPLBookState bookState = [[NYPLBookRegistry sharedRegistry]
+                             stateForIdentifier:self.book.identifier];
+
+  if (bookState == NYPLBookStateDownloadFailed) {
```
ettore, line 441: I added this check because it seemed like if we failed the download, it seems like the observer is not going to be removed. I could be wrong though, and maybe we do need to keep it?

Simplified/Book/UI/NYPLBookCellDelegate.m
```diff
@@ -413,33 +425,50 @@ - (void)audiobookManagerDidRequestRefresh {
...
+    [[NSNotificationCenter defaultCenter] removeObserver:self];
+    [self.refreshAudiobookLock unlock];
+    return;
+  }
+
+  if (bookState == NYPLBookStateDownloadSuccessful) {
+    if (![(NSObject*)self.manager.audiobook isKindOfClass:[OverdriveAudiobook class]]) {
```
ettore, line 448: this is kind of a patch because from the stack trace it looks like we get in here even when `self.manager.audiobook` is a `NYPLAEToolkit.FindawayAudiobook` object, so not just for `OverdriveAudiobook`. I'm not sure why this is happening (I actually don't know yet how to diagnose this) but this guard seemed necessary. However, should we remove the observer? and should we unlock?

Simplified/Book/UI/NYPLBookCellDelegate.m
```diff
@@ -413,33 +425,50 @@ - (void)audiobookManagerDidRequestRefresh {
...
+      [self.refreshAudiobookLock unlock];
+      return;
+    }
+
     OverdriveAudiobook *odAudiobook = (OverdriveAudiobook *)self.manager.audiobook;
 
     NSURL *const url = [[NYPLMyBooksDownloadCenter sharedDownloadCenter] fileURLForBookIndentifier:self.book.identifier];
     NSData *const data = [NSData dataWithContentsOfURL:url];
     if (data == nil) {
+      [[NSNotificationCenter defaultCenter] removeObserver:self];
```
ettore, line 459: again here I don't know if we skipped the removal of the observer intentionally or if it is just a bug....

ErnestFan: Yeah I think we forgot to implement the removal of the observer, it should be fine to remove it before we unlock.

Simplified/Book/UI/NYPLBookCellDelegate.m
ErnestFan: Yeah, I think removing the observer and unlock is the right thing to do here.

Do you think the check `[book.distributor isEqualToString:OverdriveDistributorKey]` would help to avoid audiobooks from other distributers getting into this part of code. If so, we can add one at the beginning of this method and one where we added the observer.

Simplified/Book/UI/NYPLBookCellDelegate.m
ErnestFan: Yeah this check seems necessary

Simplified/Book/UI/NYPLBookCellDelegate.m
ettore: well, we got a successful download and we tried to get the data for a given book, I think, but somehow we found no usable data. Since we are then presenting an error for a corrupted book (see next line) I suppose we are done and we should also unlock?

Simplified/Book/UI/NYPLBookCellDelegate.m
```diff
@@ -413,33 +425,50 @@ - (void)audiobookManagerDidRequestRefresh {
...
+  NYPLBookState bookState = [[NYPLBookRegistry sharedRegistry]
+                             stateForIdentifier:self.book.identifier];
+
+  if (bookState == NYPLBookStateDownloadFailed) {
+    [[NSNotificationCenter defaultCenter] removeObserver:self];
+    [self.refreshAudiobookLock unlock];
+    return;
+  }
+
+  if (bookState == NYPLBookStateDownloadSuccessful) {
```
ettore, line 447: should this include the NYPLBookStateUsed state too (as an OR with the successful case)? I don't know the difference between the 2 states.

Simplified/Book/UI/NYPLBookCellDelegate.m
```diff
@@ -405,41 +425,70 @@ - (void)didSelectCancelForBookDownloadingCell:(NYPLBookDownloadingCell *const)ce
...
+  // download a second time before reaching a completion for the first.
   if (![self.refreshAudiobookLock tryLock]) {
     return;
   }
     
   [[NYPLBookRegistry sharedRegistry] setState:NYPLBookStateDownloadNeeded forIdentifier:self.book.identifier];
 
 #if FEATURE_OVERDRIVE
-  [[NSNotificationCenter defaultCenter] addObserver:self selector:@selector(updateODAudiobookManifest) name:NSNotification.NYPLMyBooksDownloadCenterDidChange object:nil];
+  // TODO: Configure proxy so that audiobook download fails
```
ettore, line 440: how do you trigger this method (`audiobookManagerDidRequestRefresh`)? I have been unable to see it called...

ErnestFan: [here](https://github.com/NYPL-Simplified/NYPLAudiobookToolkit/blob/master/NYPLAudiobookToolkit/UI/AudiobookPlayerViewController.swift#L681) in NYPLAudiobookToolkit
We are doing some checks in there and it should be Overdrive exclusive. Not sure how a `FindawayAudiobook` got through that.

Simplified/Book/UI/NYPLBookCellDelegate.m
ErnestFan: Yeah books with `NYPLBookStateUsed` should be included. Same here, I don't really know the difference but often times it's treated like `NYPLBookStateDownloadSuccessful` by looking at it's usage.

ettore: sounds good, I added `Used` to it.

PR #1500, opened 2021-11-05
IOS-264 Update to Xcode 13

ettore:
**What's this do?**
- Moved CardCreator back to Carthage
- Moved build of NYPLAEToolkit to a manual build since it doesn't seem to work via Carthage on CI because of permissions

Leaving both libraries as Xcode subprojects inexplicably makes Xcode 13 complain about duplicated symbols.

https://jira.nypl.org/browse/IOS-264

if the build succeeds, the regression will cover the rest.

n/a

yes

yes

@ettore

### DIFF
--- a/scripts/build-dependencies.sh
+++ b/scripts/build-dependencies.sh
@@ -8,7 +8,7 @@
 # USAGE
 #   Run this script from the root of Simplified-iOS repo:
 #
-#     ./scripts/build-3rd-party-dependencies.sh [--no-private]
+#     ./scripts/build-dependencies.sh [--no-private]
 #
 # PARAMETERS
 #   --no-private: skips integrating private secrets.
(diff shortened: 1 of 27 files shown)


ettore:
Failing because there's no Xcode 13........................... ??

ettore:
> Failing because there's no Xcode 13........................... ??

had to switch to "macOS-11" instead of "macOS-latest".

ErnestFan (approved):

scripts/build-dependencies.sh
```diff
@@ -1,9 +1,11 @@
 #!/bin/bash
 
 # SUMMARY
-#   This script integrates secrets, regenerates Readium headers, wipes
-#   the Carthage folder, and finally checks out and rebuilds all Carthage
-#   dependencies.
+#   This script integrates secrets, regenerates Readium headers, and prepares
+#   the set for the Carthage build. If this script is called to build the app
+#   woth DRM support from a CI context, it simply ends here. Otherwise,
```
ErnestFan, line 6: Do you mean `without DRM` here?

ErnestFan (approved):

scripts/build-dependencies.sh
ettore: haha I do... great catch

ettore: wait no I mean `with`! I'll fix it, the whole sentence is confusing.

PR #1509, opened 2021-11-23
iOS-232 Update sign-up logic for new CardCreation API

ErnestFan:
**What's this do?**
Update sign-up logic for new CardCreation API

[iOS-232](https://jira.nypl.org/browse/IOS-232)

We should do regression testing on the whole CardCreation process

Need to merge in CardCreator branch after QA approval and update the reference

Not yet

No

@ErnestFan

### DIFF
--- a/Simplified/SignInLogic/NYPLSignInBusinessLogic+CardCreation.swift
+++ b/Simplified/SignInLogic/NYPLSignInBusinessLogic+CardCreation.swift
@@ -9,10 +9,115 @@
 import Foundation
 import NYPLCardCreator
 
-extension JuvenileFlowCoordinator: JuvenileFlowCoordinating {}
+extension FlowCoordinator: FlowCoordinating {}
 
 extension NYPLSignInBusinessLogic {
+  /// The entry point to the regular card creation flow.
+  /// - Note: This is available only for NYPL accounts.
+  /// - Parameters:
+  ///   - completion: Always called whether the library supports
+  ///   card creation or not. If it's possible, the handler returns
+  ///   a navigation controller containing the VCs for the whole flow.
+  ///   All the client has to do is to present this navigation controller
+  ///   in whatever way it sees fit.
+  @objc
+  func startRegularCardCreation(
+    completion: @escaping (UINavigationController?, Error?) -> Void) {
+    // We don't necessary need the lock for regular card creation flow
+    // since there is no API call on eligibility check. Since the mechanism is already
+    // implemented, there is no harm to future proof this part of code.
+    guard cardCreationLock.try() else {
+      // not calling any completion because this means a flow is already going
+      return
+    }
+    
+    cardCreationIsOngoing = true
+    
+    defer {
+      cardCreationIsOngoing = false
+    }
+    
+    // If the library does not have a sign-up url, there's nothing we can do
+    guard let signUpURL = libraryAccount?.details?.signUpUrl else {
+      let description = NSLocalizedString("New patron sign up is not supported by this library.", comment: "Message describing the fact that new patron sign up is not supported by the current selected library")
+      let error = NSError(domain: NYPLErrorLogger.clientDomain,
+                          code: NYPLErrorCode.nilSignUpURL.rawValue,
+                          userInfo: [
+                            NSLocalizedDescriptionKey: description])
+      NYPLErrorLogger.logError(withCode: .nilSignUpURL,
+                               summary: "SignUp Error in Settings: nil signUp URL",
+                               metadata: [
+                                "libraryAccountUUID": libraryAccountID,
+                                "libraryAccountName": libraryAccount?.name ?? "N/A",
+      ])
+      completion(nil, error)
+      self.cardCreationLock.unlock()
+      return
+    }
 
+    // Verify if the native card creator is supported for this library,
+    // otherwise default to web
+    guard libraryAccount?.details?.supportsCardCreator ?? false else {
+      let title = NSLocalizedString("eCard",
+                                    comment: "Title for web-based card creator page")
+      let msg = NSLocalizedString("The page could not load due to a connection error.",
+                                  comment: "Message for errors loading a HTML page")
+      let webVC = RemoteHTMLViewController(URL: signUpURL,
+                                           title: title,
+                                           failureMessage: msg)
+      completion(UINavigationController(rootViewController: webVC), nil)
+      self.cardCreationLock.unlock()
+      return
+    }
+    
+    guard let coordinator = makeCardCreationCoordinator() else {
+      // This should only happen when CardCreator credentials decode failed (which is very unlikely to happen)
+      // and the errors are already logged when we retrieve the credentials
+      let description = NSLocalizedString("Sign up process fails to initiate.", comment: "Message describing the CardCreator flow failed to be initiated")
+      let error = NSError(domain: NYPLErrorLogger.clientDomain,
+                          code: NYPLErrorCode.cardCreatorCredentialsDecodeFail.rawValue,
+                          userInfo: [
+                            NSLocalizedDescriptionKey: description
+                          ])
+      completion(nil, error)
+      self.cardCreationLock.unlock()
+      return
+    }
+    
+    cardCreationCoordinator = coordinator
+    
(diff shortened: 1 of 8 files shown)


Simplified/en.lproj/Localizable.strings
```diff
@@ -223,6 +223,8 @@
 "SigningInAgree" = "By signing in, you agree to the End User License Agreement.";
 "Want a card for your child?" = "Want a card for your child?";
 "Cannot confirm library card eligibility." = "Cannot confirm library card eligibility.";
+"New patron sign up is not supported by this library." = "New patron sign up is not supported by this library.";
+"Sign up process fails to initiate." = "Sign up process fails to initiate.";
```
ErnestFan, line 227: The wording for these error messages need to be reviewed by @killerpuppytails,
and also need translation from @ettore after.

Simplified/SignInLogic/NYPLSignInBusinessLogic+CardCreation.swift
```diff
@@ -85,9 +194,14 @@ extension NYPLSignInBusinessLogic {
       completion()
       return
     }
-
-    let coordinator = juvenileCardCreationCoordinator ?? makeJuvenileCardCreationCoordinator(using: parentBarcode)
-    juvenileCardCreationCoordinator = coordinator
+    
+    guard let coordinator = cardCreationCoordinator ?? makeCardCreationCoordinator(using: parentBarcode) else {
+      Log.info(#function, "Ineligible for card creation: Card Creator credentials invalid.")
```
ettore, line 199: this should probably be a `Log.error`

Simplified/en.lproj/Localizable.strings
ettore: for the time being here's the italian translations for the current strings (I will fix them once we get the approved wording from Risa):

"New patron sign up is not supported by this library." = "La creazione di nuovi utenti non è supportata da questa biblioteca";
"Sign up process fails to initiate." = "Il process di registrazione non è potuto iniziare.";

Simplified/SignInLogic/NYPLSignInBusinessLogic+CardCreation.swift
```diff
@@ -9,10 +9,115 @@
...
+  /// The entry point to the regular card creation flow.
+  /// - Note: This is available only for NYPL accounts.
+  /// - Parameters:
+  ///   - completion: Always called whether the library supports
+  ///   card creation or not. If it's possible, the handler returns
+  ///   a navigation controller containing the VCs for the whole flow.
+  ///   All the client has to do is to present this navigation controller
+  ///   in whatever way it sees fit.
+  @objc
+  func startRegularCardCreation(
```
ettore, line 24: this method is very long, I am wondering if it can be split into smaller functions. Without doing the work (I can do it in the following days) do you think that's possible?

ettore (approved):
beside that one long function, this makes sense to me! Were you able to to test both flows end to end against the production servers?

ettore:
I filed https://jira.nypl.org/browse/IOS-295 for myself to track the remaining things I should do. @ErnestFan please add anything that you think is missing to that ticket. Thanks!

Simplified/SignInLogic/NYPLSignInBusinessLogic+CardCreation.swift
```diff
@@ -9,10 +9,115 @@
...
+      self.cardCreationLock.unlock()
+      return
+    }
+
+    // Verify if the native card creator is supported for this library,
+    // otherwise default to web
+    guard libraryAccount?.details?.supportsCardCreator ?? false else {
+      let title = NSLocalizedString("eCard",
+                                    comment: "Title for web-based card creator page")
+      let msg = NSLocalizedString("The page could not load due to a connection error.",
```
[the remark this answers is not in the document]
ErnestFan: This part of code is moved from the function `makeCardCreatorIfPossible`, so for sure we can update the error message if needed.

Simplified/SignInLogic/NYPLSignInBusinessLogic+CardCreation.swift
```diff
@@ -9,10 +9,115 @@
...
+                          userInfo: [
+                            NSLocalizedDescriptionKey: description])
+      NYPLErrorLogger.logError(withCode: .nilSignUpURL,
+                               summary: "SignUp Error in Settings: nil signUp URL",
+                               metadata: [
+                                "libraryAccountUUID": libraryAccountID,
+                                "libraryAccountName": libraryAccount?.name ?? "N/A",
+      ])
+      completion(nil, error)
+      self.cardCreationIsOngoing = false
```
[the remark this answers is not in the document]
ErnestFan: Yeah, thats a great idea.

Simplified/SignInLogic/NYPLSignInBusinessLogic+CardCreation.swift
ErnestFan: Right, nice catch

Simplified/SignInLogic/NYPLSignInBusinessLogic+CardCreation.swift
```diff
@@ -9,10 +9,115 @@
...
+    }
+    
+    coordinator.startRegularFlow { [weak self] result in
+      switch result {
+      case .success(let navVC):
+        completion(navVC, nil)
+      case .fail(let error):
+        NYPLErrorLogger.logError(error, summary: "Regular Card Creation error")
+        completion(nil, error)
+        self?.cardCreationCoordinator = nil
```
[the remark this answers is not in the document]
ErnestFan: We eventually do this when the whole CardCreator flow completes, on line 104. I did some testing and result is we do need the coordinator even after initiating the flow, otherwise the API requests fail. I can't really tell whats happening without doing some in-depth investigation, I am guessing we lost the reference to the `ISSOToken` since we don't hold any reference to the view controllers.

Simplified/SignInLogic/NYPLSignInBusinessLogic+CardCreation.swift
ErnestFan: Maybe we can put the completionHandler on line 88 into a function like `private func regularFlowCompletionHandler() -> ((_ username: String, _ PIN: String, _ userInitiated: Bool) -> Void)`. What do you think?

ErnestFan:
> beside that one long function, this makes sense to me! Were you able to to test both flows end to end against the production servers?

Yeah I tested both regular flow and juvenile flow and they both worked. One important thing I forgot to do is to update the *cardCreatorEndpointURL* and `cardCreatorVersion` in the *APIKeys* file. Without that, none of these will work

Simplified/SignInLogic/NYPLSignInBusinessLogic+CardCreation.swift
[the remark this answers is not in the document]
ettore: ah ok i that case no worries.

ettore:
> > beside that one long function, this makes sense to me! Were you able to to test both flows end to end against the production servers?
> 
> Yeah I tested both regular flow and juvenile flow and they both worked. One important thing I forgot to do is to update the `cardCreatorEndpointURL` and `cardCreatorVersion` in the `APIKeys` file. Without that, none of these will work

ah, ok… if you can update those that would be great, because i don’t what the right values should be.

Simplified/SignInLogic/NYPLSignInBusinessLogic+CardCreation.swift
ettore: that makes sense…. anyway don’t worry about doing that, thanks for the suggestion! hopefully i’ll do it soon or we can always do it at a later date. The flow did seem easy to understand to me despite the function length, which is what’s most important.

Simplified/SignInLogic/NYPLSignInBusinessLogic+CardCreation.swift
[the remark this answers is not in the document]
killerpuppytails: "The page could not load due to a connection error." = "We're sorry. Our sign up system is currently down. Please try again later."

PR #1518, opened 2021-12-17
IOS-148 Dependency inject NYPLAnnotations and update unit tests

ErnestFan:
**What's this do?**
Dependency inject NYPLAnnotations
Implement mocked NYPLAnnotations for unit testing
Update existing tests and add new test cases

[iOS-148](https://jira.nypl.org/browse/IOS-148)

N/A

N/A

N/A

N/A

@ErnestFan

### DIFF
--- a/SimplifiedTests/Mocks/NYPLAnnotationsMock.swift
+++ b/SimplifiedTests/Mocks/NYPLAnnotationsMock.swift
@@ -1,0 +1,140 @@
+//
+//  NYPLAnnotationsMock.swift
+//  SimplyETests
+//
+//  Created by Ernest Fan on 2021-12-07.
+//  Copyright © 2021 NYPL Labs. All rights reserved.
+//
+
+import Foundation
+import R2Shared
+@testable import SimplyE
+
+class NYPLAnnotationsMock: NYPLAnnotationSyncing {
+  static var serverBookmarks: [String: [NYPLReadiumBookmark]] = [String: [NYPLReadiumBookmark]]()
+  static var readingPositions: [String: NYPLBookmarkSpec] = [String: NYPLBookmarkSpec]()
+  
+  // For generating unique annotation id
+  static private var annotationCounter: Int = 0
+  
+  // Server status
+  
+  static func requestServerSyncStatus(forAccount userAccount: NYPLUserAccount,
+                                      completion: @escaping (_ enableSync: Bool) -> ()) {
+    completion(true)
+  }
+  
+  static func updateServerSyncSetting(toEnabled enabled: Bool, completion:@escaping (Bool)->()) {
+    completion(false)
+  }
+  
+  // Reading position
+  
+  static func syncReadingPosition(ofBook bookID: String?,
+                                  publication: Publication?,
+                                  toURL url:URL?,
+                                  completion: @escaping (_ readPos: NYPLReadiumBookmark?) -> ()) {
+    guard let id = bookID, let bookmarkSpec = readingPositions[id] else {
+      completion(nil)
+      return
+    }
+    let bookmarkData = bookmarkSpec.dictionaryForJSONSerialization()
+    let bookmark = NYPLBookmarkFactory.make(fromServerAnnotation: bookmarkData,
+                                            annotationType: .readingProgress,
+                                            bookID: id,
+                                            publication: publication)
+    completion(bookmark)
+  }
+  
+  static func postReadingPosition(forBook bookID: String, selectorValue: String) {
+    let bookmarkSpec = NYPLBookmarkSpec(time: Date(),
+                                        device: "",
+                                        motivation: .readingProgress,
+                                        bookID: bookID,
+                                        selectorValue: selectorValue)
+    readingPositions[bookID] = bookmarkSpec
+  }
+  
+  // Bookmark
+  
+  static func getServerBookmarks(forBook bookID:String?,
+                                 publication: Publication?,
+                                 atURL annotationURL:URL?,
+                                 completion: @escaping (_ bookmarks: [NYPLReadiumBookmark]?) -> ()) {
+    guard let id = bookID else {
+      completion(nil)
+      return
+    }
+    completion(serverBookmarks[id])
+  }
+  
+  static func deleteBookmarks(_ bookmarks: [NYPLReadiumBookmark]) {
+    for bookmark in bookmarks {
+      if let annotationID = bookmark.annotationId {
+        deleteBookmark(annotationId: annotationID, completionHandler: {_ in })
+      }
+    }
+  }
+  
+  static func deleteBookmark(annotationId: String,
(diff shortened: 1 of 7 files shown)


ErnestFan:
@ettore It's probably easier to review by commits since I fixed the indentation in `ReaderBookmarksBusinessLogicTest` and it's hard to see the actual difference.

SimplifiedTests/Mocks/NYPLAnnotationsMock.swift
```diff
@@ -0,0 +1,160 @@
...
+  
+  // Server status
+  
+  static func requestServerSyncStatus(forAccount userAccount: NYPLUserAccount,
+                                      completion: @escaping (_ enableSync: Bool) -> ()) {
+    completion(true)
+  }
+  
+  static func updateServerSyncSetting(toEnabled enabled: Bool, completion:@escaping (Bool)->()) {
+    completion(false)
```
ettore, line 29: here we are setting up the mock to always mock a failure, correct?  just checking if this is the intention.

SimplifiedTests/NYPLReaderBookmarksBusinessLogicTests.swift
```diff
@@ -194,8 +198,206 @@ class NYPLReaderBookmarksBusinessLogicTests: XCTestCase {
...
+  // MARK: - Test deleteBookmark/didDeleteBookmark
+  
+  func testDeleteBookmarkWithMatchingBookmark() throws {
+    let firstBookmarkLoc = newBookmarkR2Location(href: "Intro", chapter: "1", progressWithinChapter: 0.1, progressWithinBook: 0.1)
+    guard let firstBookmark = bookmarkBusinessLogic.addBookmark(firstBookmarkLoc) else {
+      XCTFail("Failed to create new bookmark")
+      return
+    }
+    
+    let secondBookmarkLoc = newBookmarkR2Location(href: "Chapter 1", chapter: "2", progressWithinChapter: 0.1, progressWithinBook: 0.1)
```
ettore, line 266: here maybe progressWithinBook should be higher than 0.1, since the first bookmark had the same progress. Same for the other functions below.

SimplifiedTests/NYPLReaderBookmarksBusinessLogicTests.swift
```diff
@@ -194,8 +198,206 @@ class NYPLReaderBookmarksBusinessLogicTests: XCTestCase {
...
+    guard let _ = bookmarkBusinessLogic.addBookmark(bookmarkLoc) else {
+      XCTFail("Failed to create new bookmark")
+      return
+    }
+    
+    // BookRegistry should have one bookmark even upload failed
+    // While server should not have any bookmarks
+    XCTAssertEqual(self.bookRegistryMock.readiumBookmarks(forIdentifier: self.bookIdentifier).count, 1)
+    
+    annotationsMock.failRequest = false
```
ettore, line 249: do we need to set this flag again, since we set it a few lines before?

SimplifiedTests/NYPLReaderBookmarksBusinessLogicTests.swift
```diff
@@ -194,8 +198,206 @@ class NYPLReaderBookmarksBusinessLogicTests: XCTestCase {
...
+        XCTFail("Failed to get bookmark from server")
+        return
+      }
+      XCTAssertFalse(bookmarks.contains(firstBookmark))
+    }
+  }
+  
+  func testDeleteBookmarkWithMultipleMatchingBookmarks() throws {
+    let firstBookmarkLoc = newBookmarkR2Location(href: "Intro", chapter: "1", progressWithinChapter: 0.1, progressWithinBook: 0.1)
+    guard let firstBookmark = bookmarkBusinessLogic.addBookmark(firstBookmarkLoc),
```
ettore, line 294: hmmm is this a bug in the bookmark logic? should it actually be possible to add the same bookmark 3 times? My expectation would be that the 2nd and 3rd time result in no-ops i.e. one bookmark in total being added (and check at line 308 should be 2). 

If it is a bug, I'd file a ticket for fixing that and change this test accordingly, commenting out for now the parts that are failing until we fix the ticket (also adding a TODO with the ticket).

ettore (approved):
nice tests, the annotations mock is a big complex but I can't think of a way to make it simpler right now. I do like the thorough tests.

You may have uncovered a bug in the bookmarks logic, but fixing that (assuming it is a bug) is out of scope for this ticket.

SimplifiedTests/Mocks/NYPLAnnotationsMock.swift
ErnestFan: This is actually some place holder code when I first implemented this mock class, this and a few other functions are not needed for the unit tests, but they are required due to the `NYPLAnnotationSyncing` protocol. I can remove the body of the functions if that's more appropriate.

SimplifiedTests/NYPLReaderBookmarksBusinessLogicTests.swift
ErnestFan: We set it to true above so the bookmark post request will fail, we need to set it back to false in order to get the number of bookmarks from the mocked server.

SimplifiedTests/NYPLReaderBookmarksBusinessLogicTests.swift
ErnestFan: Yeah, that's a bug in the business logic but don't think it would happen in reality since the bookmark button in the reader is turned on/off depends if the bookmark already exist. 
It can be an easy fix by adding a check calling the `bookmarkExisting(location:)` [function](https://github.com/NYPL-Simplified/Simplified-iOS/blob/develop/Simplified/Reader2/BusinessLogic/NYPLReaderBookmarksBusinessLogic.swift#L70) before we add the bookmark to book registry. I can include it in this PR if you don't mind.

SimplifiedTests/Mocks/NYPLAnnotationsMock.swift
ettore: got it, and no it’s fine as is.

SimplifiedTests/NYPLReaderBookmarksBusinessLogicTests.swift
ettore: if it’s that easy, sure, that would be good. That way this test can also be fixed… because right now it seems to suggest a behavior that in theory we don’t want.

ErnestFan: Fixed

SimplifiedTests/NYPLReaderBookmarksBusinessLogicTests.swift
ErnestFan: Fixed

PR #1523, opened 2022-01-06
iOS-239 Refactor NYPLCatalogUngroupFeed fetch logic

ErnestFan:
**What's this do?**
Refactor fetch logic from NYPLCatalogUngroupFeed and NYPLOPDSFeed into separate class
Implement mock class in order to test fetch logic
Implement unit tests for fetch logic

[iOS-239](https://jira.nypl.org/browse/IOS-239)
[TAD](https://docs.google.com/document/d/1acB9FZIsSwqNnkEkcOQYh8sluTBKcXYBo0iDAsXrerM/edit#)

Regression test on catalog search
**And** the edge case below
Go to NYPL library
Search `How to do everything with`
Multiple search results should be shown

N/A

Not yet

N/A

@ErnestFan


Simplified/Catalog/NYPLCatalogUngroupedFeed.m
```diff
@@ -27,31 +27,55 @@ @interface NYPLCatalogUngroupedFeed ()
 // attempt to fetch more books will be made.
 static NSUInteger const preloadThreshold = 100;
 
+// The current number of attempts of books fetching until we receive supported books,
+// and the maximum attempts allowed.
+static NSInteger fetchSupportedBooksRetryCount = 0;
```
ettore, line 32: is it possible to avoid using this static variable? The main problem is that these kind of variables have issues with multithreading. The other one (`fetchSupportedBooksRetryThreshold`) won't cause issues since it's constant, but it's not ideal either.

I think I know why you used them: the "fetch" method is static, so it's hard to keep a track of that state. Is it feasible to do one of these 2 things:
1. make `fetchSupportedBooksWithURL...` non static. If the method is a normal instance method, the "fetcher" instance can hold the value of the attempts being done so far. You will just need to manage ownership of the instance from wherever this is called from.
2. if `fetchSupportedBooksWithURL...` has to stay a class method, can you expand the signature to include the current iteration? that way you shouldn't need the `fetchSupportedBooksRetryCount` variable at all, since you'll be passing it along.

There's a third option, which is to just add a thread-safe protection to `fetchSupportedBooksRetryCount`, but I would not recommend it.

There's actually another problem maybe: what if this "fetch" method is called with a different URL while it's retrying? Can this happen, and if so shouldn't that use a separate retry counter? This is another argument from making this method a normal instance method.

Simplified/Catalog/NYPLCatalogUngroupedFeed.m
```diff
@@ -27,31 +27,55 @@ @interface NYPLCatalogUngroupedFeed ()
...
-       return;
-     }
-     
-     if(ungroupedFeed.type != NYPLOPDSFeedTypeAcquisitionUngrouped) {
-       NYPLLOG(@"Ignoring feed of invalid type.");
-       handler(nil);
-       return;
-     }
+    if(!ungroupedFeed) {
+      fetchSupportedBooksRetryCount = 0;
```
ettore, line 58: from this point of view, and unless the `provider withURL:...` method guarantees to call the completionHandler on the same thread it itself was being called on, this assignment can happen on a different thread than what happens at line 51, causing inconsistencies.

Simplified/Catalog/NYPLCatalogUngroupedFeed.m
```diff
@@ -27,31 +27,55 @@ @interface NYPLCatalogUngroupedFeed ()
...
+static NSInteger const fetchSupportedBooksRetryThreshold = 15;
+
 @implementation NYPLCatalogUngroupedFeed
 
-+ (void)withURL:(NSURL *)URL
-handler:(void (^)(NYPLCatalogUngroupedFeed *category))handler
-{
++ (void)fetchSupportedBooksWithURL:(NSURL *)URL
+                          provider:(Class <NYPLOPDSFeedProvider>)provider
+                           handler:(void (^)(NYPLCatalogUngroupedFeed *feed))handler {
```
ettore, line 39: is this handler a completion handler, semantically?

Simplified/Catalog/NYPLCatalogUngroupedFeed.m
```diff
@@ -27,31 +27,55 @@ @interface NYPLCatalogUngroupedFeed ()
...
-     handler([[self alloc] initWithOPDSFeed:ungroupedFeed]);
+    if(ungroupedFeed.type != NYPLOPDSFeedTypeAcquisitionUngrouped) {
+      NYPLLOG(@"Ignoring feed of invalid type.");
+      fetchSupportedBooksRetryCount = 0;
+      handler(nil);
+      return;
+    }
+    
+    NYPLCatalogUngroupedFeed *feed = [[self alloc] initWithOPDSFeed:ungroupedFeed];
+    if (feed.books.count == 0) {
```
ettore, line 71: if `feed` is nil, `feed.books.count`  will  evaluate to nil, i.e. 0, so it will try to call "fetch" once more, but this doesn't seem it's what we want. Am I wrong here?

SimplifiedTests/NYPLCatalogUngroupedFeedTests.swift
```diff
@@ -0,0 +1,113 @@
...
+    feedProvider.shouldReturnFeedWithSupportedBooks = false
+    feedProvider.numberOfFetchAllowed = 2
+    
+    NYPLCatalogUngroupedFeed.fetchSupportedBooks(with: testFeedUrl,
+                                                 provider: feedProvider) { feed in
+      result = feed
+      expectation.fulfill()
+    }
+    
+    waitForExpectations(timeout: 10, handler: nil)
```
ettore, line 68: do we need a timeout this long, and if so, why? usually 3 seconds or so are enough.

SimplifiedTests/Mocks/NYPLOPDSFeedMock.swift
```diff
@@ -0,0 +1,58 @@
...
+    
+    var filename:String = ""
+    
+    if (shouldReturnFeedWithSupportedBooks) {
+      filename = "NYPLCatalogUngroupedFeedWithSupportedBooks"
+    } else {
+      filename = "NYPLCatalogUngroupedFeedWithUnsupportedBooks"
+    }
+    
+    let feedUrl = Bundle.init(for: NYPLOPDSFeedMock.self)
```
ettore, line 45: is this feedUrl ever different from what we receive in input (`URL`) ? It seems like we calculate the feedURL at the call sites (in `NYPLCatalogUngroupedFeedTests.swift`) the same way, so we could maybe just use URL here? I could be missing something.

ettore (changes requested):
I made some suggestions with regards to the main fetch method implementation. Ideally that should be an instance method, but let me know what you think. I wonder if this actually requires a bigger refactor because of how this plugs into the caller code. Definitely open to ideas.

Simplified/Catalog/NYPLCatalogUngroupedFeed.m
ErnestFan: Right, thanks for catching the multithreading issue. The second option would be an easier fix, but I am going to explore the first option since it looks like a more appropriate implementation.

Simplified/Catalog/NYPLCatalogUngroupedFeed.m
ErnestFan: I will make this counter and the function `fetchSupportedBooksWithURL` non-static and it should resolve this issue.

Simplified/Catalog/NYPLCatalogUngroupedFeed.m
ErnestFan: Yes, I will rename it for clarity

Simplified/Book/Models/NYPLBookRegistry.m
```diff
@@ -310,98 +310,97 @@ - (void)syncResettingCache:(BOOL)shouldResetCache
...
-                                 summary:@"Unable to fetch loans"
-                                metadata:@{
-                                  @"shouldResetCache": @(shouldResetCache),
-                                  @"errorDict": error ?: @"N/A"
-                                }];
-       return;
-     }
+  [NYPLOPDSFeedFetcher fetchOPDSFeedWithUrl:[[[AccountsManager sharedInstance] currentAccount] loansUrl]
+                           shouldResetCache:shouldResetCache
+                                 completion:^(NYPLOPDSFeed * _Nullable feed, NSDictionary<NSString *,id> * _Nullable errorDict) {
```
ErnestFan, line 315: The body of this completion block contains no changes except the name of the parameters and indentation.

Simplified/OPDS/NYPLOPDSFeedFetcher.swift
```diff
@@ -0,0 +1,118 @@
...
+        }
+        return
+      }
+      
+      guard let feedXML = NYPLXML.init(data: result) else {
+        Log.info(#function, "Failed to parse data as XML.")
+        NYPLErrorLogger.logError(withCode: .feedParseFail,
+                                 summary: "NYPLOPDSFeed: Failed to parse data as XML",
+                                 metadata: [
+                                  "requestURL": url.absoluteString as Any,
```
ettore, line 85: we have a helper method for URLRequest called `loggableString`, I would just use that instead because it includes the `url` plus more info.

ErnestFan: That's what I was doing and its what causing the crash I mentioned in the ticket. The `URLRequest` object is released from the memory when it reaches this line of code, giving an error code `EXC_BAD_ACCESS`. Which is really weird since `URLRequest` is passed by value but not reference, and should not cause such error? It happens only on iOS 15 from what I am seeing. Any thoughts?

Simplified/OPDS/NYPLOPDSFeedFetcher.swift
```diff
@@ -0,0 +1,118 @@
...
+  ///   - Parameter retryCount: Current number of attempts to fetch supported books, use 0 as default value
+  ///   - Parameter completion: Always invoked at the end no matter what,
+  ///   providing an ungrouped feed object in case of success and nil otherwise.
+  class func fetchCatalogUngroupedFeed(url: URL,
+                                       retryCount: Int,
+                                       completion: @escaping (_ feed: NYPLCatalogUngroupedFeed?) -> Void) {
+    
+    if (retryCount >= fetchSupportedBooksRetryThreshold) {
+      Log.info(#function, "Retry threshold reached while fetching Catalog feed.")
+      DispatchQueue.global(qos: .userInitiated).async {
```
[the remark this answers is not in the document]
ErnestFan: The reason I did this is because this function can possibly recursively call itself multiple times, which can be a time-consuming process. Running the completion handler in `DispatchQueue.global(qos: .userInitiated)` gives it a higher priority and complete the fetch faster. I am not 100% sure on this, so feel free to make any suggestions.

Simplified/OPDS/NYPLOPDSFeedFetcher.swift
```diff
@@ -0,0 +1,118 @@
...
+//
+//  Created by Ernest Fan on 2022-01-09.
+//  Copyright © 2022 NYPL. All rights reserved.
+//
+
+import Foundation
+
+@objcMembers class NYPLOPDSFeedFetcher: NSObject {
+  // The the maximum attempts of books fetching allowed until we receive supported books.
+  private static let fetchSupportedBooksRetryThreshold = 15
```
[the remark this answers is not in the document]
ErnestFan: You are right, I am using 15 as a placeholder, so it could be anything. Let me know what number you think is appropriate.

Simplified/OPDS/NYPLOPDSFeedFetcher.swift
ettore: actually now I don't see the urlRequest instance.... maybe I got confused. We can keep the URL.

Simplified/OPDS/NYPLOPDSFeedFetcher.swift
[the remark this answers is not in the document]
ettore: I personally don't know, because it's kind of a heuristic. The server team might have a better idea of how many pages it's safe to fetch at max.

Simplified/OPDS/NYPLOPDSFeedFetcher.swift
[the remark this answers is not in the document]
ErnestFan: Sure, I will call it directly then.

Simplified/OPDS/NYPLOPDSFeedFetcher.swift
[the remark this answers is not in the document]
ErnestFan: Sure, I will ask the server team.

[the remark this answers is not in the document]
ErnestFan: Confirmed with Leonard 15 is fine as the maximum number of attempts

Simplified/OPDS/NYPLOPDSFeedFetcher.swift
```diff
@@ -0,0 +1,118 @@
...
+  
+  ///   Fetch catalog feed for given URL.
+  ///   If the returned feed does not contain supported books, it will call itself recursively
+  ///   until supported books are received or retry threshold reached.
+  ///
+  ///   - Parameter url: URL for fetching the catalog feed
+  ///   - Parameter retryCount: Current number of attempts to fetch supported books, use 0 as default value
```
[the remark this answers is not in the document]
ErnestFan: Done

Simplified/OPDS/NYPLOPDSFeedFetcher.swift
```diff
@@ -0,0 +1,118 @@
...
+  ///   - Parameter completion: Always invoked at the end no matter what,
+  ///   providing an ungrouped feed object in case of success and nil otherwise.
+  class func fetchCatalogUngroupedFeed(url: URL,
+                                       retryCount: Int,
+                                       completion: @escaping (_ feed: NYPLCatalogUngroupedFeed?) -> Void) {
+    
+    if (retryCount >= fetchSupportedBooksRetryThreshold) {
+      Log.info(#function, "Retry threshold reached while fetching Catalog feed.")
```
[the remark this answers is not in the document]
ErnestFan: Done

Simplified/OPDS/NYPLOPDSFeedFetcher.swift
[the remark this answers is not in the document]
ErnestFan: Done

Simplified/OPDS/NYPLOPDSFeedFetcher.swift
```diff
@@ -0,0 +1,118 @@
...
+  ///   - Parameter url: The URL to contact
+  ///   - Parameter shouldResetCache: Pass YES to wipe the whole cache.
+  ///   - Parameter completion: Always invoked at the end no matter what,
+  ///   providing an OPDS feed object in case of success and an dictionary containing error information otherwise.
+  class func fetchOPDSFeed(url: URL,
+                           shouldResetCache: Bool,
+                           completion: @escaping (_ feed: NYPLOPDSFeed?, _ error: [String: Any]?) -> Void) {
+    let cachePolicy: NSURLRequest.CachePolicy = shouldResetCache ? .reloadIgnoringCacheData : .useProtocolCachePolicy
+    
+    _ = NYPLNetworkExecutor.shared.GET(url,
```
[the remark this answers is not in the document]
ErnestFan: Done

ErnestFan:
Apparently the `nextURL` of a `NYPLCatalogUngroupedFeed` object can be `nil` and I didn't know that until I got a crash when doing more testing. I have update the parameter of the fetch function to accept an optional URL, and added a `guard` check to make sure we have a valid URL before performing the fetch. Also updated the unit test to include a test for nil URL. @ettore let me know if you have any thoughts on this.

Simplified/OPDS/NYPLOPDSFeedFetcher.swift
```diff
@@ -0,0 +1,122 @@
...
+    }
+  }
+  
+  ///   Fetch OPDS feed with a GET request for the given URL.
+  ///
+  ///   - Parameter url: The URL to contact
+  ///   - Parameter shouldResetCache: Pass YES to wipe the whole cache.
+  ///   - Parameter completion: Always invoked at the end no matter what,
+  ///   providing an OPDS feed object in case of success and an dictionary containing error information otherwise.
+  class func fetchOPDSFeed(url: URL,
```
ettore, line 67: similarly, I'd make the url param optional, since this function can be called from objc, similarly to what you did for the other refactored function.

Simplified/Book/Models/NYPLBookRegistry.m
```diff
@@ -310,98 +310,98 @@ - (void)syncResettingCache:(BOOL)shouldResetCache
...
-        }];
-       [NYPLErrorLogger logErrorWithCode:NYPLErrorCodeApiCall
-                                 summary:@"Unable to fetch loans"
-                                metadata:@{
-                                  @"shouldResetCache": @(shouldResetCache),
-                                  @"errorDict": error ?: @"N/A"
-                                }];
-       return;
-     }
+  [NYPLOPDSFeedFetcher fetchOPDSFeedWithUrl:[[[AccountsManager sharedInstance] currentAccount] loansUrl]
```
ettore, line 313: the `loansUrl` can be nil, as well as the `currentAccount`, so the input `url` param should be optional.

ettore (approved):
let's make the URL param optional even for `fetchOPDSFeed`, other than that I think this looks good.

PR #1524, opened 2022-01-08
IOS-311 Add CardCreator+NYPLAudiobookToolkit as gitsubmodules, use local SPM packages

ettore:
**What's this do?**
First of all this adds NYPLAudiobookToolkit and CardCreator as git submodules.

Secondly, to make it easier to modify the contents of the various Swift Packages while coding in SimplyE, I changed the SPM dependencies to refer to local instead of remote content; this way Xcode allows us to modify the SPM package sources directly from the Simplified Xcode project, instead of having to open the code from each individual Package repo.

So using git submodules for all 1st party dependencies (except NYPLUtilities) allows us to (1) still fetch the remote content automatically, and (2) make it available locally for Xcode consumption in a SimplyE/OpenE context.

NYPLUtilities is still fetched by SPM remotely to avoid any nesting issues with all the other libraries.

https://jira.nypl.org/browse/IOS-311

if build passes we should be ok. The rest will be covered by regression.

this depends on:
https://github.com/NYPL-Simplified/Axis-iOS/pull/10
https://github.com/NYPL-Simplified/CardCreator-iOS/pull/48
https://github.com/NYPL-Simplified/NYPLAEToolkit/pull/25

yes to verify build

yes

@ettore


ettore:
@ErnestFan let me know if you have any objections to this. Essentially the Xcode navigator now will show this:
Xcode-proj-nav

All the packages at the top are directly modifiable inside the Simplified Xcode project, since they are fetched by local git. The 2 at the bottom are not since they are fetched remotely by SPM.

ErnestFan (approved):

ErnestFan:
> @ErnestFan let me know if you have any objections to this. Essentially the Xcode navigator now will show this: Xcode-proj-nav
> 
> All the packages at the top are directly modifiable inside the Simplified Xcode project, since they are fetched by local git. The 2 at the bottom are not since they are fetched remotely by SPM.

Looks clean and I love it! Are we able to put breakpoints in the modules (local or even remote) and do debugging?

ettore:
> Looks clean and I love it! Are we able to put breakpoints in the modules (local or even remote) and do debugging?

yes indeed to both. 

The only thing it's not optimal is that if we for example find a bug in NYPLUtilities or other remote dependencies, to fix it we will have to open the NYPLUtilities, fix the bug, push it to a branch, and then update ALL other dependencies to refer to the same branch (temporarily). However this is not a big deal because you can do that from the SimplyE project, by simply updating the Package.swift of audiobook-ios-overdrive, CardCreator, etc. to use the temp branch instead of `main`. Then in Xcode, choose File menu > Packages > Update to latest....

PR #1529, opened 2022-01-25
OE-328 / IOS-324 Avoid unneeded catalog loading attempts

ettore:
**What's this do?**
before loading the catalog when we should compare whether we actually need to load it, for example:
- if the catalog URL is nil, there's no point in attempting to load the catalog.
- if we loaded the auth doc, we should update the catalog URL only is it's different from the current one
- we have some legacy logic that loads the catalog VC and at that time attempts to update the catalog URL and notify of an account change: it should only do that if there's an actual change in the URL

https://jira.nypl.org/browse/OE-328

I'll update the ticket 

n/a

yes

yes

@ettore


ettore:
These changes are in a fragile area of the codebase, let me know if they don't make sense to you @ErnestFan.

ErnestFan (approved):

PR #1532, opened 2022-02-05
IOS-336 Add Axis support to SimplyE

ettore:
**What's this do?**
- compiles Adobe AND Axis for SimplyE
- fixes/hacks to abort Axis decryption if not present, and likewise prevent Adobe content decryption if Axis DRM is present.

https://jira.nypl.org/browse/IOS-332

will update ticket

I think we should do 2 things:
- first review the code and see if the hacks are safe and if there's a chance for them to cause side effects.
- if that checks out, let's merge it now so that a build can start.
- We should though fix the hacks right after, and merge them to develop in a separate PR

no -- will start a release branch after this.

yes

@ettore / @ErnestFan


ErnestFan (approved):

ettore (commented):
these changes seem functionally safe for now to me, but there are indeed hacks that will cause maintainability issues. I think we should address these soon after the release of 3.8.3, given the time it will take to implement a better solution. They could go into the Demolition epic. I will file a few tickets.

PR #1534, opened 2022-02-10
iOS-339 Fixes for hacks introduced by banned book feature

ErnestFan:
**What's this do?**
Encapsulate content protection and sign-in process for each DRMs

[iOS-339](https://jira.nypl.org/browse/IOS-339)
[Banned Books PR](https://github.com/NYPL-Simplified/Simplified-iOS/pull/1532)

Regression testing for account sign-in and sign-out on SE and OE
Open and read publications protected by each DRMs or without DRM

N/A

Yes, will do before merge

No

@ErnestFan

### DIFF
--- a/Simplified/MyBooks/NYPLMyBooksDownloadCenter.m
+++ b/Simplified/MyBooks/NYPLMyBooksDownloadCenter.m
@@ -1523,7 +1523,7 @@ didFinishDownload:(BOOL)didFinishDownload
   // See Container::Open(const string& path) in container.cpp.
   //
   if(![rightsData writeToFile:[[[self fileURLForBookIndentifier:book.identifier] path]
-                               stringByAppendingString:@"_rights.xml"]
+                               stringByAppendingString:RIGHTS_XML_SUFFIX]
                    atomically:YES]) {
     NYPLLOG(@"Failed to store rights data.");
   }
--- a/Simplified/MyBooks/NYPLMyBooksDownloadCenter.m
+++ b/Simplified/MyBooks/NYPLMyBooksDownloadCenter.m
@@ -1523,7 +1523,7 @@ didFinishDownload:(BOOL)didFinishDownload
   // See Container::Open(const string& path) in container.cpp.
   //
   if(![rightsData writeToFile:[[[self fileURLForBookIndentifier:book.identifier] path]
-                               stringByAppendingString:RIGHTS_XML_SUFFIX]
+                               stringByAppendingString:ADOBE_RIGHTS_XML_SUFFIX]
                    atomically:YES]) {
     NYPLLOG(@"Failed to store rights data.");
   }
--- a/Simplified/Reader2/ReaderStackConfiguration/AdobeDRM/AdobeDRMContentProtection.swift
+++ b/Simplified/Reader2/ReaderStackConfiguration/AdobeDRM/AdobeDRMContentProtection.swift
@@ -35,7 +35,7 @@ class AdobeDRMContentProtection: ContentProtection {
       return
     }
     
-    guard let encryptionData = fetcher.fetchAdobeEncryptionData() else {
+    guard let encryptionData = fetcher.fetchAdobeEncryptionData(from: fileAsset) else {
       // .success(nil) means it is not an asset protected with Adobe DRM
       // Streamer continues to iterate over available ContentProtections
       // Don't use .failure(...) in this case, it means .epub is protected with this type of Content Protection,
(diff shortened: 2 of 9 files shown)


Simplified/MyBooks/NYPLMyBooksDownloadCenter.m
```diff
@@ -1523,7 +1523,7 @@ - (void)    adept:(__attribute__((unused)) NYPLADEPT *)adept
   // See Container::Open(const string& path) in container.cpp.
   //
   if(![rightsData writeToFile:[[[self fileURLForBookIndentifier:book.identifier] path]
-                               stringByAppendingString:@"_rights.xml"]
+                               stringByAppendingString:RIGHTS_XML_SUFFIX]
```
ettore, line 1526: is this is specific to Adobe? if so can you rename it

Simplified/SignInLogic/NYPLSignInBusinessLogic+Adept.swift
```diff
@@ -76,51 +77,33 @@ extension NYPLSignInBusinessLogic {
       VendorID: \(vendor ?? "N/A")
       """)
 
-    drmAuthorizerAdobe?
-      .authorize(withVendorID: vendor,
-                 username: username,
-                 password: password) { success, error, deviceID, userID in
-
-                  // make sure to cancel the previously scheduled selector
```
ettore, line 84: if I understand correctly, I think you moved all this code to the shared NYPLSignInBusinessLogic+DRM.swift file, correct? ideally this code should stay here, because it's related to Adobe only. The only thing that doesn't really belong here is the call to `finalizeSignIn`, because that's shared with Axis, and we only want to call that once. So here's what I'd do:
-  I'd revert back this `drmAuthorizeAdobe` function and instead of calling `finalizeSignIn` I'd simply call a `completion(success, error)`. This `completion` variable is of course an input param to drmAuthorizeAdobe. That's it here.
- Then proceeding bottom up, the caller `authorizeWithAdobe` will have some compile errors to fix....
- and finally the call to `authorizeWithAdobe` inside `drmAuthorizeUserData` will specify the closure literal with the `finalizeSignIn` call:
```
authorizeWithAdobe(userProfile: profileDoc,
                   loggingContext: loggingContext) { success, err in
  finalizeSignIn(forDRMAuthorization: success, error: err)
}
```
- Same for Axis, but as we know the completion will have a preprocessor condition:
```
drmAuthorizeAxis(username: profileDoc.authorizationIdentifier ?? "",
                 password: profileDoc.authorizationIdentifier,
                 loggingContext: loggingContext) { success, err in
#if !FEATURE_DRM_CONNECTOR
  finalizeSignIn(forDRMAuthorization: success, error: err)
#endif
    }
```
While this is still not great (`#if` stuff never is) at least this way we'll have one function (`drmAuthorizeUserData`) explicitly in charge of the `finalizeSignIn` logic, instead of having the lower function be aware of what's above them. The 2 conditional calls to `finalizeSignIn` will also be close to each other.

If we instead had 2 separate functions as you added to the bottom of NYPLSignInBusinessLogic+DRM.swift, there will still be interdependencies between the 2 functions and we'll have to jump from one to the other to understand exactly what's going on.

Let me know what you think or if I am missing something.

Simplified/SignInLogic/NYPLSignInBusinessLogic+Adept.swift
```diff
@@ -76,51 +77,33 @@ extension NYPLSignInBusinessLogic {
...
+                                  username: username,
+                                  password: password,
+                                  completion: completion)
 
     NYPLMainThreadRun.asyncIfNeeded { [weak self] in
       self?.perform(#selector(self?.dismissAfterUnexpectedDRMDelay), with: self, afterDelay: 25)
     }
   }
   
+  @objc func logInIfUserAuthorized() {
```
ettore, line 90: cool

Simplified/Reader2/ReaderStackConfiguration/AdobeDRM/AdobeDRMContentProtection.swift
```diff
@@ -59,10 +59,19 @@ class AdobeDRMContentProtection: ContentProtection {
...
 private extension Fetcher {
-  func fetchAdobeEncryptionData() -> Data? {
-    // Since publications protected by Axis DRM also contain the `META-INF` directory,
-    // we need to add another check to make sure this is not an AxisDRM protected file.
-    if let _ = try? get("/" + NYPLAxisKeysProvider().userKey).read().get() {
+  /// Fetch the encryption file data from the file asset.
+  ///
+  /// - Parameter asset: File asset containing the encryption file.
+  /// - Returns: Data of the encryption file if asset is protected by Adobe and file is found. Otherwise `nil`.
+  func fetchAdobeEncryptionData(from asset: FileAsset) -> Data? {
```
ettore, line 66: this is much better :)

PR #1536, opened 2022-02-15
iOS-149 Use NYPLErrorLogger to log events in NYPLAnnotations

ErnestFan:
**What's this do?**
Use NYPLErrorLogger to log events in NYPLAnnotations

[ios-149](https://jira.nypl.org/browse/IOS-149)
[Previous PR on NYPLAnnotations](https://github.com/NYPL-Simplified/Simplified-iOS/pull/1232#discussion_r513035443)

No testing needed

N/A

N/A

N/A

N/A

### DIFF
--- a/Simplified/Reader2/Bookmarks/NYPLAnnotations.swift
+++ b/Simplified/Reader2/Bookmarks/NYPLAnnotations.swift
@@ -395,18 +395,22 @@ protocol NYPLAnnotationSyncing: AnyObject {
     request.timeoutInterval = NYPLDefaultRequestTimeout
     
     let task = URLSession.shared.dataTask(with: request) { (data, response, error) in
-      let response = response as? HTTPURLResponse
-      if response?.statusCode == 200 {
-        Log.info(#file, "200: DELETE bookmark success")
-        completionHandler(true)
-      } else if let code = response?.statusCode {
-        Log.error(#file, "DELETE bookmark failed with server response code: \(code)")
+      guard let statusCode = (response as? HTTPURLResponse)?.statusCode,
+            statusCode == 200
+      else {
+        let metadata: [String: Any] = ["annotationId": annotationId]
+        NYPLErrorLogger.logNetworkError(error,
+                                        code: .responseFail,
+                                        summary: "NYPLAnnotations::deleteBookmark error",
+                                        request: request,
+                                        response: response,
+                                        metadata: metadata)
         completionHandler(false)
-      } else {
-        guard let error = error as NSError? else { return }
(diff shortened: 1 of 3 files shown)


ErnestFan:
@ettore There are few places in this file that uses `Log.error` and potentially we can replace them with `NYPLErrorLogger`. I have looked into them but don't really find the need to log them with `NYPLErrorLogger`. Let me know if you find it differently.
[permissionUrlRequest](https://github.com/NYPL-Simplified/Simplified-iOS/blob/develop/Simplified/Reader2/Bookmarks/NYPLAnnotations.swift#L167)
[syncSettingUrlRequest](https://github.com/NYPL-Simplified/Simplified-iOS/blob/develop/Simplified/Reader2/Bookmarks/NYPLAnnotations.swift#L223)
[annotationID(fromNetworkData:)](https://github.com/NYPL-Simplified/Simplified-iOS/blob/develop/Simplified/Reader2/Bookmarks/NYPLAnnotations.swift#L591) (This parse the annotationID from the response data received **after** we post an annotation)

Simplified/Reader2/Bookmarks/NYPLAnnotations.swift
```diff
@@ -494,21 +498,28 @@ protocol NYPLAnnotationSyncing: AnyObject {
...
+        NYPLErrorLogger.logError(withCode: .serializationFail,
+                                 summary: "NYPLAnnotations::parseAnnotationsResponse error",
+                                 metadata: metadata)
         return nil
     }
 
     guard let first = json["first"] as? [String: Any],
       let items = first["items"] as? [[String: Any]] else {
-        Log.error(#file, "Missing required key from Annotations response, or no items exist.")
+        NYPLErrorLogger.logError(withCode: .annotationFeedNoData,
```
ettore, line 520: curious to know what you think about this. So, we currently have a `opdsFeedNoData` error code. What if instead of creating a new code (`annotationFeedNoData`) for the same type of error we just had a single `noData` code? I think it will be just as informative because the `summary` of these errors will provide the needed context anyway. A single `noData` code will allow us to slice the Crashlytics data orthogonally and easily find "all the instances where we obtained "no data" where we expected to. You can see this in action for example for the `900` error code ("no URL")



I'd be in favor but let me know what you think.

ettore (approved):

Simplified/Reader2/Bookmarks/NYPLAnnotations.swift
ErnestFan: Yeah this sounds right to me, updated

PR #1537, opened 2022-02-24
iOS-349 Implement timer for saving audiobook progress when app inactive

ErnestFan:
**What's this do?**
Implement a timer to save audiobook progress when app is playing audiobook in background

[iOS-349](https://jira.nypl.org/browse/IOS-349)

Play audiobooks
Move app to background
Wait for a minute or more
Kill app
Restart app and listen to audiobook
Position should be within a minute of the last listened position

N/A

Will do

N/A

@ErnestFan


Simplified/Book/UI/NYPLBookCellDelegate.h
```diff
@@ -19,6 +19,7 @@ shared cover registry, shared download center, et cetera. */
 @property (nonatomic) NYPLBook *book;
 #if FEATURE_AUDIOBOOKS
 @property DefaultAudiobookManager *manager;
+@property NSTimer *audiobookProgressSavingTimer;
```
ettore, line 22: since this is not used outside of the class, can we move it to a class extension in the `.m` ?

Simplified/Book/UI/NYPLBookCellDelegate+Audiobooks.m
```diff
@@ -20,6 +20,8 @@
 #import "NYPLJSON.h"
 #import "NSString+NYPLStringAdditions.h"
 
+static NSInteger AudiobookProgressSavingInterval = 60;
```
ettore, line 23: can you make this a `const` and add NYPL as a prefix?

Simplified/Book/UI/NYPLBookCellDelegate+Audiobooks.m
```diff
@@ -216,6 +223,34 @@ - (void)scheduleTimerForAudiobook:(NYPLBook *)book
...
+  if (self.audiobookProgressSavingTimer) {
+    return;
+  }
+  
+  self.audiobookProgressSavingTimer = [NSTimer scheduledTimerWithTimeInterval:AudiobookProgressSavingInterval
+                                                                      repeats:YES
+                                                                        block:^(NSTimer *_Nonnull timer) {
+    [NYPLMainThreadRun asyncIfNeeded:^{
+      if ([[UIApplication sharedApplication] applicationState] == UIApplicationStateActive) {
+        [timer invalidate];
```
ettore, line 240: it seems like the only way to stop / invalidate the timer is here, that is when the app becomes active again. But what it doesn't become active again, and the user simply  stops playback from the device lock screen (without killing the app) ?  Will this loop continue to run forever ?

Simplified/Book/UI/NYPLBookCellDelegate+Audiobooks.m
```diff
@@ -216,6 +223,34 @@ - (void)scheduleTimerForAudiobook:(NYPLBook *)book
...
+        self.audiobookProgressSavingTimer = nil;
+        return;
+      }
+    }];
+    
+    if (self.manager && !self.manager.audiobook.player.isPlaying) {
+      return;
+    }
+    
+    [[NYPLBookRegistry sharedRegistry] save];
```
ettore, line 250: is this what does the saving of the listening position?

ettore: also, does this need to run in the main thread? it seems like this method saves to disk, so it can be expensive

ErnestFan: It now runs on a background thread

Simplified/Book/UI/NYPLBookCellDelegate+Audiobooks.m
ErnestFan: The timer is now suspended if user stops playback from lock screen. It will also be released if it's owner (AudiobookManager) is released from memory.

Simplified/Book/UI/NYPLBookCellDelegate+Audiobooks.m
ErnestFan: This has been renamed and moved to the `NYPLBookCellDelegate+AudiobookProgressSaving.swift`

Simplified/Book/UI/NYPLBookCellDelegate.h
ErnestFan: The timer is now part of the `AudiobookManager` object, and it will be released/cancelled when the audiobook manager is released from memory.

Simplified/Book/UI/NYPLBookCellDelegate+Audiobooks.m
```diff
@@ -110,6 +111,11 @@ - (void)openAudiobook:(NYPLBook *)book withJSON:(NSDictionary *)json decryptor:(
 
       // poll audiobook player so that we can save the reading position
       [self scheduleTimerForAudiobook:book manager:manager viewController:audiobookVC];
+      
+      [[NSNotificationCenter defaultCenter] addObserver:self
+                                               selector:@selector(scheduleAudiobookProgressSavingTimer)
+                                                   name:UIApplicationWillResignActiveNotification
+                                                 object:nil];
```
ErnestFan, line 118: For some reason, calling the Swift function here and passing in the `manager` object ends up not calling the function at all. I use a bridge function on line 225 below in order to call the Swift function. Let me know if there is workaround/fix.

Simplified/Book/UI/NYPLBookCellDelegate.h
```diff
@@ -1,5 +1,5 @@
 #if FEATURE_AUDIOBOOKS
-@import NYPLAudiobookToolkit;
```
ErnestFan, line 2: The `NYPLBookCellDelegate.h` file is now in the bridging header, importing the `NYPLAudiobookToolkit` here will result in an error with message `Module not found`, probably caused by the order of compiling. Therefore, I forward-declared the class `DefaultAudiobookManager` here to fix this issue.

Simplified/Utilities/CwlDispatch.swift
```diff
@@ -0,0 +1,151 @@
...
+//
+
+import Foundation
+
+public extension DispatchSource {
+  // An overload of timer that immediately sets the handler and schedules the timer
+  class func singleTimer(interval: DispatchTimeInterval, leeway: DispatchTimeInterval = .nanoseconds(0), queue: DispatchQueue, handler: @escaping () -> Void) -> DispatchSourceTimer {
+    let result = DispatchSource.makeTimerSource(queue: queue)
+    result.setEventHandler(handler: handler)
+    #if swift(>=4)
```
ettore, line 28: since we only support swift 5 and we are checking in this code, I wonder if we should just remove the swift < 5 cases to simplify reading this file. What do you think?

Simplified/Book/UI/NYPLBookCellDelegate+AudiobookProgressSaving.swift
```diff
@@ -0,0 +1,44 @@
...
+    manager.progressSavingTimer = DispatchSource.repeatingTimer(interval: NYPLAudiobookProgressSavingInterval) {
+      var isActive = false
+      
+      NYPLMainThreadRun.sync {
+        isActive = UIApplication.shared.applicationState == .active
+      }
+
+      DispatchQueue.global(qos: .background).sync {
+        if isActive {
+          weakManager?.progressSavingTimer = nil
```
ettore, line 34: do we need to call `suspend` before deallocating the DispatchSourceTimer?

Simplified/Book/UI/NYPLBookCellDelegate+AudiobookProgressSaving.swift
```diff
@@ -0,0 +1,44 @@
...
+    Log.info(#function, "349 - Testing DispatchSourceTimer")
+    weak var weakManager = manager
+    manager.progressSavingTimer = DispatchSource.repeatingTimer(interval: NYPLAudiobookProgressSavingInterval) {
+      var isActive = false
+      
+      NYPLMainThreadRun.sync {
+        isActive = UIApplication.shared.applicationState == .active
+      }
+
+      DispatchQueue.global(qos: .background).sync {
```
ettore, line 32: this queue is concurrent, I wonder if we may run Into thread issues here. On the other hand. NTPLBookRegistry is thread safe.

Simplified/Book/UI/NYPLBookCellDelegate+AudiobookProgressSaving.swift
```diff
@@ -0,0 +1,44 @@
...
+        if isActive {
+          weakManager?.progressSavingTimer = nil
+        } else {
+          if weakManager?.progressSavingTimer != nil {
+            NYPLBookRegistry.shared().save()
```
ettore, line 37: Can we dependency inject NYPLBookRegistry instead of using it directly? I realize this might be complicated by NYPLBookCellDelegate design though. At least maybe let's move this into a "savePosition" method, which is more explicit.

Simplified/Utilities/CwlDispatch.swift
```diff
@@ -0,0 +1,151 @@
+//
+//  CwlDispatch.swift
```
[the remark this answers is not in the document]
ErnestFan: Moved to NYPLUtilities and removed code for swift 4 or earlier version

Simplified/Book/UI/NYPLBookCellDelegate+AudiobookProgressSaving.swift
```diff
@@ -0,0 +1,44 @@
...
+private let NYPLAudiobookProgressSavingInterval: DispatchTimeInterval = .seconds(60)
+
+@objc extension NYPLBookCellDelegate {
+  
+  // Create a timer that saves the audiobook progress periodically when the app is inactive.
+  // We do save the progress when app is being killed and applicationWillTerminate: is called,
+  // but applicationWillTerminate: is not always called when users force quit the app.
+  // This method is triggered when app resigns active.
+  @objc(scheduleProgressSavingTimerForAudiobookManager:)
+  func scheduleProgressSavingTimer(for manager: DefaultAudiobookManager) {
```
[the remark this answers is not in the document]
ErnestFan: fixed

Simplified/Book/UI/NYPLBookCellDelegate+AudiobookProgressSaving.swift
ErnestFan: fixed

Simplified/Book/UI/NYPLBookCellDelegate+AudiobookProgressSaving.swift
ErnestFan: Fixed with the `savePosition` suggestion

Simplified/Book/UI/NYPLBookCellDelegate+AudiobookProgressSaving.swift
```diff
@@ -0,0 +1,56 @@
...
+    guard let manager = manager else {
+      return
+    }
+
+    let audiobookProgressSavingQueue = DispatchQueue(label: "audiobookProgressSavingQueue")
+    
+    weak var weakManager = manager
+    
+    let timer = DispatchSource.repeatingTimer(interval: NYPLAudiobookProgressSavingInterval,
+                                              queue: audiobookProgressSavingQueue) { [weak self] in
```
ettore, line 36: is the timer owning the queue?

ErnestFan: I look it up but couldn't find anything, the docs say `the dispatch queue to which to execute the installed handlers.`. Is there a specific reason why we don't want the timer to own the queue?

ettore: the problem here is that if the timer does _not_ own the queue, the queue may be dealloced at any time, because we do not retain it here. I checked the docs too and couldn't find a clear answer. 

To be on the safe side we could have the `audiobookProgressSavingQueue` as a property of the CellDelegate, and initialize it externally, in the NYPLBookCellDelegate initializer for example. Having one queue should be fine, because after all we only listen to one book at a time. What do you think?

ErnestFan: I did not thought about the queue being dealloc, nice catch.
I think your suggestion would work, what exactly do you mean when you say initialize it **externally**?

ettore: oh sorry I meant externally from this method. If you initialize it in the BookCellDelegate initializer it should work I think. It's just my first thought though...

ettore (approved):

PR #1543, opened 2022-03-19
Update fastlane in CI

ErnestFan:
**What's this do?**
Move the fastlane update task to the right position, so Gemfile is available before fastlane update
Update fastlane

We need to run `gem install bundler:1.17.2 && bundle update fastlane` in order to update fastlane to the version specify in `Gemfile`

This was tested on the `feature/ci-update-fastlane` branch and the fastlane version has updated successfully on CI

N/A

N/A

N/A

@ErnestFan


ettore (approved):
this is fine, it's good we keep up to date with fastlane in the long term, but do we know until when we need to do install that specific version of bundler?

ErnestFan:
> this is fine, it's good we keep up to date with fastlane in the long term, but do we know until when we need to do install that specific version of bundler?

I don't think we need that specific version actually, I used that because that's the suggested missing version. I will update the PR so the bundler is up to date.

ettore (approved):

PR #1548, opened 2022-04-05
OE-216 OAuth Client Credentials flow

ettore:
**What's this do?**
Implements OAuth Client Credentials flow, aka OAuth 2.0 Password Grant.
This is done at the network layer level in order to make it work across the app.

Sorry for the big PR, but I couldn't find another way to break this into smaller chunks.

The main changes involve:
- Switch FirstBook sign in to use OAuth C.C. instead of Basic auth.
- While the short lived OAuth token will authenticate each individual request after authentication, we also need to save username and password in order to refresh the token when it expires. This is done in separate entries in NYPLUserAccount
- the NYPLNetworkExecutor must now be furnished with a credentials provider for it to be able to refresh the token.
- Requests to refresh the token must be carried out serially to avoid race conditions.
- While a token refresh is in progress all other requests are blocked.

https://jira.nypl.org/browse/OE-216

I'll update the ticket

This flow only applies to Open eBooks currently, but once the server enables it on SimplyE, it should work identically there.

yes, for OE

yes

@ettore

### DIFF
--- a/Simplified/Network/NYPLNetworkExecutor.swift
+++ b/Simplified/Network/NYPLNetworkExecutor.swift
@@ -7,12 +7,7 @@
 //
 
 import Foundation
-
-@objc protocol NYPLNetworkExecuting {
-  func GET(_ reqURL: URL,
-           cachePolicy: NSURLRequest.CachePolicy,
-           completion: @escaping (_ result: Data?, _ response: URLResponse?,  _ error: Error?) -> Void) -> URLSessionDataTask
-}
+import NYPLUtilities
 
 /// Use this enum to express either-or semantics in a result.
 enum NYPLResult<SuccessInfo> {
@@ -35,7 +30,7 @@ enum NYPLResult<SuccessInfo> {
   private let responder: NYPLNetworkResponder
 
   /// Designated initializer.
-  /// - Parameter credentialsProvider: The object responsible with providing cretdentials
+  /// - Parameter credentialsProvider: The object responsible with providing credentials.
   /// - Parameter cachingStrategy: The strategy to cache responses with.
   /// - Parameter delegateQueue: The queue where callbacks will be called.
   @objc init(credentialsProvider: NYPLBasicAuthCredentialsProvider? = nil,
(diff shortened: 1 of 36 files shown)


Simplified/Network/NYPLNetworkExecutor.swift
```diff
@@ -108,80 +76,102 @@ extension NYPLNetworkExecutor {
...
+extension NYPLNetworkExecutor: NYPLRequestExecuting {
 
-  /// Performs a GET request using the specified URL, if oauth token is available, it is added to the request
+  /// Executes a given request.
   /// - Parameters:
-  ///   - reqURL: URL of the resource to GET.
+  ///   - req: The request to perform.
   ///   - completion: Always called when the resource is either fetched from
   /// the network or from the cache.
-  @objc func addBearerAndExecute(_ request: URLRequest,
```
ettore, line 164: this method was a dupe of the regular GET(), so I just removed it

Simplified/Network/NYPLNetworkExecutor.swift
```diff
@@ -108,80 +76,102 @@ extension NYPLNetworkExecutor {
...
+  func GET(_ reqURL: URL,
+           completion: @escaping (_ result: NYPLResult<Data>) -> Void) {
+    let req = request(for: reqURL)
+    executeRequest(req, completion: completion)
   }
 }
 
-// Objective-C compatibility
-extension NYPLNetworkExecutor: NYPLNetworkExecuting {
-  @objc class func bearerAuthorized(request: URLRequest) -> URLRequest {
```
ettore, line 118: this logic is already captured in the `request(for:)` method, so it's another duplication that I just removed. Now all requests in the executor are prepared with `request(for:)` and executed with `executeRequest(:completion:)`.

Simplified/Network/NYPLNetworkExecutor.swift
```diff
@@ -108,80 +76,102 @@ extension NYPLNetworkExecutor {
...
 
-  /// Performs a GET request using the specified URL
+  /// Fetches a short lived OAuth token and communicates the intention to use
+  /// that as authentication mechanism.
+  ///
   /// - Parameters:
-  ///   - reqURL: URL of the resource to GET.
-  ///   - completion: Always called when the resource is either fetched from
-  /// the network or from the cache.
-  @objc func download(_ reqURL: URL,
```
ettore, line 142: this method is unused, so I removed it.

ErnestFan (approved):

PR #1551, opened 2022-04-14
iOS-390/391 Implement audiobook download in background

ErnestFan:
**What's this do?**
Refactor code to create audiobook object for background download process
Implement downloader to download audiobooks in queue
Implement download status protocol to interfere with MyBooksDownloadCenter

[iOS-390](https://jira.nypl.org/browse/IOS-390)
[iOS-391](https://jira.nypl.org/browse/IOS-391)

Will be tested once the UI ticket is done

NYPLAudiobookToolkit and NYPLAEToolkit have been updated

Not until UI is ready

N/A

@ErnestFan

### DIFF
--- a/Simplified/Audiobooks/AudiobookManifestAdapter.swift
+++ b/Simplified/Audiobooks/AudiobookManifestAdapter.swift
@@ -9,8 +9,11 @@
 
 import Foundation
 import NYPLAudiobookToolkit
-import OverdriveProcessor
 import UIKit
+
+#if FEATURE_OVERDRIVE_AUTH
+import OverdriveProcessor
+#endif
 
 @objc enum AudiobookManifestError: Int {
   case none
--- a/Simplified/Audiobooks/AudiobookManifestAdapter.swift
+++ b/Simplified/Audiobooks/AudiobookManifestAdapter.swift
@@ -18,11 +18,11 @@ import UIKit
   case corrupted
 }
 
-@objc class AudiobookManifestHelper: NSObject {
-  @objc class func parseAudiobookManifest(book: NYPLBook,
-                                          completion: @escaping (_ manifest: [String: Any]?,
-                                                                 _ decryptor: DRMDecryptor?,
-                                                                 _ error: AudiobookManifestError) -> Void)
+@objc class AudiobookManifestAdapter: NSObject {
+  @objc class func transformAudiobookManifest(book: NYPLBook,
+                                              completion: @escaping (_ manifest: [String: Any]?,
+                                                                     _ decryptor: DRMDecryptor?,
+                                                                     _ error: AudiobookManifestError) -> Void)
   {
     guard let url = NYPLMyBooksDownloadCenter.shared().fileURL(forBookIndentifier: book.identifier),
           let data = try? Data.init(contentsOf: url),
--- a/Simplified/Audiobooks/AudiobookManifestAdapter.swift
+++ b/Simplified/Audiobooks/AudiobookManifestAdapter.swift
@@ -22,17 +22,22 @@ import OverdriveProcessor
 }
 
 @objc class AudiobookManifestAdapter: NSObject {
-  @objc class func transformAudiobookManifest(book: NYPLBook,
-                                              completion: @escaping (_ manifest: [String: Any]?,
-                                                                     _ decryptor: DRMDecryptor?,
-                                                                     _ error: AudiobookManifestError) -> Void)
+  /// Transform the book manifest into a dictionary for creating an audiobook object,
+  /// and add needed information for the dedicated DRM method.
+  @objc class func transformManifestToDictionary(for audiobook: NYPLBook?,
+                                                 fileURL: URL?,
+                                                 completion: @escaping (_ manifest: [String: Any]?,
+                                                                        _ decryptor: DRMDecryptor?,
+                                                                        _ error: AudiobookManifestError) -> Void)
   {
-    guard let url = NYPLMyBooksDownloadCenter.shared().fileURL(forBookIndentifier: book.identifier),
+    guard let book = audiobook,
+          let url = fileURL,
           let data = try? Data.init(contentsOf: url),
           let json = try? JSONSerialization.jsonObject(with: data, options: []) as? [String: Any] else {
             completion(nil, nil, .corrupted)
             return
           }
+    
     var dict: [String: Any] = json
 #if FEATURE_OVERDRIVE_AUTH
     if (book.distributor == OverdriveAPI.distributorKey) {
--- a/Simplified/Audiobooks/NYPLAudiobookDownloader.swift
+++ b/Simplified/Audiobooks/NYPLAudiobookDownloader.swift
@@ -10,7 +10,7 @@ import Foundation
 import NYPLAudiobookToolkit
 
 @objc protocol NYPLAudiobookDownloadStatusDelegate {
-  func audiobookDidUpdateDownloadProgress(progress: Float, bookID: String)
+  func audiobookDidUpdateDownloadProgress(_ progress: Float, bookID: String)
   func audiobookDidCompleteDownload(bookID: String)
   func audiobookDidCompleteDownloadFirstElement(bookID: String)
   func audiobookDidReceiveDownloadError(error: NSError?, bookID: String)
@@ -26,8 +26,13 @@ class NYPLAudiobookDownloadObject {
   }
 }
 
-@objcMembers class NYPLAudiobookDownloader: NSObject {
-  weak var delegate: NYPLAudiobookDownloadStatusDelegate?
+/// This class is designed to download audiobook files in the background.
+/// Only one audiobook would be downloaded at a time,
+/// when it completes or being cancelled, the next audiobook in queue will be downloaded.
+/// When the download fails, a second download attempt will be triggered immediately.
+/// If the case of a second failure, the audiobook will be removed and user can manually retry download.
+@objc class NYPLAudiobookDownloader: NSObject {
+  @objc weak var delegate: NYPLAudiobookDownloadStatusDelegate?
   
   private var serialQueue: DispatchQueue = DispatchQueue(label: "org.nypl.labs.NYPLAudiobooksDownloader")
   private var downloadObjects: [NYPLAudiobookDownloadObject] = []
@@ -44,15 +49,19 @@ class NYPLAudiobookDownloadObject {
       self.downloadObjects.append(NYPLAudiobookDownloadObject(bookID: bookID,
                                                               audiobookManager: audiobookManager))
     }
-    fetchNext()
+    fetchNextIfNeeded()
   }
   
-  @objc func cancelDownload(for bookID: String) {
+  @objc func cancelDownloadFetchingNextIfNeeded(for bookID: String?) {
+    guard let bookID = bookID else {
+      return
+    }
+    
     if let downloadObject = currentDownloadObject,
        downloadObject.bookID == bookID {
       cancelDownload(for: downloadObject)
       releaseCurrentDownloadObject()
-      fetchNext()
+      fetchNextIfNeeded()
     } else if let index = downloadObjects.firstIndex(where: { $0.bookID == bookID }) {
       cancelDownload(for: downloadObjects[index])
       serialQueue.async {
@@ -61,7 +70,7 @@ class NYPLAudiobookDownloadObject {
     }
   }
   
-  func audiobookManager(for bookID: String) -> DefaultAudiobookManager? {
+  @objc func audiobookManager(for bookID: String) -> DefaultAudiobookManager? {
     return downloadObject(for: bookID)?.audiobookManager
   }
   
@@ -73,14 +82,14 @@ class NYPLAudiobookDownloadObject {
     }
   }
   
-  private func fetchNext() {
+  private func fetchNextIfNeeded() {
     serialQueue.async {
       guard self.currentDownloadObject == nil else {
         return
       }
       
       if let downloadObject = self.downloadObjects.popFirst() {
-        Log.debug(#file, "Fetch initiated for \(downloadObject.bookID)")
+        Log.info(#file, "Fetch initiated for \(downloadObject.bookID)")
         downloadObject.audiobookManager.networkService.registerDelegate(self)
         downloadObject.audiobookManager.networkService.fetch()
         self.currentDownloadObject = downloadObject
@@ -91,7 +100,7 @@ class NYPLAudiobookDownloadObject {
   private func cancelDownload(for downloadObject: NYPLAudiobookDownloadObject) {
     downloadObject.audiobookManager.networkService.cancelFetch()
     downloadObject.audiobookManager.networkService.removeDelegate(self)
-    Log.debug(#file, "Fetch cancelled for \(downloadObject.bookID)")
+    Log.info(#file, "Fetch cancelled for \(downloadObject.bookID)")
   }
   
   private func downloadObject(for bookID: String) -> NYPLAudiobookDownloadObject? {
@@ -108,7 +117,7 @@ class NYPLAudiobookDownloadObject {
 extension NYPLAudiobookDownloader: AudiobookNetworkServiceDelegate {
   func audiobookNetworkService(_ audiobookNetworkService: AudiobookNetworkService, didCompleteDownloadFor spineElement: SpineElement) {
     if let downloadObject = currentDownloadObject {
-      delegate?.audiobookDidUpdateDownloadProgress(progress: audiobookNetworkService.downloadProgress,
+      delegate?.audiobookDidUpdateDownloadProgress(audiobookNetworkService.downloadProgress,
                                                    bookID: downloadObject.bookID)
     }
   }
@@ -121,10 +130,10 @@ extension NYPLAudiobookDownloader: AudiobookNetworkServiceDelegate {
     {
       delegate?.audiobookDidCompleteDownload(bookID: downloadObject.bookID)
       downloadObject.audiobookManager.networkService.removeDelegate(self)
-      Log.debug(#file, "\(downloadObject.bookID) download completed and removed")
+      Log.info(#file, "Audiobook - \(downloadObject.bookID) download completed and removed")
 
       releaseCurrentDownloadObject()
-      fetchNext()
+      fetchNextIfNeeded()
     }
   }
   
@@ -135,7 +144,7 @@ extension NYPLAudiobookDownloader: AudiobookNetworkServiceDelegate {
     if let downloadObject = currentDownloadObject {
       delegate?.audiobookDidReceiveDownloadError(error: error, bookID: downloadObject.bookID)
       releaseCurrentDownloadObject()
-      fetchNext()
+      fetchNextIfNeeded()
     }
   }
 }
--- a/Simplified/Audiobooks/NYPLAudiobookDownloader.swift
+++ b/Simplified/Audiobooks/NYPLAudiobookDownloader.swift
@@ -19,6 +19,7 @@ import NYPLAudiobookToolkit
 class NYPLAudiobookDownloadObject {
   var bookID: String
   var audiobookManager: DefaultAudiobookManager
+  var didRetryDownload: Bool = false
   
   init(bookID: String, audiobookManager: DefaultAudiobookManager) {
     self.bookID = bookID
@@ -142,9 +143,16 @@ extension NYPLAudiobookDownloader: AudiobookNetworkServiceDelegate {
   func audiobookNetworkService(_ audiobookNetworkService: AudiobookNetworkService, didReceive error: NSError?, for spineElement: SpineElement) {
     audiobookNetworkService.cancelFetch()
     if let downloadObject = currentDownloadObject {
-      delegate?.audiobookDidReceiveDownloadError(error: error, bookID: downloadObject.bookID)
-      releaseCurrentDownloadObject()
-      fetchNextIfNeeded()
+      if downloadObject.didRetryDownload {
+        Log.error(#file, "Audiobook download failed, bookID - \(downloadObject.bookID)")
+        delegate?.audiobookDidReceiveDownloadError(error: error, bookID: downloadObject.bookID)
+        releaseCurrentDownloadObject()
+        fetchNextIfNeeded()
+      } else {
+        Log.info(#file, "Audiobook retrying download, bookID - \(downloadObject.bookID)")
+        downloadObject.audiobookManager.networkService.fetch()
+        downloadObject.didRetryDownload = true
+      }
     }
   }
 }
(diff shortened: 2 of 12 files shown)


Simplified/Book/UI/NYPLBookCellDelegate+Audiobooks.m
```diff
@@ -25,48 +25,63 @@ @implementation NYPLBookCellDelegate (Audiobooks)
...
+  if (audiobookManager) {
+    [self presentAudiobook:book withAudiobookManager:audiobookManager];
     return;
   }
+  
+  [AudiobookManifestAdapter transformAudiobookManifestWithBook:book
+                                                    completion:^(NSDictionary<NSString *,id> * _Nullable json,
+                                                                 id<DRMDecryptor> _Nullable decryptor,
+                                                                 enum AudiobookManifestError error) {
+    if (error == AudiobookManifestErrorCorrupted) {
```
ettore, line 39: since `error` is actually an `enum` it would be more solid if we handled this with a `switch` because then we can list all the possible cases, and the compiler can tell us if we missed one.

Simplified/Audiobooks/AudiobookManifestAdapter.swift
```diff
@@ -0,0 +1,64 @@
...
+#endif
+
+@objc enum AudiobookManifestError: Int {
+  case none
+  case unsupported
+  case corrupted
+}
+
+@objc class AudiobookManifestAdapter: NSObject {
+  @objc class func transformAudiobookManifest(book: NYPLBook,
```
ettore, line 25: can you add some docs about what kind of transformation we're doing here? It seems like we are taking the audiobook manifest file and converting that to a Dictionary? If so I'd rename the method to something more direct, like `transformManifestToDictionary(forAudiobook:completion...)`

Simplified/MyBooks/NYPLMyBooksDownloadCenter+AudiobooksDownload.swift
```diff
@@ -0,0 +1,31 @@
...
+import Foundation
+import NYPLAudiobookToolkit
+
+extension NYPLMyBooksDownloadCenter: NYPLAudiobookDownloadStatusDelegate {
+  func audiobookDidUpdateDownloadProgress(progress: Float, bookID: String) {
+    downloadProgressDidUpdate(to: Double(progress), forBookIdentifier: bookID)
+  }
+  
+  func audiobookDidCompleteDownload(bookID: String) {
+    NYPLBookRegistry.shared().setStateWithCode(NYPLBookState.DownloadSuccessful.rawValue,
```
ettore, line 18: can you pass in the BookRegistry as a protocol instead?

Simplified/MyBooks/NYPLMyBooksDownloadCenter.m
```diff
@@ -392,8 +400,15 @@ - (void)URLSession:(__attribute__((unused)) NSURLSession *)session
      setState:NYPLBookStateDownloadFailed
      forIdentifier:book.identifier];
   }
-
+  
+#if FEATURE_AUDIOBOOKS
+  if (book.defaultBookContentType == NYPLBookContentTypeAudiobook) {
+    [self downloadAudiobookForBook:book];
+    return;
```
ettore, line 407: return is not needed here.

Simplified/MyBooks/NYPLMyBooksDownloadCenter.m
```diff
@@ -392,8 +400,15 @@ - (void)URLSession:(__attribute__((unused)) NSURLSession *)session
...
+  }
+#else
```
ettore, line 409: why don't we do the broadcast if we have feature audiobooks enabled?

Simplified/MyBooks/NYPLMyBooksDownloadCenter.m
```diff
@@ -1620,5 +1638,56 @@ - (void)downloadProgressDidUpdateTo:(double)progress forBook:(NYPLBook * _Nonnul
 }
 #endif
 
+#if FEATURE_AUDIOBOOKS
+/// Create an audiobook object and add to download queue
+- (void)downloadAudiobookForBook:(NYPLBook *)book {
```
ettore, line 1643: would it be possible to have this in a swift extension? For large methods usually that's the preference, instead of adding more objc code. But I also know there are issues sometimes with that. Just wondering.

Simplified/MyBooks/NYPLMyBooksDownloadCenter.m
```diff
@@ -1620,5 +1638,56 @@ - (void)downloadProgressDidUpdateTo:(double)progress forBook:(NYPLBook * _Nonnul
...
+        [self broadcastUpdate:book.identifier];
+        return;
+      }
+      AudiobookMetadata *metadata = [[AudiobookMetadata alloc] initWithTitle:book.title
+                                                                     authors:@[book.authors]];
+      
+      DefaultAudiobookManager *manager = [[DefaultAudiobookManager alloc] initWithMetadata:metadata
+                                                                                 audiobook:audiobook];
+      
+      [[NYPLBookRegistry sharedRegistry]
```
ettore, line 1664: it would be good to pass in the registry singleton instead of hardcoding it. Even better if it's a protocol.

Simplified/Book/UI/NYPLBookCellDelegate+Audiobooks.m
```diff
@@ -25,48 +25,63 @@ @implementation NYPLBookCellDelegate (Audiobooks)
...
-      @"fileURL": url ?: @"N/A"
-    }];
+  // If an audiobook download is in progress, we should use the existing AudiobookManager.
+  DefaultAudiobookManager *audiobookManager = [[NYPLMyBooksDownloadCenter sharedDownloadCenter] audiobookManagerForBookID:book.identifier];
+  if (audiobookManager) {
+    [self presentAudiobook:book withAudiobookManager:audiobookManager];
     return;
   }
+  
+  [AudiobookManifestAdapter transformAudiobookManifestWithBook:book
```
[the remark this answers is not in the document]
ettore: one thing to watch out is if `book` can be nil for any reason. The `transform` method does not accept an optional, so it would crash in that case.

Simplified/MyBooks/NYPLMyBooksDownloadCenter.m
```diff
@@ -1620,5 +1638,56 @@ - (void)downloadProgressDidUpdateTo:(double)progress forBook:(NYPLBook * _Nonnul
 }
 #endif
 
+#if FEATURE_AUDIOBOOKS
+/// Create an audiobook object and add to download queue
+- (void)downloadAudiobookForBook:(NYPLBook *)book {
+  [AudiobookManifestAdapter transformAudiobookManifestWithBook:book
```
ettore, line 1644: can `book` be nil?

Simplified/Audiobooks/NYPLAudiobookDownloader.swift
```diff
@@ -0,0 +1,142 @@
...
+  var bookID: String
+  var audiobookManager: DefaultAudiobookManager
+  
+  init(bookID: String, audiobookManager: DefaultAudiobookManager) {
+    self.bookID = bookID
+    self.audiobookManager = audiobookManager
+  }
+}
+
+@objcMembers class NYPLAudiobookDownloader: NSObject {
```
ettore, line 29: Lemme know what you think, but I'd be in favor of banning using `@objcMembers` if you agree. I find it confusing because it obfuscates what's actually needed by objc vs what's only used by swift.

Simplified/MyBooks/NYPLMyBooksDownloadCenter.m
```diff
@@ -1272,6 +1287,9 @@ - (void)cancelDownloadForBookIdentifier:(NSString *)identifier
     [[NYPLBookRegistry sharedRegistry]
      setState:NYPLBookStateDownloadNeeded forIdentifier:identifier];
   }
+#if FEATURE_AUDIOBOOKS
+  [self.audiobookDownloader cancelDownloadFor:identifier];
```
ettore, line 1291: are we positive `identifier` is always not nil? `cancelDownloadFor` accepts a non-optional.

ettore (commented):
overall it looks pretty good to me. What is missing for IOS-390 / 391 to be complete?

Simplified/MyBooks/NYPLMyBooksDownloadCenter.m
ErnestFan: If we do it in Swift extension, we need to expose the `broadcastUpdate` function and the `audiobookDownloader` variable. I was trying to avoid that when writing this part of code, but I can do that if you think it is a good trade off.

Simplified/MyBooks/NYPLMyBooksDownloadCenter.m
ErnestFan: We will do the broadcast once the audiobook download object is created and added to the downloader.

Simplified/Book/UI/NYPLBookCellDelegate+Audiobooks.m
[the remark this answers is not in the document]
ErnestFan: Yes it is mainly a refactor so the code for creating an audiobook / audiobook manager can be reused. Sure, I will add some guard statements.

Simplified/Audiobooks/NYPLAudiobookDownloader.swift
```diff
@@ -0,0 +1,142 @@
...
+    }
+    fetchNext()
+  }
+  
+  @objc func cancelDownload(for bookID: String) {
+    if let downloadObject = currentDownloadObject,
+       downloadObject.bookID == bookID {
+      cancelDownload(for: downloadObject)
+      releaseCurrentDownloadObject()
+      fetchNext()
```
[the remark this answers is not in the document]
ErnestFan: When the current audiobook download is cancelled, we will start fetching the next audiobook in queue. Maybe it should be called `fetchNextAudiobookInQueue`?

Simplified/Audiobooks/AudiobookManifestAdapter.swift
```diff
@@ -0,0 +1,64 @@
...
+                                                                     _ decryptor: DRMDecryptor?,
+                                                                     _ error: AudiobookManifestError) -> Void)
+  {
+    guard let url = NYPLMyBooksDownloadCenter.shared().fileURL(forBookIndentifier: book.identifier),
+          let data = try? Data.init(contentsOf: url),
+          let json = try? JSONSerialization.jsonObject(with: data, options: []) as? [String: Any] else {
+            completion(nil, nil, .corrupted)
+            return
+          }
+    var dict: [String: Any] = json
```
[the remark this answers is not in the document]
ErnestFan: Yes it is

Simplified/Audiobooks/AudiobookManifestAdapter.swift
```diff
@@ -0,0 +1,64 @@
...
+  case corrupted
+}
+
+@objc class AudiobookManifestAdapter: NSObject {
+  @objc class func transformAudiobookManifest(book: NYPLBook,
+                                              completion: @escaping (_ manifest: [String: Any]?,
+                                                                     _ decryptor: DRMDecryptor?,
+                                                                     _ error: AudiobookManifestError) -> Void)
+  {
+    guard let url = NYPLMyBooksDownloadCenter.shared().fileURL(forBookIndentifier: book.identifier),
```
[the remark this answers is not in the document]
ErnestFan: Actually I think I can just pass in an URL to save the hassle.

Simplified/MyBooks/NYPLMyBooksDownloadCenter.m
ErnestFan: Guard statement is added in the `AudiobookManifestAdapter` method.

Simplified/MyBooks/NYPLMyBooksDownloadCenter.m
ErnestFan: The `BookRegistry` singleton is widely used in this class, I am thinking of leaving it as-is now and we can fix it when we rewrite/refactor the `NYPLMyBooksDownloadCenter`. What do you think?

Simplified/Audiobooks/NYPLAudiobookDownloader.swift
ErnestFan: Sure, I will fix it

Simplified/MyBooks/NYPLMyBooksDownloadCenter.m
ErnestFan: IIRC, a `nil` NSString will become an empty string when it reaches the Swift function. In our case, it will do nothing since none of the download objects will match with an empty book identifier.

Simplified/Audiobooks/NYPLAudiobookDownloader.swift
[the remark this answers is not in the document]
ettore: I'd rather rename `cancelDownload(for:)` to make it explicit that we're trying to fetch the next too, like "cancelAndFetchNext" or something. But ideally it would be better if we can more the "fetch" part outside of the "cancel" method, unless that complicates things on the caller side significantly. Generally speaking, it's better to have more focused methods than methods that do more than one thing.

Simplified/MyBooks/NYPLMyBooksDownloadCenter.m
ettore: oh I see. Makes sense, I had not seen that.

Simplified/MyBooks/NYPLMyBooksDownloadCenter.m
ettore: hmm, unless something changed recently, if `identifier`  is nil and it's passed to a swift function that accepts only a non-optional, that's a runtime error.... I believe I witnessed this kind of crash many times. Pls let me know if you saw something different. 

However, we should probably just add a guard at the beginning of `cancelDownloadForBookIdentifier:`, since if `identifier` is nil the rest of the method probably should (or _must_) be skipped entirely.

Simplified/MyBooks/NYPLMyBooksDownloadCenter.m
ettore: ah, I see what you mean.... yeah I do think it would be a good tradeoff, because objc is just harder to deal with, and exposing a couple more things is not a big deal (audiobookDownloader could be readonly, for instance). 

However, you already wrote it and tested in objc, so I feel bad.... it can be a separate PR and perhaps a separate ticket maybe?

Simplified/MyBooks/NYPLMyBooksDownloadCenter+AudiobooksDownload.swift
ErnestFan: The reason I didn't pass it in as a protocol is because this method comes from `NYPLAudiobookDownloader`, which does not use `NYPLBookRegistry` at all. With the NYPLMyBooksDownloadCenter refactor, I am expecting the `NYPLBookRegistry` singleton will be passed in as a protocol and stored as a property of the download center, given the amount it is being called. And we will be able to call something like `self.bookRegistryProvider` instead of the singleton after the refactoring. Let me know what do you think about it.

Simplified/MyBooks/NYPLMyBooksDownloadCenter.m
ettore: yeah, that's true... now that you mention it, probably the bookRegistry should be an instance variable of the DownloadCenter.... but yes, that's a different story for a different ticket :)

Simplified/MyBooks/NYPLMyBooksDownloadCenter.m
ErnestFan: Sure, rewriting this in Swift should not take long, I will include it in this PR.

Simplified/MyBooks/NYPLMyBooksDownloadCenter+AudiobooksDownload.swift
ettore: yeah, good point. The audiobook downloader probably shouldn't even know about the bookRegistry. On the other hand, my plan is to not have a "Download Center" abstraction in the MyBooks refactor ;) In any case, you're right, let's not worry about it now since we're going to rewrite the DownloadCenter anyway.

Simplified/Audiobooks/NYPLAudiobookDownloader.swift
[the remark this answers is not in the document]
ErnestFan: I didn't put two methods together because user is allowed to cancel downloads that has not started, aka still in the queue waiting. For such case, we do not want to fetch the next audiobook in queue.

Simplified/MyBooks/NYPLMyBooksDownloadCenter+AudiobooksDownload.swift
```diff
@@ -9,8 +9,54 @@
 import Foundation
 import NYPLAudiobookToolkit
 
+extension NYPLMyBooksDownloadCenter {
+  /// Create an audiobook object and add to download queue
+  @objc(downloadAudiobookForBook:)
+  func downloadAudiobook(for book: NYPLBook?) {
+    guard let book = book,
+          let url = self.fileURL(forBookIndentifier: book.identifier) else {
+      return
```
ettore, line 18: can you add a `Log.error` here? Make sure to log both `book.loggableShortString` and `url` so that we can know which one is nil

Simplified/MyBooks/NYPLMyBooksDownloadCenter+AudiobooksDownload.swift
```diff
@@ -9,8 +9,54 @@
...
+    }
+    
+    AudiobookManifestAdapter.transformManifestToDictionary(for: book,
+                                                              fileURL: url) { json, decryptor, error in
+      if error == .none {
+        guard let audiobook = AudiobookFactory.audiobook(json, decryptor: decryptor) else {
+          Log.info(#file, "Audiobook initiate failed for book id - \(book.identifier)")
```
ettore, line 25: this looks like a `Log.error` more than `info`. And I'd log `book.loggableShortString` instead of just the `id`

Simplified/MyBooks/NYPLMyBooksDownloadCenter+AudiobooksDownload.swift
```diff
@@ -9,8 +9,54 @@
...
+        let manager = DefaultAudiobookManager(metadata: metadata, audiobook: audiobook)
+        
+        NYPLBookRegistry.shared().setStateWithCode(NYPLBookState.Downloading.rawValue, forIdentifier: book.identifier)
+        
+        self.downloadProgressDidUpdate(to: Double(manager.networkService.downloadProgress),
+                                       forBookIdentifier: book.identifier)
+        
+        self.audiobookDownloader.downloadAudiobook(for: book.identifier, audiobookManager: manager)
+      } else {
+        Log.info(#file, "Audiobook manifest corrupted/unsupported")
```
ettore, line 41: info --> error. Can you also add the book.loggableString to the logged statement?

ettore (approved):

Simplified/MyBooks/NYPLMyBooksDownloadCenter+AudiobooksDownload.swift
```diff
@@ -0,0 +1,111 @@
...
+    let metadata = [
+      "book": book.loggableShortString(),
+      "fileURL": url.absoluteString
+    ]
+    
+    AudiobookManifestAdapter.transformManifestToDictionary(for: book,
+                                                              fileURL: url) { json, decryptor, error in
+      if error == .none,
+        let json = json {
+        AudioBookVendorsHelper.updateVendorKey(book: json) { [weak self] error in
```
ErnestFan, line 36: @ettore I missed the part which updates the vendor DRM key when refactoring the code from `NYPLBookCellDelegate+Audiobooks`. I have updated the code and added more documentation above.

Simplified/MyBooks/NYPLMyBooksDownloadCenter+AudiobooksDownload.swift
```diff
@@ -0,0 +1,111 @@
...
+          guard let audiobook = AudiobookFactory.audiobook(json, decryptor: decryptor) else {
+            if let error = error {
+              NYPLErrorLogger.logError(error,
+                                       summary: "Audiobooks: DRM Error",
+                                       metadata: metadata)
+            } else {
+              Log.error(#file, "Audiobooks: unsupported - \(String(describing: book.loggableShortString))")
+            }
+            
+            // TODO: Handle Error, set state in book registry
```
ErnestFan, line 46: @ettore I am waiting for Risa on the behaviour of this failure. Will update once I heard back from her. Same for the one below.

ettore: sounds good... btw generally speaking, feel free to file a ticket and do things you forgot on subsequent PRs if that's easier... after all we're merging to a feature branch, so you have a little more freedom :)

Simplified/Audiobooks/NYPLAudiobookDownloader.swift
```diff
@@ -142,9 +143,17 @@ extension NYPLAudiobookDownloader: AudiobookNetworkServiceDelegate {
   func audiobookNetworkService(_ audiobookNetworkService: AudiobookNetworkService, didReceive error: NSError?, for spineElement: SpineElement) {
     audiobookNetworkService.cancelFetch()
     if let downloadObject = currentDownloadObject {
-      delegate?.audiobookDidReceiveDownloadError(error: error, bookID: downloadObject.bookID)
-      releaseCurrentDownloadObject()
-      fetchNextIfNeeded()
+      Log.info(#file, "Audiobook did receive download error, bookID - \(downloadObject.bookID)")
```
ettore, line 146: this log line seems redundant, since you're already logging in the `if` below.

Simplified/Audiobooks/NYPLAudiobookDownloader.swift
```diff
@@ -142,9 +143,17 @@ extension NYPLAudiobookDownloader: AudiobookNetworkServiceDelegate {
...
+      if downloadObject.didRetryDownload {
+        Log.info(#file, "Audiobook download failed, bookID - \(downloadObject.bookID)")
```
ettore, line 148: very minor: I'd turn this into a `Log.error`

Simplified/MyBooks/NYPLMyBooksDownloadCenter+AudiobooksDownload.swift
ettore: the new changes look good to me. If you have a lot more changes it might be better to do them in a subsequent PR since this one is getting a bit hard to follow.

PR #1554, opened 2022-04-22
IOS-400 Upgrade to R2 2.3.0

ettore:
**What's this do?**
Upgrades R2 and removes some related deprecations.

https://jira.nypl.org/browse/IOS-400

will update ticket, but likely a regression will cover this

n/a

yes

n/a

I did


Simplified/Reader2/ReaderStackConfiguration/LibraryService.swift
```diff
@@ -94,9 +92,13 @@ final class LibraryService: Loggable {
...
-  private func preparePresentation(of publication: Publication, book: NYPLBook) {
-    // If the book is a webpub, it means it is loaded remotely from a URL, and it doesn't need to be added to the publication server.
-    guard publication.format != .webpub else {
+  private func preparePresentation(of publication: Publication) {
+    // What we want to avoid here it to add a webPub to the publication server,
+    // because there's no need to do that if it is loaded remotely from a URL.
+    // Since webPub is not a Publication.Profile, and we can only open
+    // publications that conform to a given Profile, if the Publication
+    // contains no conforming profiles, we know we cannot process it.
+    guard !publication.metadata.conformsTo.isEmpty else {
```
ettore, line 101: [waiting on confirmation](https://readium.slack.com/archives/C4APPBHUZ/p1650588067898029) for this, it doesn't match 1 to 1 with the previous code but it should be ok. However, if "WebPub" becomes a Publication.Profile in R2 in the future, this will break.

ErnestFan (approved):

Simplified/Reader2/ReaderStackConfiguration/LibraryService.swift
ettore: I changed this check per recommendation from Mickaël.

PR #1555, opened 2022-05-05
iOS-392 Update UI for audiobook background download

ErnestFan:
**What's this do?**
Add new book state for audiobook download and update related unit test
Update book detail view UI for audiobook

[iOS-392](https://jira.nypl.org/browse/IOS-392)

Download an audiobook
Check if the UI changes when the download cancel/fail/succeed

N/A

Will do when everything is ready

N/A

@ErnestFan


Simplified/Book/UI/NYPLBookDetailDownloadFailedView.m
```diff
@@ -56,14 +60,23 @@ - (void)traitCollectionDidChange:(UITraitCollection *)previousTraitCollection {
...
+  /// If this is presenting an error message which takes more than 1 line,
+  /// we need to set the message to an empty string when this view is being hidden,
+  /// so the other views can resize to the height they need.
+  if (hidden) {
+    self.messageLabel.text = @"";
+  }
+}
 
 - (void)updateColors {
+  self.backgroundColor = [NYPLConfiguration mainColor];
```
ErnestFan, line 75: The original logic does not set the background color when it is iOS 12.0+ and in Light mode. This fixes it. Same for the one below.

Simplified/Book/UI/NYPLBookDetailView.m
```diff
@@ -610,9 +610,17 @@ - (void)setState:(NYPLBookState)state
...
       self.buttonsView.state = NYPLBookButtonsStateDownloadInProgress;
       break;
+    case NYPLBookStateDownloadingUsable:
+      self.downloadFailedView.hidden = YES;
+      [self hideDownloadingView:NO];
+      self.buttonsView.hidden = NO;
+      self.buttonsView.state = NYPLBookButtonsStateDownloadingUsable;
+      break;
     case NYPLBookStateDownloadFailed:
+      self.downloadFailedView.showAudiobookError = self.book.defaultBookContentType == NYPLBookContentTypeAudiobook;
```
ettore, line 620: since this view is reused (I'm guessing), I was wondering if we need to reset the `showAudiobookError` flag to `NO` in the other cases of the switch, to avoid for example showing the error after a successful retry.

ettore (approved):

Simplified/Book/UI/NYPLBookDetailView.m
ErnestFan: This flag set an audiobook specific error message to the `message label` within the `downloadFailedView`. Even if it is `true` after a successful retry, the `downloadFailedView` should still be hidden.

PR #1570, opened 2022-06-14
OE-512 Login navigation

ettore:
**What's this do?**
- implement new First Book sign-in UI
- hook that up with existing sign-in business logic
- change token error handling to parse problem documents if present, in order to display the server error messages to the user

https://jira.nypl.org/browse/OE-512

I will update the ticket

merging to feature branch

I will send a build out later this week, so no for now.

yes

@ettore  tested various iPhones on iOS 12 and 15, and various iPads on iOS 15.


Simplified/SignInLogic/OELoginChoiceViewController.swift
```diff
@@ -49,34 +57,42 @@ class OELoginChoiceViewController : UIViewController {
...
 
-  @IBAction func didSelectClever() {
-    didSelectAuthenticationMethod(.clever)
-  }
-
-  @IBAction func didSelectFirstBook() {
-    didSelectAuthenticationMethod(.firstBook)
-  }
-
   private func didSelectAuthenticationMethod(_ loginChoice: LoginChoice) {
```
ErnestFan, line 96: Maybe the body of this function can be moved into `didSelectClever()`? Or make this a `clever` specific method? Since `Firstbook` no longer uses the logic in this method.

ErnestFan (approved):

Simplified/SignInLogic/OELoginChoiceViewController.swift
ettore: that's true, but I ignored it because I will need to work on the Clever side soon too, so I  just left it there.

ettore:
forgot to fix a warning in the last commit...

PR #1571, opened 2022-06-15
OE-513 Hook up links to web pages for sign-in flow

ettore:
**What's this do?**
Hook up links to HTML pages in new sign in screens

https://jira.nypl.org/browse/OE-513

I'll update the ticket

merging to feature branch

not yet

yes

@ettore


Simplified/SignInLogic/OELoginFirstBookVC.swift
```diff
@@ -141,6 +141,28 @@ class OELoginFirstBookVC: UIViewController {
     businessLogic.logIn()
   }
 
+  @IBAction func showTroubleshooting() {
+    let url = URL(string: "https://www.openebooks.org/signin-help")!
```
ErnestFan, line 145: Is there a reason we don't store this and the URLs below in the `NYPLLibraryAccountURLsProvider` class?

ettore: yeah, I looked at that and that logic is hooked up with the Account class and Settings logic, so it didn't seem the right thing to do. These links are not included in the authentication document or library catalog after all.

ErnestFan (approved):

PR #1569, opened 2022-06-10
iOS-411 Implement delete server data

ErnestFan:
**What's this do?**
Implement UI for `Delete my SimplyE data`
Remove NYPLSettingsAdvancedVC

**Why are we doing this? (w/ JIRA link if applicable)**
[iOS-411](https://jira.nypl.org/browse/IOS-411)

**How should this be tested? / Do these changes have associated tests?**
1. Log in to NYPL
2. Go to Settings -> Accounts -> NYPL
3. Enable `Sync Bookmarks`
4. Select `Delete my SimplyE data` which is under `Sync Bookmarks`
5. Press `Delete Reading Data` button on the screen that shown up
6. With internet connection, you should be back to the account detail page and an alert of deletion succeed should show up
7. Without internet connection, you should stay on the same page and a `Please try again` alert should show up

**Dependencies for merging? Releasing to production?**
N/A

**Does this include changes that require a new SimplyE/Open eBooks build for QA?**
Will bump build number

**Has the application documentation been updated for these changes?**
N/A

**Did someone actually run this code to verify it works?**
@ErnestFan

PR #1572, opened 2022-06-16
iOS-413 Implement Delete Library Card

ErnestFan:
**What's this do?**
Implement UI for delete library card

[iOS-413](https://jira.nypl.org/browse/IOS-413)

1. Log in to NYPL
2. Go to Settings -> Accounts -> NYPL
3. Select `Delete Library Card` which is at the very bottom
4. Press `Email to cancel your library account` button on the screen that shown up
5. An email composer should show up
6. It should have the receiver set to the library support email and has patron barcode in the email body
7. Edit the receiver to yourself and send a test email
8. An alert should show up and you are navigated back to account details screen after dismissing the alert

N/A

Will check UI behaviour and wordings with Andrew while Risa is out

N/A

@ErnestFan


Simplified/Settings/NYPLSettingsDeleteLibraryCardViewController.swift
```diff
@@ -0,0 +1,205 @@
...
+    case .failed:
+      if let error = error {
+        let alert = NYPLAlertUtils.alert(title: "Error", error: error as NSError)
+        NYPLAlertUtils.presentFromViewControllerOrNil(alertController: alert,
+                                                      viewController: self,
+                                                      animated: true,
+                                                      completion: nil)
+      }
+    case .sent:
+      // TODO: Check wordings with Risa
```
ErnestFan, line 181: @ettore The ticket/figma does not describe the behaviour after email is sent, so I am displaying an alert and a dismiss action here. Will check with Andrew/Risa.

Also, this PR contains commits that are for the iOS-411 ticket and those are reviewed. Only the last commit (762441f23a2) needs to be reviewed for this PR.

ettore: looks good to me, so does this mean you will just merge this PR instead of the other one?

ErnestFan: Yeah I think merging this PR instead of both PRs make more sense

ettore (approved):

Simplified/Settings/NYPLSettingsDeleteLibraryCardViewController.swift
ErnestFan: Turns out github automatically merged [the other PR](https://github.com/NYPL-Simplified/Simplified-iOS/pull/1569) when I merge this PR, because this PR contains the commits from the other PR

PR #1587, opened 2022-07-21
OE-604 Fix delete data wordings for OEB

ErnestFan:
**What's this do?**
The alert after deleting server data now shows Open eBooks instead of SimplyE
Bump version to 2.5.1

[OE-604](https://jira.nypl.org/browse/OE-604)

Log into Open eBooks
Go to 'Account' -> 'Delete Server Data'
After deleting server data, it should show an alert with Open eBooks name on it

N/A

Yes

N/A

@ErnestFan

### DIFF
--- a/Simplified/it.lproj/Localizable.strings
+++ b/Simplified/it.lproj/Localizable.strings
@@ -257,12 +257,13 @@
 "13 or Older" = "13 o di più";
 
 // Account / Data deletion
-"Delete my SimplyE Data" = "Rimuovi i miei dati di SimplyE";
-"Delete my Open eBooks Data" = "Rimuovi i miei dati di Open eBooks";
+"Delete my SimplyE Data" = "Rimuovi i miei dati";
+"Delete my Open eBooks Data" = "Rimuovi i miei dati";
 "Delete Reading Data" = "Rimozione dati";
 "Please wait..." = "Attendere, prego...";
 "Delete all the bookmarks you have saved in the cloud." = "Tutti i segnalibri salvati sul server saranno rimossi.";
-"You have successfully deleted your SimplyE data" = "I tuoi dati di SimplyE sono stati rimossi";
+"You have successfully deleted your SimplyE data" = "I dati del tuo utente SimplyE sono stati rimossi";
+"You have successfully deleted your Open eBooks data" = "I dati del tuo utente Open eBooks sono stati rimossi";
 "Unsubscribe from emails" = "Rimuovimi dalle email";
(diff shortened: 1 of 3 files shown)


Simplified/it.lproj/Localizable.strings
```diff
@@ -263,6 +263,7 @@
 "Please wait..." = "Attendere, prego...";
 "Delete all the bookmarks you have saved in the cloud." = "Tutti i segnalibri salvati sul server saranno rimossi.";
 "You have successfully deleted your SimplyE data" = "I tuoi dati di SimplyE sono stati rimossi";
+"You have successfully deleted your Open eBooks data" = "I tuoi dati di Open eBooks sono stati rimossi";
```
ErnestFan, line 266: @ettore Can you check if the Italian wording looks right to you?

ettore: actually these strings do look bit odd.... can you change both OE and SE strings to say: "I dati del tuo utente di SimplyE / Open eBooks sono stati rimossi"

ettore: also at line 260 / 261 let's just do simply for both apps: "Rimuovi i miei dati". It's implicit we're talking about SE/OE, but if I add SimplyE / OpenE it always sounds weird in Italian. (Italian doesn't have the Saxon genitive form, so I'd be forced to use the equivalent of `of`, and that gets weird because we're referring to "my" data, but also the data "of" SimplyE/OE....)

Simplified.xcodeproj/project.pbxproj
```diff
@@ -5771,7 +5771,7 @@
 				CLANG_WARN_NON_LITERAL_NULL_CONVERSION = YES_ERROR;
 				CODE_SIGN_ENTITLEMENTS = Simplified/SimplyE.entitlements;
 				CODE_SIGN_IDENTITY = "iPhone Developer";
-				CURRENT_PROJECT_VERSION = 7;
+				CURRENT_PROJECT_VERSION = 0;
```
ettore, line 5774: this is now obsolete since we are already at build 3 for OE 2.5.1 on develop, I think. Which makes me think that it's also weird that GitHub reports 2.5.1 as a change, since it's already on develop.

Do you want to verify/update (or remove) the version.build numbers and just merge this PR? there are just minor changes in the Italian localization.

ErnestFan: Updated the Italian wordings and removed the version.build numbers. However, I will wait until the audiobook feature branch is merged, to avoid conflicts on the other PR.

ettore (approved):

PR #1595, opened 2022-08-04
iOS-434 Use barcode/pin for request header when auth token is missing

ErnestFan:
**What's this do?**
Use barcode/pin for request header when auth token is missing

[iOS-434](https://jira.nypl.org/browse/IOS-434)

See ticket

N/A

Yes

N/A

@ErnestFan

### DIFF
--- a/Simplified.xcodeproj/project.pbxproj
+++ b/Simplified.xcodeproj/project.pbxproj
@@ -5600,7 +5600,7 @@
 				CLANG_WARN_NON_LITERAL_NULL_CONVERSION = YES_ERROR;
 				CODE_SIGN_ENTITLEMENTS = Simplified/SimplyE.entitlements;
 				CODE_SIGN_IDENTITY = "iPhone Developer";
-				CURRENT_PROJECT_VERSION = 3;
+				CURRENT_PROJECT_VERSION = 0;
 				DEVELOPMENT_TEAM = 7262U6ST2R;
 				ENABLE_BITCODE = NO;
 				"EXCLUDED_ARCHS[sdk=iphonesimulator*]" = arm64;
@@ -5620,7 +5620,7 @@
 					"$(inherited)",
 					"@executable_path/Frameworks",
 				);
-				MARKETING_VERSION = 3.9.0;
+				MARKETING_VERSION = 3.9.1;
 				PRODUCT_BUNDLE_IDENTIFIER = org.nypl.labs.SimplyE;
 				PRODUCT_MODULE_NAME = SimplyE;
 				PRODUCT_NAME = "SimplyE-R2dev";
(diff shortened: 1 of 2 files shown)


Simplified/Network/NYPLNetworkExecutor.swift
```diff
@@ -85,6 +85,17 @@ enum NYPLResult<SuccessInfo> {
       ]
 
       headers.forEach { request.setValue($0.value, forHTTPHeaderField: $0.key) }
+    } else if let username = responder.credentialsProvider.username,
```
ettore, line 88: the "Authorization" header is a reserved header and in theory we are not supposed to set it per official docs: https://developer.apple.com/documentation/foundation/nsurlrequest#1776617

The official way to set authorization headers is via the authentication challenge: for us this is in NYPLBasicAuth.swift, which _should_ be triggered automatically if the request goes through NYPLNetworkExecutor. It is after all for all other requests. So there must be something particular about the annotation requests that prevents this somehow. I can also take a look since I wrote this code.

The one exception to this rule is for Bearer token authentication, for which there's a known bug that prevents the challenge mechanism from being called.

ettore: I just did a test using the simulator and Charles proxy and I do see the Authorization header being added to the request, eg. `authorization: Basic aG93bXXXXXXXX4Ng==`.  I do get an error that points to a "invalid annotation format" problem ("detail": "The annotation could not be parsed as JSON-LD."). The content type we are sending is `application/x-www-form-urlencoded`, I wonder if that needs to be changed?

ettore: that was it. I just did a request rewrite in Charles changing `application/x-www-form-urlencoded` to `application/json` and it worked. Did we change this recently?  Or maybe it's just a bug I introduced during the last annotations refactor where I routed calls thru NYPLNetworkExecutor.

ettore: sure enough, we used to have a method in NYPLAnnotations where we set the Authorization and Content-type headers: https://github.com/NYPL-Simplified/Simplified-iOS/commit/524098af47340b308aa45164811bbe2d0d8e395b#diff-8ec0b795b7914971e927c857341fde15782df4f7a9d853b2e1a9c7c3843422a4L634-L660

So I think the best thing to do would be to set the Content-Type back to that value for all those annotations requests instead than in NYPLNetworkExecutor (which will affect _every_ request)

ErnestFan: Didn't know it was already done in NYPLBasicAuth.swift via the authentication challenge, thank you for catching this. I have updated the PR so it passes in the content type from NYPLAnnotations.

ettore (approved):

PR #1601, opened 2022-08-19
IOS-412 Add unsubscribe from emails logic

ettore:
**What's this do?**
@ErnestFan your logic seemed good to me, or at least I couldn't find any problems. All I did was:
1. I removed the hardcoded URL,
2. refactored the method to read/write the auth doc URLs so that they are simpler/shorter, 
3. added a `stringValue` method to convert the URLType objc-style enum int values to string
4. other minor stuff

@jkim-git I tried to add you to reviewers but it's not letting me, probably because you didn't accept the invite to join the repo as a collaborator yet. No worries though, just wanted to let you know.

https://jira.nypl.org/browse/IOS-412

see ticket

no -- the URL once loaded in the web view shows the header and footer navigation, but the team that owns the page does not have any bandwidth to remove them. In my testing only one link opens inside the web view, all others open up the Safari app.

yes

n/a

@ettore


ErnestFan (approved):

ettore:
I added a commit to implement the requested change to open the Unsubscribe URL in Safari instead of webview.

ErnestFan (approved):

PR #1608, opened 2022-10-07
iOS-444 Implement serialization test for NYPLAudiobookBookmark

ErnestFan:
**What's this do?**
Implement serialization test for NYPLAudiobookBookmark

[iOS-444](https://jira.nypl.org/browse/IOS-444)

Run unit tests

NYPLAudiobookToolkit will be updated once the PR on that repo is merged

N/A, not testable by QA yet

N/A

@ErnestFan


SimplifiedTests/NYPLBookmarkSpec+AudiobookTests.swift
```diff
@@ -1,137 +0,0 @@
-//
```
ErnestFan, line 1: @ettore This file is being removed because it was migrated to NYPLAudiobookToolkit, I just forgot to remove it in previous PR

ettore: thank you for being thorough @ErnestFan  -- it's really difficult to keep track for things that are missing or should be missing.

ettore (approved):

ettore:
the test failure  is probably caused by not referring to the right commit on NYPLAudiobookToolkit?

ErnestFan:
> the test failure is probably caused by not referring to the right commit on NYPLAudiobookToolkit?

Yes, it passes unit tests on both SE and OEB when I ran them locally

PR #1606, opened 2022-09-21
iOS-423 Implement NYPLAudiobookBookmarkBusinessLogic

ErnestFan:
**What's this do?**
Implement NYPLAudiobookBookmarkBusinessLogic
Update NYPLBookRegistry, NYPLBookRegistryRecord and NYPLAnnotations to work with audiobook bookmarks
Fix existing unit tests

**Why are we doing this? (w/ JIRA link if applicable)**
[iOS-423](https://jira.nypl.org/browse/IOS-423)

**How should this be tested? / Do these changes have associated tests?**
Will implement unit tests

**Dependencies for merging? Releasing to production?**
NYPLAudiobookToolkit will be updated once the PR on that repo is merged

**Does this include changes that require a new SimplyE/Open eBooks build for QA?**
N/A, not testable by QA yet

**Has the application documentation been updated for these changes?**
N/A

**Did someone actually run this code to verify it works?**
@ErnestFan, it builds and the Readium bookmark is working, audiobook bookmark is yet to be tested

PR #1609, opened 2022-10-27
iOS-425 Integrate audiobook bookmark business logic into audiobook manager

ErnestFan:
**What's this do?**
Integrate bookmark business logic into audiobook manager
Update NYPLAudiobookToolkit commit ref
Update localized strings
Update related unit tests

[iOS-424](https://jira.nypl.org/browse/IOS-424)
[iOS-425](https://jira.nypl.org/browse/IOS-425)

See ticket

Updated NYPLAudiobookToolkit

Yes

N/A

@ErnestFan

### DIFF
--- a/Simplified/Audiobooks/Bookmark/NYPLAudiobookBookmarksBusinessLogic.swift
+++ b/Simplified/Audiobooks/Bookmark/NYPLAudiobookBookmarksBusinessLogic.swift
@@ -9,11 +9,23 @@
 import Foundation
 import NYPLAudiobookToolkit
 
-class NYPLAudiobookBookmarksBusinessLogic: NYPLAudiobookBookmarksBusinessLogicDelegate {
+class NYPLAudiobookBookmarksBusinessLogic: NYPLAudiobookBookmarking {
   
   // MARK: - Properties
   
-  var bookmarks: [NYPLAudiobookBookmark]
+  var bookmarks: [NYPLAudiobookBookmark] {
+    didSet {
+      updateBookmarksDictionary()
+    }
+  }
+  
+  /// Given that we need to constantly check if there exists a bookmark
+  /// at the audio player current location.
+  /// We use the chapter info as key and store an array of bookmarks
+  /// of the same chapter into the dictionary. This improve the performance by
+  /// avoiding comparison between the current location and
+  /// all the existing bookmarks of this audiobook.
+  var bookmarksDictionary: [String: [NYPLAudiobookBookmark]]
   
   let book: NYPLBook
   private let drmDeviceID: String?
@@ -25,6 +37,14 @@ class NYPLAudiobookBookmarksBusinessLogic: NYPLAudiobookBookmarksBusinessLogicDe
     return bookmarks.count
   }
   
+  var shouldAllowRefresh: Bool {
+    return annotationsSynchronizer.syncIsPossibleAndPermitted()
+  }
+  
+  var noBookmarksText: String {
+    return NSLocalizedString("There are no bookmarks for this book.", comment: "Text showing in bookmarks view when there are no bookmarks")
+  }
+  
   // MARK: - Init
   
   init(book: NYPLBook,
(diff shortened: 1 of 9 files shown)


Simplified/Audiobooks/Bookmark/NYPLAudiobookBookmarksBusinessLogic.swift
```diff
@@ -9,11 +9,23 @@
...
   // MARK: - Properties
   
-  var bookmarks: [NYPLAudiobookBookmark]
+  var bookmarks: [NYPLAudiobookBookmark] {
+    didSet {
+      updateBookmarksDictionary()
+    }
+  }
+  
+  /// Given that we need to constantly check if there exists a bookmark
```
ErnestFan, line 22: @ettore Feel free to make any suggestions if you have a better solution to improve the performance.

Simplified/Book/Models/NYPLBookRegistry.h
```diff
@@ -29,7 +29,15 @@ typedef NS_ENUM(NSInteger, NYPLBookState);
 
 @end
 
+#if FEATURE_AUDIOBOOKS
+// Hiding warnings for protocol defined in Swift is not accessible in ObjC
+#pragma clang diagnostic push
```
ErnestFan, line 34: @ettore From the previous PR ([link](https://github.com/NYPL-Simplified/Simplified-iOS/pull/1606#discussion_r978741090)), we decided to declare the conformance in the `NYPLBookRegistry.m` file. However, that would lead to compiling error when we dependency inject this class into the business logic, because the conformance cannot be seen on the Swift side.

Simplified/Book/UI/NYPLBookCellDelegate+AudiobookBookmark.swift
```diff
@@ -0,0 +1,27 @@
...
+#if FEATURE_AUDIOBOOKS
+import Foundation
+import NYPLAudiobookToolkit
+import UIKit
+
+/// The NYPLAudiobookBookmarksBusinessLogic is not accessible from ObjC,
+/// so we do this in a Swift extension file.
+@objc extension NYPLBookCellDelegate {
+  @objc(setBookmarkBusinessLogicForBook:AudiobookManager:)
+  func setBookmarkBusinessLogic(for book: NYPLBook, _ audiobookManager: DefaultAudiobookManager) {
```
ettore, line 18: can you inject at least the BookRegistry, since it's already present where this is called from?

ettore (commented):
looks good overall -- I was wondering if it's possible to add some unit tests for the business logic code. These could also go in a separate ticket.

Simplified/Book/Models/NYPLBookRegistry.h
ErnestFan: We will need to implement all the functions in this protocol, otherwise it will not compile since Swift compiler does not see the `NYPLBookRegistry` conforming to the protocol. Maybe we can implement the required functions and call the ObjC functions from the Swift extension?

ErnestFan:
@ettore I pushed a new commit for updating the unit tests regarding the newly added public functions in the business logic, let me know what you think.

SimplifiedTests/NYPLAudiobookBookmarksBusinessLogicTests.swift
```diff
@@ -414,18 +414,75 @@ class NYPLAudiobookBookmarksBusinessLogicTests: XCTestCase {
...
+    bookmarkBusinessLogic.addAudiobookBookmark(thirdChapterLocation)
+    
+    guard let firstBookmark = bookmarkBusinessLogic.bookmarkExisting(at: firstChapterLocation),
+          let secondBookmark = bookmarkBusinessLogic.bookmarkExisting(at: secondChapterLocation),
+          let thirdBookmark = bookmarkBusinessLogic.bookmarkExisting(at: thirdChapterLocation) else {
+      XCTFail("Failed to add/retrieve bookmark")
+      return
+    }
+    
+    XCTAssertEqual(bookmarkBusinessLogic.bookmarkIsFirstInChapter(firstBookmark), true)
```
ettore, line 462: nit: you can use XCTAssertTrue / XCTAssertFalse for brevity

ettore (approved):

ErnestFan:
Currently we are muting a bunch of build warnings in `NYPLBookRegistry.h` because we forward declared `NYPLAudiobookBookRegistry` protocol. The compiler is complaining the protocol is empty because the actual protocol is defined in Swift and can not be seen in the ObjC header file.

Based on the discussion with @ettore on slack, I have tried the workaround below.
**Expose some properties of the NYPLBookRegistry, so we can write a Swift extension around it to avoid the forward declaration.**
I ran into some problems with this approach.

1. Build Issue
In `NYPLBookRegistry`, we access `NYPLBookRegistryRecord` to read/write the bookmarks of a book and store it in the book registry. However, since `NYPLAudiobookBookmark` is also forward declared on `NYPLBookRegistryRecord` header file. The Swift compiler is confused because it sees both the Swift definition and the ObjC forward declaration of `NYPLAudiobookBookmark`. This would cause build errors with any functions related to `NYPLAudiobookBookmark` in `NYPLBookRegistryRecord` because they are not recognized by the compiler.
2. Possible thread safety
In ObjC side of `NYPLBookRegistry`, we use `@synchronized(self)` to ensure thread safety, which is unaccessible in Swift. So even we can make the Swift extension thread-safe, it might still run into a race condition with the ObjC side of code. We can possibly fix this issue by defining a dispatch queue for thread safety on both ObjC and Swift side.

Both issues can be fixed by rewriting `NYPLBookRegistry` and `NYPLBookRegistryRecord` in Swift, but we do not think it worths the hassle. We will stick with the ugly fix (muting the warnings) for now and revisit this later.

ettore:
@ErnestFan sorry for being late, but would it be possible to ignore only the warning we actually need to ignore instead of `-Weverything`?

ErnestFan:
> @ErnestFan sorry for being late, but would it be possible to ignore only the warning we actually need to ignore instead of `-Weverything`?

That's actually the first thing I tried before I used `-Weverything`, I tried `-Wprotocol` but the warnings did not go away. I then looked into the warnings and none of those has the warning `type` in the warning details.

PR #1611, opened 2022-11-07
PJR-69 Remove download logic dependency from book opening logic

ettore:
**What's this do?**
Removes download logic dependency from class that opens an Axis book.

Other minor renames and cleanups.

[PJR-69](https://jira.nypl.org/browse/PJR-69)

no breaking changes, will be covered by regression.

This is dependent on https://github.com/NYPL-Simplified/Axis-iOS/pull/22 which should be merged first. I'll then need to update the commit ref for the Axis-iOS submodule.

no

yes

@ettore

### DIFF
--- a/Simplified.xcodeproj/project.pbxproj
+++ b/Simplified.xcodeproj/project.pbxproj
@@ -2527,7 +2527,6 @@
 				73579046276BE7F2009F1ADF /* NYPLUtilities in Frameworks */,
 				732F046D2633783B0018A82E /* PDFRendererProvider.xcframework in Frameworks */,
 				7347B4E5265EFD3C00E4E386 /* PromisesObjC.xcframework in Frameworks */,
-				731465202786106900027CFA /* PureLayout in Frameworks */,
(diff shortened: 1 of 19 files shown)


Simplified.xcodeproj/project.pbxproj
```diff
@@ -2527,7 +2527,6 @@
 				73579046276BE7F2009F1ADF /* NYPLUtilities in Frameworks */,
 				732F046D2633783B0018A82E /* PDFRendererProvider.xcframework in Frameworks */,
 				7347B4E5265EFD3C00E4E386 /* PromisesObjC.xcframework in Frameworks */,
-				731465202786106900027CFA /* PureLayout in Frameworks */,
```
ErnestFan, line 2530: Why are we removing this?

ettore: it was a duplicate, not sure how it got added twice

jkim-git (approved):

PR #1613, opened 2022-11-10
PJR-70 Dependency-inject NYPLAnnotations calls

ettore:
**What's this do?**
Makes it so that the notion of NYPLAnnotations is not implicit inside the reader, but is injected at the entry point (NYPLR2Owner).

PJR-70

No breaking changes. A regression will cover this

n/a

no

yes

@ettore

### DIFF
--- a/Simplified/Reader2/Networking/NYPLAnnotations.swift
+++ b/Simplified/Reader2/Networking/NYPLAnnotations.swift
@@ -6,7 +6,7 @@ import NYPLAudiobookToolkit
 #endif
 
 protocol NYPLAnnotationSyncing: AnyObject {
-  // Server status
+  // Server sync status
   
   static func requestServerSyncStatus(forAccount userAccount: NYPLUserAccount,
                                       settings: NYPLAnnotationSettings,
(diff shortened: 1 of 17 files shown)


Simplified/Reader2/Networking/NYPLAnnotations.swift
```diff
@@ -595,16 +594,17 @@ final class NYPLAnnotations: NSObject, NYPLAnnotationSyncing {
...
   }
 
-  /// Annotation-syncing is possible only if the given `account` is signed-in
+  /// Annotation-syncing is possible only if the current user account is signed-in
   /// and if the currently selected library supports it.
-  class func syncIsPossible(_ account: NYPLUserAccount) -> Bool {
+  private class func syncIsPossible() -> Bool {
     let library = AccountsManager.shared.currentAccount
-    return account.hasCredentials() && library?.details?.supportsSimplyESync == true
+    let userAccount = NYPLUserAccount.sharedAccount()
```
ettore, line 601: @jkim-git  I made NYPLUserAccount implicit because if we go the hybrid route the need for the user account status will be entirely handled by the web side. Essentially it's an implementation detail. 

In a fully native world of course having this as an implicit dependency would not be good.

ErnestFan (approved):

PR #1615, opened 2022-11-15
Merge release branch to master

ErnestFan:
**What's this do?**
Merge release branch to master

Part of post-release flow

N/A

N/A

N/A

N/A

@ErnestFan


ettore:
> So instead of merging the release branch on to master, I rebase master into the release branch in order to keep things up-to-date.

From looking at the the branches it seems like you merged the release branch into `release_se_3.9.2_master_fix`, and then we are merging that branch onto master here. If so I think that's fine. 

But what you wrote above maybe I am misunderstanding? Basically we should never rebase `master` or any of the long lived branches (like `develop`, or any release branch) because `rebase` means rewriting history, and we don't want that on those branches. However, I don't think that's what happened here, but let me know.

ErnestFan:
> > So instead of merging the release branch on to master, I rebase master into the release branch in order to keep things up-to-date.
> 
> From looking at the the branches it seems like you merged the release branch into `release_se_3.9.2_master_fix`, and then we are merging that branch onto master here. If so I think that's fine.
> 
> But what you wrote above maybe I am misunderstanding? Basically we should never rebase `master` or any of the long lived branches (like `develop`, or any release branch) because `rebase` means rewriting history, and we don't want that on those branches. However, I don't think that's what happened here, but let me know.

The `release_se_3.9.2_master_fix` is originally the 3.9.2 release branch with `git pull master --rebase`. If I use `master` as base and do `git merge release/simplye/3.9.2' or vice versa, some deleted files stay like I state above. Maybe the last post-release PR I did messed up the commit history and caused it. What do you think?

PR #1617, opened 2022-11-16
PJR-86 Remove AccountManager implicit dependency from reader

ettore:
**What's this do?**
Removes the dependency on the AccountManager singleton from the reader classes. It turns out all it was needed was the flag to see if we had permission to sync bookmarks. So I just passed the flag in. 

Note that this solution assumes that this flag won't change while the reader is open, which is currently the case for SimpleE/OpenE and it seems like a reasonable assumption to me. However if you think this assumption is invalid, I have also implemented it by passing a protocol in instead. This alternative is implemented by the [1st commit](https://github.com/NYPL-Simplified/Simplified-iOS/commit/3c003f5c5229918c7bcb981317a52bb6ede2a70e). The 2nd checkin replaces the protocol with the direct flag value.  

Please let me know which solution you prefer.

[PJR-86](https://jira.nypl.org/browse/PJR-86)

regression will cover this. 

n/a 
No breaking changes.

no

yes

@ettore verified reading position and bookmarks are still saved and synced across devices.

### DIFF
--- a/Simplified/Accounts/Library/Account.swift
+++ b/Simplified/Accounts/Library/Account.swift
@@ -27,9 +27,13 @@ class OPDS2SamlIDP: NSObject, Codable {
   func isSignedIn() -> Bool
 }
 
+@objc protocol NYPLReaderServerPermissions {
+  var syncPermissionGranted:Bool {get}
+}
+
 // MARK: AccountDetails
 // Extra data that gets loaded from an OPDS2AuthenticationDocument,
-@objcMembers final class AccountDetails: NSObject {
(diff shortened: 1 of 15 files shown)


jkim-git:
I think it is a reasonable assumption that the permission will not change -- but I will check in with Risa to make sure that is correct

jkim-git (approved):
Since Risa is out today, I will approve it so you can be unblocked. We can always use the other approach if needed. Thank you for your thoughtfulness in including both

jkim-git:
Just checked in with Risa, and that is a reasonable assumption we can make from a product stand point as well

ErnestFan (approved):

PR #1618, opened 2022-11-19
PJR-87 Use NYPLAnnotations as a regular class & refactor related protocols

ettore:
**What's this do?**
- Remove NYPLNetworkExecutor dependency from NYPLAnnotations API
- Change the way NYPLAnnotations is used by making it a regular class so it can have its own related state (e.g. the `failFastExecutor` instance var)
- Split NYPLAnnotationSyncing protocol by making it more granular. This way downstream classes can refer to what they actually need. For example, NYPLLastReadPositionPoster only needs the API to update the last read position. NYPLSignInBusinessLogic only needs the api to check and update the sync toggle.

PJR-87

regression should cover this

no breaking changes

no

yes

@ettore

### DIFF
--- a/Simplified/Book/UI/NYPLBookCellDelegate+AudiobookBookmark.swift
+++ b/Simplified/Book/UI/NYPLBookCellDelegate+AudiobookBookmark.swift
@@ -23,7 +23,7 @@ import UIKit
       drmDeviceID: NYPLUserAccount.sharedAccount().deviceID,
       bookRegistryProvider: audiobookRegistryProvider,
       syncPermission: AccountsManager.shared.syncPermissionGranted,
-      annotationsSynchronizer: NYPLAnnotations.self)
+      annotationsSynchronizer: NYPLRootTabBarController.shared().annotationsSync)
     audiobookManager.bookmarkBusinessLogic = bizLogic
   }
 }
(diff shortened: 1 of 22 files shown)


Simplified/Book/UI/NYPLBookCellDelegate+AudiobookBookmark.swift
```diff
@@ -23,7 +23,7 @@ import UIKit
       drmDeviceID: NYPLUserAccount.sharedAccount().deviceID,
       bookRegistryProvider: audiobookRegistryProvider,
       syncPermission: AccountsManager.shared.syncPermissionGranted,
-      annotationsSynchronizer: NYPLAnnotations.self)
+      annotationsSynchronizer: NYPLRootTabBarController.shared().annotationsSynchronizer)
```
ettore, line 26: this is not ideal. I did this because (1) the NYPLAnnotations type is not objc-compatible, and (2) because this is in a class external to the reader.

Simplified/Reader2/Networking/NYPLLastReadPositionSynchronizer.swift
```diff
@@ -25,21 +25,16 @@ class NYPLLastReadPositionSynchronizer: NYPLLastReadPositionSynchronizing {
...
   /// Designated initializer.
   ///
   /// - Parameters:
   ///   - bookRegistry: The registry that stores the reading progresses.
   init(bookRegistry: NYPLBookRegistryProvider,
-       synchronizer: NYPLAnnotationSyncing.Type) {
+       synchronizer: NYPLLastReadPositionSupportAPI) {
     self.bookRegistry = bookRegistry
     self.synchronizer = synchronizer
-    failFastNetworkExecutor = NYPLNetworkExecutor(
```
ettore, line 39: this was moved inside NYPLAnnotations because it belongs there with the networking logic. Ideally we should dependency-inject it into the NYPLAnnotations initializer.

ErnestFan (approved):

PR #1621, opened 2023-01-28
iOS-457 Fix crash when opening audiobook while downloading

ErnestFan:
**What's this do?**
Fix crash when opening audiobook while downloading

[iOS-457](https://jira.nypl.org/browse/IOS-457)

See ticket

N/A

Yes

N/A

@ErnestFan


Simplified/Book/UI/NYPLBookCellDelegate+AudiobookLastListenPosition.swift
```diff
@@ -81,7 +81,9 @@ private let NYPLAudiobookPositionSyncingInterval: DispatchTimeInterval = .second
         // Present audio player
         NYPLRootTabBarController.shared().pushViewController(audiobookPlayerVC, animated: true)
         // Call completion handler when audiobook has been successfully opened
-        successCompletion()
+        if let successCompletion = successCompletion {
```
ettore, line 84: nit: this is fine, but you can also do `successCompletion?()` for brevity

ettore (approved):
looks good!

PR #1625, opened 2023-10-19
IOS-578 Suppress Readium 2.5 PublicationServer deprecation warnings

ettore:
**What's this do?**
The switch from `PublicationServer` to `GCDHTTPServer` in Readium 2.5.0 caused some regressions when browsing ePubs with VoiceOver:
- rendering issues while browsing (swift L/R) when scroll mode is not set
- no automatic scroll when continuous reading mode is enabled.

This silences the many deprecation warnings that are now emitted while still using the old PublicationServer.

It also fixes a few other Xcode 14.3 warnings.

https://jira.nypl.org/browse/IOS-578

You should not see any warnings, except for one in `LibraryService.swift`, which seems unavoidable.

n/a

no

yes

yes, @ettore

### DIFF
--- a/Simplified/Network/NYPLNetworkQueue.swift
+++ b/Simplified/Network/NYPLNetworkQueue.swift
@@ -84,7 +84,7 @@ final class NetworkQueue: NSObject {
       let methodString = method.rawValue
       let dateCreated = NSKeyedArchiver.archivedData(withRootObject: Date())
       // left for backward compatibility
-      let headerData = NSKeyedArchiver.archivedData(withRootObject: [:])
+      let headerData: Data? = nil
 
       guard let db = self.startDatabaseConnection() else { return }
       
--- a/Simplified/Reader2/ReaderStackConfiguration/LibraryService.swift
+++ b/Simplified/Reader2/ReaderStackConfiguration/LibraryService.swift
@@ -15,6 +15,43 @@ import UIKit
 import R2Shared
 import R2Streamer
 
+// =============================================================================
+// MARK: - Workaround for VoiceOver Issues
+
+/// Temporary workaround to continue using the old Readium `PublicationServer`.
+///
+/// Readium 2.5.0 deprecates `PublicationServer` and introduces
+/// `GCDHTTPServer` in its place. However, this causes significant scrolling
+/// and book content navigation issues with VoiceOver. By casting your
+/// `PublicationServer` references to `PublicationServing`, this workaround
+/// silences the deprecation warnings until we find an official solution.
+protocol PublicationServing: ResourcesServer {
+  func add(_ publication: Publication) throws
+  func removeAll()
+}
+
+extension PublicationServer: PublicationServing {
+  @available(*, deprecated, message: "To suppress this warning, cast to PublicationServing protocol")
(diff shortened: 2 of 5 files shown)


Simplified/Network/NYPLNetworkQueue.swift
```diff
@@ -84,7 +84,7 @@ final class NetworkQueue: NSObject {
       let methodString = method.rawValue
       let dateCreated = NSKeyedArchiver.archivedData(withRootObject: Date())
       // left for backward compatibility
-      let headerData = NSKeyedArchiver.archivedData(withRootObject: [:])
+      let headerData: Data? = nil
```
ettore, line 87: I looked at the code in the history, since we are no longer sending header data as a parameter (line 78) this is equivalent according to the history.

Simplified/Reader2/ReaderStackConfiguration/LibraryService.swift
```diff
@@ -24,14 +61,20 @@ import R2Streamer
...
+  let publicationServer: PublicationServing
   private var drmLibraryServices = [DRMLibraryService]()
   
   private lazy var documentDirectory = try! FileManager.default.url(for: .documentDirectory, in: .userDomainMask, appropriateFor: nil, create: true)
   
-  init(publicationServer: PublicationServer) {
-    self.publicationServer = publicationServer
-    
+  init() {
+    guard let server = (PublicationServerFactory() as PublicationServerMaking).make() else {
```
ettore, line 70: this is the opposite of dependency injection, so it's a "bad practice" per se. I only did it to control the creation of the PublicationServer where it's actually needed. This also allows us to keep those workaround protocols `private`. This code is hopefully temporary anyway (last famous words).

I can also move it out where it was if we want.

Simplified/Reader2/ReaderStackConfiguration/LibraryService.swift
```diff
@@ -15,6 +15,43 @@ import UIKit
...
+/// `GCDHTTPServer` in its place. However, this causes significant scrolling
+/// and book content navigation issues with VoiceOver. By casting your
+/// `PublicationServer` references to `PublicationServing`, this workaround
+/// silences the deprecation warnings until we find an official solution.
+protocol PublicationServing: ResourcesServer {
+  func add(_ publication: Publication) throws
+  func removeAll()
+}
+
+extension PublicationServer: PublicationServing {
```
ettore, line 33: this still emits a warning because `PublicationServer` as a class is deprecated. I don't see a way to work around this, but if you do let me know.

Simplified/Reader2/ReaderStackConfiguration/LibraryService.swift
```diff
@@ -15,6 +15,43 @@ import UIKit
...
+  @available(*, deprecated, message: "To suppress this warning, cast to PublicationServing protocol")
+  func add(_ publication: Publication) throws {
+    try add(publication, at: UUID().uuidString)
```
ettore, line 36: the UUID string was a default param for the `PublicationServer::add(_, at:)` method, but we can't have default params in protocols so I have to add it here explicitly.

PR #1626, opened 2023-11-08
iOS-584 Disable ipa upload to firebase

ExplorerNautilus:
**What's this do?**
Disable ipa upload to firebase due to error

Firebase authentication is deprecated and we'll be using TestFlight instead

N/A

N/A

N/A

Yes

Will be tested on CI

### DIFF
--- a/scripts/firebase-upload.sh
+++ b/scripts/firebase-upload.sh
@@ -32,12 +32,14 @@ else
   fatal "Unable to upload to firebase: missing ad-hoc export!"
 fi
 
+# Disabling the IPA upload to firebase because we encountered authentication error,
+# possibly due to `token` authentication is deprecated.
 # upload app binary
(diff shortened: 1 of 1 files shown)


scripts/firebase-upload.sh
```diff
@@ -32,9 +32,11 @@ else
...
 fi
 
+# Disabling the IPA upload to firebase because we encountered authentication error,
+# possibly due to `token` authentication is deprecated.
 # upload app binary
-echo "Using ipa at $IPA_PATH"
-firebase appdistribution:distribute \
+#echo "Using ipa at $IPA_PATH"
+#firebase appdistribution:distribute \
   --token "${FIREBASE_TOKEN}" \
```
ettore, line 40: hmm, I think you might need to disable these lines 40-42 as well

ettore (approved):

PR #1628, opened 2024-02-01
IOS-639 Upgrade to Xcode 15, drop iOS 11 support, remove banned books without expiration date

ettore:
**What's this do?**
- Upgrade to Xcode 15. This required dropping iOS 11.
- Fix deprecations from dropping iOS 11 support.
- Fix banned books removal code to include books without an expiration date.

IOS-639

- Ideally if we have a device with banned book checked out in January 2023 (?), hence without expiration, updating to this build should remove that book.
- Alternatively we could instrument a QA server to return a book without expiration. That book can be downloaded, but once you relaunch the app it will be removed.

Before merging this, we should merge the following PRs:
https://github.com/NYPL-Simplified/NYPLAudiobookToolkit/pull/184
https://github.com/NYPL-Simplified/DRM-iOS-Adobe/pull/28
and then update the commit refs of those 2 submodules.

not yet

yes

I did

### DIFF
--- a/Simplified/Book/Models/NYPLBookRegistry.m
+++ b/Simplified/Book/Models/NYPLBookRegistry.m
@@ -454,10 +454,11 @@ static NSString *const RecordsKey = @"records";
     NSMutableArray *booksToRemove = [[NSMutableArray alloc] init];
     for (NSString *bookIdentifer in self.identifiersToRecords) {
       NYPLBookRegistryRecord *record = self.identifiersToRecords[bookIdentifer];
+      NYPLOPDSAcquisition *acquisition = record.book.defaultAcquisition;
       // Add the book to remove list if it is distributed by Axis360 and expired
-      if([record.book.defaultAcquisition.type isEqualToString:ContentTypeAxis360]) {
-          if (record.book.defaultAcquisition.availability.until &&
-            [record.book.defaultAcquisition.availability.until compare:[NSDate date]] == NSOrderedAscending) {
+      // or with no expiration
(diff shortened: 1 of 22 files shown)


Simplified/Book/Models/NYPLBookRegistry.m
```diff
@@ -454,10 +454,11 @@ - (void)removeExpiredBannedBooks
...
     for (NSString *bookIdentifer in self.identifiersToRecords) {
       NYPLBookRegistryRecord *record = self.identifiersToRecords[bookIdentifer];
+      NYPLOPDSAcquisition *acquisition = record.book.defaultAcquisition;
       // Add the book to remove list if it is distributed by Axis360 and expired
-      if([record.book.defaultAcquisition.type isEqualToString:ContentTypeAxis360]) {
-          if (record.book.defaultAcquisition.availability.until &&
-            [record.book.defaultAcquisition.availability.until compare:[NSDate date]] == NSOrderedAscending) {
+      // or with no expiration
+      if ([acquisition.type isEqualToString:ContentTypeAxis360]) {
+          if ([acquisition.availability.until compare:[NSDate date]] != NSOrderedDescending) {
```
ettore, line 461: @russellcullen  in objc if `acquisition.availability.until` is nil, the function call `[nil compare:[NSDate date]]` will not result in a crash and instead will just return `0`. This is pretty common in objc! 

So, then, `0` is `!= NSOrderedDescending` and test succeeds 🎉

ExplorerNautilus (approved):

